PR #92, opened 2022-10-26
Migrate CommCare

elias-ba:
Please review, merge, and archive [language-commcare](https://github.com/OpenFn/language-commcare)


josephjclark:
I have pushed a commit here which changes the order of Build and Test in github actions, which should restore the CI build. Happy for this to be merged down.

Issue #119, opened 2022-11-01
New docs have appeared in some packages


josephjclark:
Some packages (`http` and I think `googlesheets` and `dhis2`) have suddenly had new data appearing in ast.json since the last merge.

Unsure if this is significant but it needs checking.

josephjclark:
Ok so this is fine - basically the doc output of `common` changed in 1.7.2 or 1.7.3. When we run `changeset version`, some dependencies are now getting the latest version of common and "inheriting" the changes to the underlying docs.

In other words, this is reflecting a change to `common`'s documentation.

I'n not sure why or when common changed by the way, but the diff looks totally fine and correct to me.

PR #120, opened 2022-11-01
Update Dependencies

josephjclark:
This PR updates a few dependency versions across the adaptors, fixing #119

This is the reason why docs are changing when we run changeset version. The changes themselves are fine, they just need checking in.

I've also noticed a couple of eccentricities in `http`'s dependencies and fixed those.

At the time of writing, the dts build on http is broken.


josephjclark:
I've now merged a fix to DTS generation which should fix the build here.

PR #127, opened 2022-11-08
Add pull request template

josephjclark:
Knocked up a pull request template.

It didn't work for this PR, I presume it'll work when merged into main though.


mtuchi:
> @mtuchi Would you take a look at this and merge it if a) it make sense, b) you can't see any typos, c) you can't think of anything else to add?
> 
> Thanks!

Merging this since I am happy with it, for now, we can modify it in the future if needed @josephjclark

Issue #165, opened 2022-11-17
Update adaptors docs string 


mtuchi:
### Description 
Adaptors doc string are not properly maintained and need to be updated to have the right `jsdoc` tags and include examples for each helper function

| Active Adaptors | Other Adaptors |
| --------------- | -------------- |
| asana           | beyonic        |
| postgresql      | fhir           |
| primero         | bigquery       |
| sftp            | openhim        |
| msgraph         | rapidpro       |
| mysql           | dynamics       |
| mssql           | cartodb        |
| dhis2           | facebook       |
| salesforce      | godata         |
| googlesheets    | khanacademy    |
| http            | magpi          |
| commcare        | progres        |
| mailchimp       | maximo         |
| kobotoolbox     | medicmobile    |
| openmrs         | mogli          |
| openspp         | nexmo          |
| surveycto       | mongodb        |
| ocl             | resourcemap    |
| openfn          | smpp           |
| mailgun         | telerivet      |
| common        | twilio      |
|                 | vtiger      |
|                 | zoho      |

mtuchi:
We should break this down into smaller issues that are adaptor specific. That way it will lighten the load and make it easy to prioritize. Here is our starting point https://github.com/OpenFn/docs/issues/417

mtuchi:
I think we should close this issue and pickup our top 5 adaptor to start with 
cc @aleksa-krolls @josephjclark any thoughts ?

josephjclark:
Sure - do we know what the top 5 is yet?

Do we have a clear spec for what the docs should look like?

I suggest we don't take action until we can answer those two questions... maybe we need to pick up that common issue first

josephjclark:
I am going to close this issue down: we've broken out the top 5 into separate stories, and besides we've done a lot of docs work over the last six months

PR #176, opened 2022-11-18
`primero` remove examples from developer readme

aleksa-krolls:
examples should be in the inline adaptor documentation, not in the dev readme.

@mtuchi , can you discuss this with @taylordowns2000 and make sure that all examples and comments removed here are included in the inline adaptor documentation?


mtuchi:
@aleksa-krolls we have an issue that will fix the `jsdoc` #165 , this will include moving examples from readme to doc string 
I will chat with @taylordowns2000 on how we can prioritize this because there is a lot to be covered

mtuchi:
@taylordowns2000, I have added all examples to adaptors docstring now, you can review now

taylordowns2000 (dismissed):
@mtuchi , can you please:
1. Run a linter with our standard prettier settings on these examples. They look improperly spaced and line broken for readability 
2. Confirm that the new arguments are working with the primero v2 api? Are these actual working examples, and if so how did you verify them?

For number 2, this is a question we will keep coming back to over and over as we update the docs for various adaptors, so it's worth thinking through well.

mtuchi:
> @mtuchi , can you please:
> 
> 1. Run a linter with our standard prettier settings on these examples. They look improperly spaced and line broken for readability
> 2. Confirm that the new arguments are working with the primero v2 api? Are these actual working examples, and if so how did you verify them?
> 
> For number 2, this is a question we will keep coming back to over and over as we update the docs for various adaptors, so it's worth thinking through well.

I have update the examples that i could find from different primero repo @taylordowns2000

taylordowns2000 (approved):

PR #182, opened 2022-12-06
Add `convertCSVToJSON` helper function

elias-ba:
## Summary

This PR propose to add a helper function for converting and cleaning CSV rows to JSON object data.

## Details

In `packages/common` I have added a new function named `convertCSVToJSON`. It's purpose is to convert CSV rows to JSON string representation. The function does also clean some unusual invisible characters that implementation team have noticed in the CSV column names. In the current implementation that cleaning is done by default but I really think it should be parameterized and given as option to the helper function. Sometime we need to have more control on what the function does. But as this is just a proposal, I am happy to do that as improvement later. Depending on your opinions on the value of the function itself.

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/strong-kings-do.md
+++ b/.changeset/strong-kings-do.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-common': minor
+'@openfn/language-sftp': minor
+---
+
+Add convertCSVToJSON helper function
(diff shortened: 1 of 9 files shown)


mtuchi:
Closing this as there is a better implementation for `csv` parsing tracked on this issue #252

PR #185, opened 2022-12-15
Fix exports

mtuchi:
## Summary
Fixed exports for `big-query` and `commcare`  in `index.js`


josephjclark (approved):
Good catch with bigquery!

Issue #187, opened 2022-12-22
Adaptor tests will fail in node19


josephjclark:
Node 19 is getting stricter about ESM modules and removing the `--experimental-module-specifier` flag.

Here are some big changes coming to node19 which will catch us out:

* Relative imports MUST have a file extension (ie `import {} from './Adaptor` becomes `import {} from './Adaptor.js`)
* Folder imports will not work ((ie `import {} from './src` becomes `import {} from './src/index.js`)
* `lodash-fp`, which many adaptors depend on, is deprecated. Strictly speaking we should be using `lodash/fp`, although this isn't itself ESM compliant (see [lodash/issues/5285](https://github.com/lodash/lodash/issues/5285))

The upshot of all this is that tests and maybe builds will fail when we upgrade node beyond v18.

I actually thought the experimental arguments enabled future features early. Turns out they're actually masking deprecated functionality. Whoops - that's my bad.

Note that this doesn't affect any of the built and deployed adaptors - those are ESM and CJS compliant.

josephjclark:
Switching to `@swc/loader` should fix this - see https://github.com/OpenFn/kit/pull/833 for reference

PR #188, opened 2022-12-22
Start supporting node19

josephjclark:
## Summary

This PR updates common, salesforce and the migration guide allow support for node19+.

## Details

As explained in #187, node19 is going to bring big problems for our testing (and maybe building).

I've been through common and salesforce  and made them compatible with node19. This is really helpful for understanding the impact of the changes needed.

In summary:

* I've switched out `lodash-fp` for `lodash`, with tree-shakable, ESM compliant imports
* I've added extensions to import paths
* I've removed the redundant experimental command line args on unit tests
* I've updated the migration guide and tool

Note that build tooling (like babel) can help us with things like module extensions, but ESM modules don't support extension lookup and I think it's time we stopped pretending they do. It's not that big a deal. nodejs is hardening its ESM support and for the sake of clarity I think it's a good idea to follow along.

This all works fine on node18.

## Issues

#187 

## Review Checklist

      migration guide followed?


taylordowns2000 (commented):
@josephjclark , is this still a problem or should we close down this branch? (feels stale given the work that took place last week.)

PR #193, opened 2023-01-11
`kobotoolbox` Add `getDeploymentInfo` helper function

mtuchi:
## Summary

I added a `getDeploymentInfo` function which is available on version 2 of [kobotoolbox API](https://kf.kobotoolbox.org/api/v2/assets/)

## Details

The function retrieves information about the existing deployment if any.

## Review Checklist

      migration guide followed?


taylordowns2000 (approved):
beautiful

PR #194, opened 2023-01-11
Update `@constructor` to `@function`

mtuchi:
## Summary

Updated `@constructor` to `@function` for all adaptors functions that we're using @`constructor`. And also removed `/** @module Adaptor */` which was causing the documentation for the adaptor function to have `Adaptor.{function}`

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/cool-feet-bathe.md
+++ b/.changeset/cool-feet-bathe.md
@@ -1,0 +1,37 @@
+---
+'@openfn/language-asana': patch
+'@openfn/language-beyonic': patch
+'@openfn/language-bigquery': patch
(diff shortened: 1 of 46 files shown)


josephjclark (commented):
Just one comment which may or may not be an issue.

I'm in favour of this - I'm just not sure how to test it from the Lightning point of view :thinking: I guess we just need to release it and see - if everything breaks even more then we'll come back and take another look.

taylordowns2000 (approved):
@mtuchi , this is great. please release this and then merge to main once we confirm that it's working properly on Lightning.

PR #209, opened 2023-03-01
Use typedoc for magic lookup unit tests

josephjclark:
So we have a problem in #196 that we can't unit test magic lookups define in typescript files (ie, all the interesting stuff in dhis2).

This is pretty bad for devx because the model of being able to unit test directly against the jsdoc is gonna be really improtant.

This PR uses `typedoc` to parse `index.ts` and pull out any jsdoc annotations on object interfaces.

It basically works - there's a working unit test! - but I'm having some major drama in finding a config that works across all packages (including tools). Not going to document this in much detail just yet.

Basically when I call typedoc I want to pass in a tsconfig directly. But it stubbornly only accept a path to a tsconfig, which makes it hard to dynamically alter a config to suit a different package. There will be a solution but I need to hammer it around some more.

Still to do:

* [x] Sort out the the tsconfig thing (which is killing me)
*  [x] ~~Replace jsdoc2md  with typedoc ~~
* [x] Clear up the error/warning spam coming out of typedoc

### DIFF
--- a/packages/dhis2/package.json
+++ b/packages/dhis2/package.json
@@ -20,8 +20,8 @@
     "clean": "rimraf dist types docs",
     "integration-test": "mocha --experimental-specifier-resolution=node --no-warnings test/integration.cjs",
     "pack": "pnpm pack --pack-destination ../../dist",
-    "test:watch": "mocha -w --experimental-specifier-resolution=node --no-warnings \"**/*.test.js\"",
-    "test": "mocha --experimental-specifier-resolution=node --no-warnings \"test/**/*.test.js\""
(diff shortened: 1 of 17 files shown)


josephjclark:
After a bit of exploration I don't think typedoc will be able to parse js files for us.

Theoretically it should just parse what the ts compiler does, and there's an [old issue](https://github.com/TypeStrong/typedoc/issues/643) where this seems eminently possible. But the API has changed since then and having run some experiments and checked the docs, I'm not so sure.

I don't think this is worth spending loads of time on so I'm gonna merge this lot, as it actually works, and we can move on to something more valuable.

PR #211, opened 2023-03-13
Fix dhis2  `apiVersion` url build

mtuchi:
## Summary

`apiVersion` configuration was not properly built using in `DHIS2` adaptor, i updated the `buildUlr` utility function to properly build a valid DHIS2 API endpoint   

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/kind-candles-arrive.md
+++ b/.changeset/kind-candles-arrive.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-dhis2': patch
+---
+
+update buildUrl to handle api version
--- a/packages/dhis2/src/Utils.js
+++ b/packages/dhis2/src/Utils.js
@@ -23,8 +23,10 @@ export class Log {
 }
 
 export function buildUrl(urlString, hostUrl, apiVersion) {
-  const pathSuffix = apiVersion ? `/${apiVersion}${urlString}` : `${urlString}`;
-  return hostUrl + '/api' + pathSuffix;
+  const pathSuffix = apiVersion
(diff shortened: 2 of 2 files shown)


taylordowns2000:
Hey @mtuchi , I'm closing this as I think it's actually working the way we want. I could be wrong here, but the confusion is likely because of how various instances of dhis2 "play" are served.

When the URL is built, we want it to follow the pattern: `hostUrl` + `/api` + `/apiVersion` (optionally) + `/resourceUrl`.

What makes this confusing is that the `hostUrl` for "play instances" includes the **_DHIS2 version_** that they're hosting... which is _different_ than the `apiVersion` you're accessing in a particular request.

Examples:

1. _This_ is a valid URL for DHIS2: `https://play.dhis2.org/2.38.2.1/api/33/me`. ✅ 
2. And so is _this_: `https://play.dhis2.org/api/33/me` ✅ 
3. But _this_ is not: `https://play.dhis2.org/33/api/me` ❌ 

In example 1, the `hostUrl` is (confusingly) `https://play.dhis2.org/2.38.2.1`.

If you want to reach that URL without any `apiVersion` specified, you'd use `https://play.dhis2.org/2.38.2.1/api/me` - but again, that's _not_ specifying any particular `apiVersion`, it's just that the hostUrl includes some info about the dhis2 version.

Please feel free to re-open this if I'm not following or wrong about the desired behaviour. Here's the basic API documentation: https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-239/introduction.html#webapi_authentication

PR #217, opened 2023-03-21
`mssql` dependencies version bumps

mtuchi:
## Summary

I have updated the versions for `requests` and `common`

## Review Checklist

      migration guide followed?


mtuchi:
[Suggestion to change request dependency to superagent ](https://github.com/OpenFn/adaptors/issues/236) cc @taylordowns2000 @josephjclark

PR #227, opened 2023-03-30
Fix metadata Dependency

### LINKED ISSUE #226: Packages depend on @openfn/metadata
I'd gotten confused somewhere down the line - `tools/metadata` is a runtime dependency for magic adaptors. 

Because it's in `tools` I thought it was just dev/build apis, but some of the actual modelling APIs come from there.

So I need to work out what to do about that. It means preparing `tools/metadata` for publication and probably moving it out of tools.

### LINKED ISSUE #230: `salesforce` fails `./scripts/install-lp` beyond version 3.0.0
To reproduce the bug, run `./scripts/install-lp` in platform-app after adding back a salesforce version beyond `3.0.0`.

(To save some time, drop everything else from `adaptors.json`. Otherwise you'll need to install lots of adaptor-versions!)

josephjclark:
## Summary

An error on my part means that production adaptors have a dependency on an internal private package (tools/metadata). Which they should not have.

This causes the latest version of `dhis2` and `salesforce` to fail in platform.

This PR moves the metadata helper functions out of the private tools and into common. I had to update the build process to support this.

## Issues

Fixes #226 

Fixes #230 

## Review Checklist

      migration guide followed?


mtuchi:
@josephjclark do you know why the salesforce test is failing https://github.com/OpenFn/adaptors/actions/runs/4566965311/jobs/8060173456?pr=227#step:7:3562

josephjclark:
@mtuchi The test is now passing - this is ready for merge and release

PR #235, opened 2023-03-31
make postgres return relevant stuff in statedata

### LINKED ISSUE #231: Make postgres return relevant stuff in `state.data`
Today I found myself writing an _extra operation_ after a bunch of `language-postgres` operations:

```js
sql(() => 'select * from whatever;');
fn(state => ({ ...state, data: { result: state.response.body.rows } }));
```

Is it weird that postgres returns everything good inside `state.response.body.rows`?

I'd like us to keep all the cool metadata that comes back from the inner psql adaptor, but maybe make the rows show up in `state.data.result`. Or at least in `state.data.rows`?

The `response` thing is almost an anti-pattern now that @josephjclark is pushing us to strengthen the contract around `state.data.`

@mtuchi , I swear you've made a very similar suggestion to this a while ago, but I can't find the issue anywhere. What do you and Joe think?

mtuchi:
## Summary

Add `callback` as an option in `sql()` function and update the response structure for the results

## Issues

Fixes #231 

## Review Checklist

      migration guide followed?


taylordowns2000 (changes requested):
this looks great but needs to be implemented for all the functions here. if we're doing the MAJOR version bump, we should adjust the API for all of these at the same time.

taylordowns2000 (approved):
this looks great @mtuchi . please add a comment below to confirm that (1) all these new signatures have been tested and (2) that it's possible for the user to access the full response from the DB adaptor _in the callback_ so that they can pick out extra bits for use in `state.data` as they require.

taylordowns2000:
From @mtuchi : i haven't been able to test,  the following functions
-  findValue
-  upsert
-  upserIf
-  upsertMany

But they all use a function called queryHandler()  and that is the function that i modified to handle callbacks() and return response to state.data

mtuchi:
@taylordowns2000 
for `findValue` since it returns value only, It doesn't need a callback, So I have updated the `jsdoc string` to reflect that. Below is the code i used to test `findValue`
```js
fn(async state => {
  const findVal = await findValue({
    uuid: 'id',
    relation: 'teams',
    where: { email: 'red-team@openfn.org' },
    operator: { email: 'like' },
  })(state);
  console.log(findVal);
  return state;
});
```
- For `upsert()`, `upsertIf()` and `upsertMany()` The callback is working but since we are updating a record in the database, the response is empty
```js
upsert(
  'patient', // the DB table
  'patient_id', // a DB column with a unique constraint OR a CONSTRAINT NAME
  {
    patient_name: 'Jane',
    patient_id: 9,
  },
  {
    setNull: ["''", "'undefined'"],
    writeSql: true,
    execute: true,
    logValues: true,
  },
  state => {
    console.log(state.data, 'This is response');
    console.log('This is a callback');

    return state;
  }
);


upsertIf(
  'Jane',
  'patient', // the DB table
  'patient_id', // a DB column with a unique constraint OR a CONSTRAINT NAME
  {
    patient_name: 'Eloide',
    patient_id: 9,
  },
  {
    setNull: ["''", "'undefined'"],
    writeSql: true,
    execute: true,
    logValues: true,
  },
  state => {
    console.log('This is a callback');

    return state;
  }
);


upsertMany(
  'teams',
  'email',
  [
    { name: 'One Team', email: 'one@openfn.org' },
    { name: 'Team two', email: 'two@openfn.org' },
  ],
  { logValues: true },
  state => {
    console.log('Yes it worked', state.data);

    return state;
  }
);
```

mtuchi:
@taylordowns2000 for (2), Yes you can access the full response in a callback by `state.response`

Issue #237, opened 2023-04-04
Randomly failing salesforce tests


josephjclark:
Salesforce metadata tests seem to intermittently fail, as reported by @mtuchi 

This seemed stable to me until I zapped and reinstaleld node modules. Now tests seem to fail occasionally at random.

I did make a curious fix the other day against the metadata tests - I wonder if that's related to whatever this is :thinking:

PR #238, opened 2023-04-04
Salesforce: stabilise metadata tests

josephjclark:
Some salesforce metadata tests are failing intermittently.

Since the tests run on the same fixture data, I can only think that the fails occur because the array ordering comes out differently.

So I've made the tests more robust by not relying on array order - which can only be an improvement, even if the actual issue isn't resolved.

@mtuchi this doesn't address the fails in your loom video, but I still think that's an install issue(the `jsonpath` module is not installed).

It looks OK on my machine :tm:  but I'd appreciate some more QA. I'll re-run CI a few times.

## Issues

#237

## Review Checklist

### DIFF
--- a/packages/salesforce/test/metadata/metadata.test.js
+++ b/packages/salesforce/test/metadata/metadata.test.js
@@ -26,7 +26,7 @@ describe('metadata function', () => {
       mockHelper
     );
 
-    const [asset] = model.children;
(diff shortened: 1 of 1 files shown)


taylordowns2000:
@mtuchi , this is 10/10 for me. going to merge for now and close the issue. if it pops up on CI we re-open?
image

PR #249, opened 2023-05-11
Add contributing section to README

amberrignell:
## Summary

Add a high-level, single-sentence summary of what this PR changes.

## Details

Add technical details of what you've changed (and why).

## Issues

Reference any related issues here.

## Review Checklist

Before merging, the reviewer should check the following items:

- [ ] Does the PR do what it claims to do?
- [ ] If this is a new adaptor, has the migration tool been run and the
      migration guide followed?
- [ ] Are there any unit tests? Should there be?
- [ ] Is there a changeset associated with this PR? Should there be? Note that
      dev only changes don't need a changeset.

PR #259, opened 2023-05-25
Update template

josephjclark:
Quick pass over the template to update some things that were out of date.

To be honest the template looks in pretty good shape to me? I've made some adjustments but I think its generally fine.

This is probably the best tested adaptor in the repo, by the way :sweat_smile:


amberrignell:
Thanks @joe! cc @taylordowns2000 once this is merged if you could also review and merge https://github.com/OpenFn/adaptors/pull/249 that would be great :) (no rush)

taylordowns2000 (approved):

PR #260, opened 2023-05-29
Expand adaptor docs

amberrignell:
## Summary

## Details

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,14 @@ please make one and assign yourself.
 #### 1. Setup
 
 - Create a copy of the adaptor [template](packages/template), and rename it.
-  `cp -R packages/template packages/adaptorname`
+  `cp -R packages/template packages/youradaptorname`
 
 - Open package.json, and update the `name`, `description` and `build` from
-  `template` to your `newadaptorname`. Update the version.
+  `template` to your `youradaptorname`. Update the version.
 
-- Update the changelog heading `# @openfn/language-template` to your adaptor
-  name and delete the example content. Delete ast.json, then run `pnpm clean`
-  (this will get rid of the dist, types and docs folders).
+- Update the top-level changelog comment `# @openfn/language-template` to your
+  adaptor name and delete the example content. Delete ast.json, then run
+  `pnpm clean` (this will remove the dist, types and docs folders).
 
 - Update your adaptor's README.md to use your new adaptor name (we'll come back
   to this later to update the sample operation)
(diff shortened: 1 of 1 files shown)


taylordowns2000 (changes requested):
hey was the line break on line 65 intentional? please remove if not. (there still be something going on with your formatter and the prettierrc configuration.)

README.md
```diff
@@ -62,24 +62,84 @@ please make one and assign yourself.
...
 
 - Create a copy of the adaptor [template](packages/template), and rename it.
-  `cp packages/template packages/newadaptorname`
+  `cp -R packages/template packages/youradaptorname`
 
-- Define the configuration schema
+- Open package.json, and update the `name`, `description` and `build` from `template` to your `youradaptorname`. Update the version.
 
-  The configuration schema defines required parameters and their expected values
+- Update the top-level changelog comment `# @openfn/language-template` to your adaptor name and delete the example content. Delete ast.json, then run `pnpm clean` (this will remove the dist, types and docs folders).
```
josephjclark, line 76: We probably still want the `language`- prefix so can we say:
```
Update the top-level changelog comment `# @openfn/language-template` to `# @openfn/language-youradaptorname`
```

README.md
```diff
@@ -90,13 +150,47 @@ To create a new adaptor follow the steps below
...
   [Here's an example of a configuration schema](packages/template/configuration-schema.json)
 
-- To add or update functions to an existing adaptor:
 
-  - Create or update adaptor functions.
+  #### 4. Test it manually 
+  - Make sure to run build again
+  - Run your job `openfn tmp/job.js -O -ma youradaptorname`
+  - Make sure it returns what you expect
+  - You can delete this folder once you have finished testing your job manually
```
josephjclark, line 158: Which folder? :thinking: 

If you mean `tmp`, then yes technically it can be removed but it's usually useful to keep the job code for later.

README.md
```diff
@@ -62,24 +62,84 @@ please make one and assign yourself.
...
+- Open package.json, and update the `name`, `description` and `build` from `template` to your `youradaptorname`. Update the version.
 
-  The configuration schema defines required parameters and their expected values
+- Update the top-level changelog comment `# @openfn/language-template` to your adaptor name and delete the example content. Delete ast.json, then run `pnpm clean` (this will remove the dist, types and docs folders). 
+
+- Update your adaptor's README.md to use your new adaptor name (we'll come back to this later to update the sample operation)
+
+#### 2. Create a job with the operation you'd like to add, and run it so it fails
+
+- Create a _temporary_ folder where you can test your operation manually by running a job: `mkdir tmp`
```
josephjclark, line 82: I wouldn't emphasis _temporary_, it makes it look like you have to do something special. Like, how do I create a  temporary folder?

Probably important to note that the contents of `tmp` don't get checked in to github - so this folder name isn't arbitrary, it's a convention we use.

josephjclark (commented):
I've made a couple of comments.

It would be really nice to have a monorepo tool to initialise a new adaptor using the template. It could do most of the monkeywork of coping files around and swapping out adaptor names. That would simplify a lot of this stuff!

README.md
taylordowns2000: Ah this because folks often keep secrets in a tmp folder. I think the pattern we want to encourage is to make sure that people keep their local convenience folder (like tmp) _out_ of git.

josephjclark: Yes, keeping sensitive stuff out of git is an even better pattern. We could consider re-writing all this to use a path like `~/openfn/jobs/adaptorname` or something.

Another thought - these are very unixy paths and commands we're using - are we alienating windows users?

josephjclark:
I am merging this in for today's release. It's an improvement on the current docs and we can raise issues/PRs later if we want to make further changes.

Thanks @amberrignell !!

PR #270, opened 2023-06-05
`ocl` improvements

mtuchi:
## Implementation Plan
Based on the `OCL` API structure, i envision the following helper function and a configuration-schema

```
// https://api.openconceptlab.org/orgs/MSFOCG/collections/lime-demo/concepts/
// https://api.openconceptlab.org/orgs/MSFOCG/sources/DHIS2DataElements/concepts/
// https://api.openconceptlab.org/orgs/MSFOCG/sources/OpenMRS/mappings/
// https://api.openconceptlab.org/orgs/MSFOCG/collections/lime-demo/references/
```
`getSource(path, query, callback);` , `getCollection(path, query, callback);`
Eg's: 
- getSource("DHIS2DataElements/concepts/");
- getSource("DHIS2DataElements/mappings/");
- getCollection('lime-demo/mappings');
- getCollection('lime-demo/concepts');
- getCollection('lime-demo/references');

Configuration 
```
{
  "baseUrl": "https:api.openconceptlab.org",
  "owner": "MSFOCG"
}
```
## Details

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/tame-eyes-march.md
+++ b/.changeset/tame-eyes-march.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-ocl': minor
+---
+
+Add get and getMappings functionwq
--- a/.changeset/tame-eyes-march.md
+++ b/.changeset/tame-eyes-march.md
@@ -2,4 +2,4 @@
 '@openfn/language-ocl': minor
 ---
 
-Add get and getMappings functionwq
+Add `get()` and `getMappings()` function
--- a/packages/ocl/src/Adaptor.js
+++ b/packages/ocl/src/Adaptor.js
@@ -2,7 +2,7 @@ import {
   execute as commonExecute,
   expandReferences,
 } from '@openfn/language-common';
-import { buildUrl, handleResponse, request } from './Util';
+import { buildMappingsUrl, handleResponse, request } from './Util';
 
 /**
  * Execute a sequence of operations.
@@ -135,6 +135,8 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
     repositoryId = expandReferences(repositoryId)(state);
     options = expandReferences(options)(state);
 
+    const { baseUrl } = state.configuration;
+
     const defaultOptions = {
       ownerId: ownerId,
       repositoryId: repositoryId,
@@ -146,8 +148,11 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
       content: 'mappings',
     };
 
-    const optionsMerge = { ...defaultOptions, ...options };
-    const { url, query } = buildUrl(state.configuration, optionsMerge);
+    const urlParams = { ...defaultOptions, ...options };
+    const { url, query } = buildMappingsUrl({
+      baseUrl: baseUrl,
+      ...urlParams,
+    });
 
     return request(url, query).then(response =>
       handleResponse(response, state, callback)
@@ -158,27 +163,6 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
 /**
  * Get a resource in OCL
  * @public
- * @example
- * get(
- *   "mappings",
- *   {
- *     ownerId: "MSFOCG",
- *     ownerType: "users", // Default to orgs
- *     repository: "collections",
- *     repositoryId: "lime-demo",
- *     version: "HEAD",
- *     page: 1,
- *     exact_match: "off",
- *     limit: 200,
- *     verbose: false,
- *     sortDesc: "_score",
- *   },
- *   (state) => {
- *     // Add state oclMappings
- *     const oclMappings = state.data;
- *     return { ...state, data: {}, references: [], response: {}, oclMappings };
- *   }
- * );
  * @example
  *  get(
  *   "orgs/MSFOCG/collections/lime-demo/HEAD/mappings",
@@ -203,17 +187,13 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
  */
 export function get(path, options, callback = false) {
   return state => {
-    const defaultOptions = {
-      content: path,
-    };
-
     path = expandReferences(path)(state);
     options = expandReferences(options)(state);
+    const { baseUrl } = state.configuration;
 
-    const optionsMerge = { ...defaultOptions, ...options };
-    const { url, query } = buildUrl(state.configuration, optionsMerge);
+    const url = `${baseUrl}/${path}`;
 
-    return request(url, query).then(response =>
+    return request(url, options).then(response =>
       handleResponse(response, state, callback)
     );
   };
--- a/packages/ocl/src/Adaptor.js
+++ b/packages/ocl/src/Adaptor.js
@@ -3,7 +3,7 @@ import {
   expandReferences,
   http,
 } from '@openfn/language-common';
-import { buildUrl, handleResponse, handleError } from './Util';
+import { buildUrl, handleResponse, handleError, request } from './Util';
 
 /**
  * Execute a sequence of operations.
@@ -142,14 +142,15 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
       ownerType: 'orgs', // Default to orgs | orgs or users
       repository: 'collections', // Default to collections, collections or sources
       version: 'HEAD', // Default to HEAD, Eg: HEAD, 0.04
+      expansions: 'expansions',
+      expansionId: 'autoexpand-HEAD',
       content: 'mappings',
     };
 
     const optionsMerge = { ...defaultOptions, ...options };
-    const urlConfig = buildUrl(state.configuration, optionsMerge);
+    const { url, query } = buildUrl(state.configuration, optionsMerge);
 
-    return http
-      .get(urlConfig)(state)
+    return request(url, query)
       .then(response => handleResponse(response, state, callback))
       .catch(handleError);
   };
@@ -204,7 +205,6 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
 export function get(path, options, callback = false) {
   return state => {
     const defaultOptions = {
-      ownerType: 'orgs', // Default to orgs | orgs or users
       content: path,
     };
 
@@ -212,10 +212,9 @@ export function get(path, options, callback = false) {
     options = expandReferences(options)(state);
 
     const optionsMerge = { ...defaultOptions, ...options };
-    const urlConfig = buildUrl(state.configuration, optionsMerge);
+    const { url, query } = buildUrl(state.configuration, optionsMerge);
 
-    return http
-      .get(urlConfig)(state)
+    return request(url, query)
       .then(response => handleResponse(response, state, callback))
       .catch(handleError);
   };
--- a/packages/ocl/src/Adaptor.js
+++ b/packages/ocl/src/Adaptor.js
@@ -138,13 +138,11 @@ export function getMappings(ownerId, repositoryId, options, callback = false) {
     const { baseUrl } = state.configuration;
 
     const defaultOptions = {
-      ownerId: ownerId,
-      repositoryId: repositoryId,
       ownerType: 'orgs', // Default to orgs | orgs or users
+      ownerId: ownerId,
       repository: 'collections', // Default to collections, collections or sources
+      repositoryId: repositoryId,
       version: 'HEAD', // Default to HEAD, Eg: HEAD, 0.04
-      expansions: 'expansions',
-      expansionId: 'autoexpand-HEAD',
       content: 'mappings',
     };
 
--- a/packages/ocl/src/Adaptor.js
+++ b/packages/ocl/src/Adaptor.js
@@ -1,12 +1,12 @@
 import {
   execute as commonExecute,
-  composeNextState,
   expandReferences,
   http,
 } from '@openfn/language-common';
+import { Log, buildUrl, handleResponse, handleError } from './Util';
 
-const { axios } = http;
-export { axios };
+// const { axios } = http;
+// export { axios };
 
 /**
  * Execute a sequence of operations.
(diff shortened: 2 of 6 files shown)


mtuchi:
@taylordowns2000 see my suggestion for the implementation plan above ☝🏽, We can sync tomorrow in the AM if you have suggestion

taylordowns2000:
image

mtuchi:
@taylordowns2000 new take on the helper function

Instead of having a `get(path, query, callback)` function that will take `path | Object` , `query | Object | Optional` and `callback | Function | Optional` Eg👇🏽 
```
get({
  owner: 'mtuchi', // User or Org Id
  repository: 'repository/repositoryId', //Eg :sources/sourceId or collections/collectionId
  content: 'content/contentId', // Eg: mappings/mappingId or concepts/conceptId
});
```

We can have two functions `getSource(path, query, callback)` and `getCollection(path, query, callback)`.
Where `path | Object` , `query | Object | Optional` and `callback | Function | Optional` 

```
getSource({
  source: 'sourceId',
  owner: 'mtuchi',
  version: '',
  content: 'mappings/1',
});
```

```
getCollection(
  {
    owner: 'mtuchi',
    collection: 'collectionId',
    version: '', // Optional collectionVersion
    expansion: '', // Optional expansionID https://docs.openconceptlab.org/en/latest/oclapi/apireference/expansions.html#get-an-expansion-for-a-collection-version
    content: 'mappings/1',
  },
  query, // Query object
  callback() // Callback fn
);
```

behind the scene of the two function above, there is a util function `buildUrl(path, query)`. This function is responsible for building the `url` for our `get()` request

```
// Real example
// baseUrl: "https://api.openconceptlab.org"
// owner: "MSFOCG"
// collection: "lime-demo"
// version: "HEAD",
// expansion: "autoexpand-HEAD"

getCollection(
  {
    owner: 'MSFOCG',
    collection: 'lime-demo',
    version: 'HEAD',
    expansion: 'autoexpand-HEAD',
    content: 'mappings',
  },
  {
    page: 1,
    exact_match: 'off',
    limit: 200,
    verbose: false,
    fromConceptOwner: 'MSFOCG',
    toConceptOwner: 'MSFOCG',
    toConceptSource: 'DHIS2DataElements',
    sortDesc: '_score',
  } // Query parameters
);
```

mtuchi:
Todo's 

- [ ] Util function for building url `buildUrl(path, query)`
- [ ] Util function for handling response `handleResponse(response, state, callback)`
- [ ] Util function for handling errors `handleError(error)`
- [ ] Add `getSource(path, query, callback = false)`
   - [ ] add in `docstring` path parameters
   - [ ] add in `docstring` query parameters
- [ ] Add `getCollection(path, query, callback = false)`
   - [ ]  add in `docstring` path parameters
   - [ ] add in `docstring` query parameters

mtuchi:
```
getSource('mtuchi', 'sourceId', { }, callback())
getCollection('mtuchi', 'collectionId', { }, callback())
```

josephjclark:
Swagger API: https://api.openconceptlab.org/swagger/

mtuchi:
@taylordowns2000 , i had a sync with @josephjclark , We have a new direction 😬 ,  Since at the moment we only need to get mappings from `ocl` , to reduce the complexity of having so many task specific function like `getSource()` and `getCollection` 
We are going to create only two function `get(path, options, callback)` and `getMappings(owner, repositoryId, options, callback)`

```
https://api.openconceptlab.org/users/mtuchi/sources/1/mappings/1/
getMappings(owner, repositoryId, options, callback)

getMappings("MSFOCG", "lime-demo", {
    page: 1,
})

We will have default options
options: {
    owner_type: "orgs", // Default to orgs 
    repository: "collections", // Default  collections
    version: "HEAD", // Default to HEAD
    page: ""
}
```

packages/ocl/src/Adaptor.js
```diff
@@ -1,13 +1,12 @@
...
   execute as commonExecute,
-  composeNextState,
   expandReferences,
   http,
 } from '@openfn/language-common';
+import { Log, buildUrl, handleResponse, handleError } from './Util';
 
-const { axios } = http;
-export { axios };
+// const { axios } = http;
```
josephjclark, line 8: This ought to be removed, rather than uncommented

packages/ocl/src/Adaptor.js
```diff
@@ -110,82 +109,119 @@ function getMappingInfo(params) {
...
+ *   "lime-demo",
+ *   { page: 1, exact_match: "off", verbose: false },
+ *   (state) => {
+ *     // Add state oclMappings
+ *     const oclMappings = state.data;
+ *     return { ...state, data: {}, references: [], response: {}, oclMappings };
+ *   }
+ * );
+ * @function
+ * @param {string} owner - An OCL user or organization
```
josephjclark, line 127: I think this is `ownerId`? Or `ownerName`? Just helps differentiate from some kind of owner object

josephjclark (commented):
Saw this flash past and I thought I'd leave some comments.

The API feels great to me - I like the new function and feel good about the "escape hatch" of the general getter.

I'd double check the docs and I still think the logging is over-complicated tbh

packages/ocl/src/Adaptor.js
```diff
@@ -110,82 +109,119 @@ function getMappingInfo(params) {
...
+
+    owner = expandReferences(owner)(state);
+    repositoryId = expandReferences(repositoryId)(state);
+    options = expandReferences(options)(state);
+
+    const optionsMerge = { ...defaultOptions, ...options };
+    const urlConfig = buildUrl(state.configuration, optionsMerge);
+
+    return http
+      .get(urlConfig)(state)
```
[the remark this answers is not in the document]
taylordowns2000: Yes, this is the standard practice. But... looking back on it after a couple of years, I feel as though it's _not_ necessarily a helpful abstraction. (Is it _that hard_ to use fetch? No, probably not. But back when this was designed, i think we wanted all REST-based adaptors to import and use the `http` module from common.)

I'd vote to use regular old node fetch in here and take this as the way-to-go from here on out. from now on, whenever we write or update an adaptor we replace the call to `http.blah` with `fetch`.

packages/ocl/src/Util.js
```diff
@@ -0,0 +1,113 @@
+import { composeNextState } from '@openfn/language-common';
+
+export const Log = (() => {
```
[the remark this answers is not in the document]
taylordowns2000: Yes please drop this entirely.

.changeset/tame-eyes-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-ocl': minor
+---
+
+Add get and getMappings functionwq
```
taylordowns2000, line 5: typo

taylordowns2000 (changes requested):
thanks, @mtuchi !
1. please address joe's changes (and my comments).
2. i'm not sure i follow the "having so many task specific functions" logic. are there ever going to be more than 2? (`getMappings` and `getConcepts`).

to determine if the naming of `getMappings` should be singular or plural, i'll need to see some working code and the response. can you paste an example in Slack tomorrow?

packages/ocl/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: I am happy to hear this, I think it's the right move.

We may want to push some helper code down into common but I suggest we look at this later.

I've added a note to this effect to the best practice doc https://github.com/OpenFn/adaptors/wiki/Job-Writing-Best-Practice

packages/ocl/src/Util.js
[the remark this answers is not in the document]
mtuchi: I am dropping `log` function but I still think we need this available, in all of our adaptors. Because all adaptors should have the same log pattern. Also i think we need a global `Util.js` that can be accessed with any of the adaptor. The file can have generic functions that we don't necessary need them in common, Function like `isObjectEmpty()` or `Log`, 
```
export const log = {
  success: message => console.log(`✓ Success at ${new Date()}: ${message}`),
  warn: message => console.log(`⚠ Warning at ${new Date()}: ${message}`),
  error: message => console.log(`✗ Error at ${new Date()}: ${message}`),
  info: message => console.log(`ℹ Info at ${new Date()}: ${message}`),
};
```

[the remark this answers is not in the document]
josephjclark: _Should_ all our adaptors have the same log pattern? I am not so sure actually.

Timestamps will be added by lightning. If you want timestamps in local, CLI I can add functionality to the logger (so that ALL logs have timestamps). I am not sure of the value of this, and expect this is something you can add to your shell.

If we DO want a standard log pattern in the adaptors, we should be importing `@openfn/logger - because that's a true standard and should give you all the options you need.

Any utility functions used by all adaptors should be exported by common.

packages/ocl/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: I tried to use fetch when i was working with openMRS, i realize it still lack some of the features were familiar with when it comes to error handling, that's why i ended up using `superagent`, I think we either implement a wrapper that will make it easy to use `fetch` or we use a package that uses fetch behind the scene.

mtuchi:
> thanks, @mtuchi !
> 
> 1. please address joe's changes (and my comments).
> 2. i'm not sure i follow the "having so many task specific functions" logic. are there ever going to be more than 2? (`getMappings` and `getConcepts`).
> 
> to determine if the naming of `getMappings` should be singular or plural, i'll need to see some working code and the response. can you paste an example in Slack tomorrow?

@taylordowns2000 
1. I have address joe's comment, 
2. Ideally we envision the adaptor to have functions like `getSource()`, `getCollection()`, `createCollection()`, `createSource()` and more. But we don't want to create these helper functions because there is more than one way for doing these operations ([See swagger API](https://api.openconceptlab.org/swagger/)). Hence why we slowly want to introduce each function upon request. 

Regarding naming conventions, The function `getMappings()` return `mappings list of a collectionID` that's why the 's
For-example in this 👇🏽 , `getMappings("mtuchi","1",{"owner_type": "users", repository: "sources"})`
Screenshot 2023-06-08 at 9 12 09 AM

mtuchi:
@taylordowns2000 i have implemented the adaptor to work with `fetch()`, please review and test with the following jobs 👇🏽 
**Configuration** : `{ "baseUrl": "https://api.openconceptlab.org" }`

**Test jobs**
```
// getMappings(
//   'MSFOCG',
//   'lime-demo',
//   {
//     page: 1,
//     limit: 1,
//     verbose: false,
//     fromConceptOwner: 'MSFOCG',
//     toConceptOwner: 'MSFOCG',
//     toConceptSource: 'DHIS2DataElements',
//     sortDesc: '_score',
//   },
//   state => {
//     // Add state oclMappings
//     const oclMappings = state.data;
//     return { ...state, data: {}, references: [], response: {}, oclMappings };
//   }
// );

// get(
//   'mappings',
//   {
//     ownerId: 'MSFOCG',
//     repository: 'collections',
//     repositoryId: 'lime-demo',
//     version: 'HEAD',
//     page: 1,
//     exact_match: 'off',
//     limit: 1000,
//     verbose: false,
//     fromConceptOwner: 'MSFOCG',
//     toConceptOwner: 'MSFOCG',
//     toConceptSource: 'DHIS2DataElements',
//     sortDesc: '_score',
//   },
//   state => {
//     // Add state oclMappings
//     const oclMappings = state.data;
//     return { ...state, data: {}, references: [], response: {}, oclMappings };
//   }
// );

// get(
//   'orgs/MSFOCG/collections/lime-demo/HEAD/mappings',
//   {
//     page: 1,
//     exact_match: 'off',
//     limit: 1000,
//     verbose: false,
//     fromConceptOwner: 'MSFOCG',
//     toConceptOwner: 'MSFOCG',
//     toConceptSource: 'DHIS2DataElements',
//     sortDesc: '_score',
//   },
//   state => {
//     // Add state oclMappings
//     const oclMappings = state.data;
//     return { ...state, data: {}, references: [], response: {}, oclMappings };
//   }
// );

get(
  'orgs/MSFOCG/collections/lime-demo/HEAD/mappings',
  {
    query: {
      page: 1,
      exact_match: 'off',
      limit: 1000,
      verbose: false,
      fromConceptOwner: 'MSFOCG',
      toConceptOwner: 'MSFOCG',
      toConceptSource: 'DHIS2DataElements',
      sortDesc: '_score',
    },
  },
  state => {
    // Add state oclMappings
    const oclMappings = state.data;
    return { ...state, data: {}, references: [], response: {}, oclMappings };
  }
);
```

packages/ocl/src/Util.js
```diff
@@ -0,0 +1,106 @@
...
+  const nextState = {
+    ...composeNextState(state, response),
+    response,
+  };
+  if (callback) return callback(nextState);
+  return nextState;
+}
+
+export function handleError(error) {
+  throw error;
```
taylordowns2000, line 53: what's the purpose of `.catch(e => throw e)` ?

packages/ocl/src/Util.js
```diff
@@ -0,0 +1,106 @@
...
+  const responseString = [
+    `Request: ${method} "${url}"`,
+    `Status: ${status}`,
+    `Retrieved "${data.length}" data`,
+  ].join('\n∟ ');
+
+  throw new Error(`at ${new Date()} \n ${responseString}`);
+}
+
+export function handleResponseError(response, method) {
```
taylordowns2000, line 67: this feels significantly more complex than the example we agreed on earlier. i'm OK with it if there's some important reason. what's the benefit of creating two different functions here?

packages/ocl/src/Util.js
```diff
@@ -0,0 +1,106 @@
+import { composeNextState } from '@openfn/language-common';
+
+export function buildUrl(configuration, options) {
```
taylordowns2000, line 3: i'm skeptical of the value of this. seems like a lot of code and i'd prefer to see string interpolation.

packages/ocl/src/Util.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  try {
+    const response = await fetch(url, options);
+    console.log(url);
+    handleResponseError(response, method);
+    const data = await response.json();
+
+    if (data.length === 0)
+      handleEmptyResponse({ method, url, data, status: response.status });
```
taylordowns2000, line 101: is handleEmptyResponse used more than once in this adaptor? if not, keep it inline and use string interpolation. it's so much harder to figure out what's going on if it's tucked into a different function and uses all those extra lines of code.

taylordowns2000 (changes requested):
hey @mtuchi , this is getting there! it seems like there's a bit of extra code and a general tendency towards "over-factoring" going on. creating a bunch of additional functions and writing more lines of code than necessary when there may be a more readable (and less verbose) way of writing stuff. please address each of my 4 comments inline and we'll discuss the next steps on Monday

packages/ocl/src/Util.js
mtuchi: The idea was `handleError()` sometimes have `error.response` object. If that happens you might want to do something like this 
```
 if (error.response) {
    const { method, path } = error.response.request;
    const { status } = error.response;
    if (Object.keys(error.response.data).length === 0) {
      throw new Error(
        `Server responded with:  \n${JSON.stringify(error.response, null, 2)}`
      );
    }
 ```
 Basically any custom error handling should happen in `handleError()` function. Since we don't have that use case yet, i am replacing this `handleError()` function with a simple `.catch(e => {throw e})`

File: packages/ocl/src/Util.js (hunk 7 above)
Comment (mtuchi) [reply packages/ocl/src/Util.js]: `handleEmptyResponse()` is for `status: 200`, `result: 0`, i couldn't get the `results` from `response`.  https://developer.mozilla.org/en-US/docs/Web/API/Response. The body returned in `response` object is a readable stream not a result object. 

I have modified the `handleResponseError(response,results, method)` that way i don't need `hanldeEmptyReponse()` function

File: packages/ocl/src/Util.js (hunk 9 above)
Comment (mtuchi) [reply packages/ocl/src/Util.js]: I removed this function and updated the `handleResponseError()` to handle empty response

File: packages/ocl/src/Util.js (hunk 8 above)
Comment (mtuchi) [reply packages/ocl/src/Util.js:3]: string interpolation will have one huge string
```
const url = `${baseUrl}/${ownerType}/${ownerId}/${repository}/${repositoryId}/${version}/${expansions}/${expansionId}/${content}`;
```
 and all these values 👇🏽   are optional, so if a user does not pass them you have to write a logic that will remove `undefined` values from the string 
 ```
ownerId,
ownerType,
repository,
repositoryId,
version,
expansions,
expansionId,
```

The current implementation is readable and cleans up `undefined` optional values.
So i would vote for `const url = urlParts.join('/')` and not `const url = ${....}`

taylordowns2000 (changes requested):
1. please make expansions (both fields) optional (leave version defaulting to HEAD for now.)
then all good!

mtuchi:
> 1. please make expansions (both fields) optional (leave version defaulting to HEAD for now.)
>    then all good!

@taylordowns2000 this is ready for review now

taylordowns2000 (approved):

PR #280, opened 2023-06-16
add linting, fix reassignment issue

### LINKED ISSUE #275: expandReferences is broken in dhis2.get (and other adaptors)
This is huge. And hard to explain.

If nested in an `each`, then `dhis2.get` will only resolve references on the first iteration.

So this code breaks:
```
each('patients[*]', get('history', (state) => ({ query: { id: state.data.id } }) )
```
What this SHOULD do is for each item in `state.patients`, resolve the query object using the reference function and fetch each patient.

But actually, it will only resolve the query object once, for the first patient, and thereafter re-use the same query object and return the same patient `patients.length` times.

The problem is that the get function expands references back into the closure variable, mutating it for the next call.
```
export function get(resourceType, query, options = {}, callback = false) {
  return state => {
    console.log('Preparing get operation...');

    resourceType = expandReferences(resourceType)(state);
    query = expandReferences(query)(state);
    options = expandReferences(options)(state);
```
This means that on the first invocation of `get`, the query is a function. It will be called and its value returned to `query`. So on the second invocation, the query is a string, and its value will be re-used.

Other adaptors expand to a new variable and leave the closure alone, which is correct and good:

export function get(resourceType, query, options = {}, callback = false) {
  return state => {
    console.log('Preparing get operation...');

    const resolvedResourceType = expandReferences(resourceType)(state);
    const resolvedQuery = expandReferences(query)(state);
    const resolvedOptions = expandReferences(options)(state);
```

This problem forced @mtuchi to fall back to `fn()` and do all his iteration manually in vanilla JS.

We've identified this problematic pattern on other adaptor functions as well - they all need fixing. The fix is so significant it should cause a minor or even major version bump.

### LINKED ISSUE #279: Standardize adaptors with eslint
Share an eslint configuration across the whole monorepo, detect/prevent antipatterns

taylordowns2000:
## Summary

This adds linting and fixes param reassignment, closing #279 and fixing #275 

## Details

Added esLint with a global config that runs on `pnpm build`

## Issues

fixes #279 
fixes #275 

## Review Checklist

- [x] If this is a new adaptor, has the migration tool been run and the
      migration guide followed?


taylordowns2000:
@josephjclark , FYI, this looks like it will work quite nicely. i will keep you posted

taylordowns2000:
@josephjclark , i'd appreciate a few minutes of your time to check out the pattern here. seems to do the trick.

@mtuchi , you'll notice that `pnpm lint` (and now, `pnpm build`) fails for 3 adaptors. I'm hoping you can take a look at _why_ and try to re-write those three sections without using reassignment.

mtuchi:
@taylordowns2000 i have addressed the `eslint` errors for `sftp`, `http` and `ocl`

josephjclark (changes requested):
Looks great. I appreciate the quicck return on this.

There's a TODO that should be TODONE and I would recommend warning, not erroring, for `no-unused-vars`. See the review comments.

Is this branched from the latest main? There are a couple of strange diffs, eg in `.changeset` and `bigquery`.

Also any adaptors with user-facing changes - such as this bugfix - need a changeset. I think it should be a least a minor bump on any fixed adaptor.

PR #285, opened 2023-06-20
New expandReferences API

### LINKED ISSUE #264: Review expandAdaptors
Adaptors functions should try to "expand" all of their arguments, like this:

```
  return state => {
    const resolvedPath = expandReferences(path)(state);
    const resolvedParams = http.expandRequestReferences(params)(state);

    // ....
```
This is really important for enabling operations to be nicely composed, like this:
```
each(
 'patients',
  upsert('trackedEntityInstance', state => ({
    name: state.data.name,
    age: state.data.age
  }));
});
```

It turns out that at least one vitally important function, `dhis2.upsert`, does NOT expand its references. So actually the code I've pasted above will not work.

We should fix this ASAP, and probably investigate other commonly used adaptors.

Here's a list of currently used adaptors (provided by Mtuchi):
```
[
    "asana-v1.0.1",
    "commcare-",
    "dhis2-",
    "godata-",
    "googlesheets-",
    "http-",
    "kobotoolbox-",
    "mailchimp-",
    "mailgun-",
    "mssql-",
    "openfn-",
    "openmrs-v1.0.1",
    "postgresql-",
    "primero-v2.10.0",
    "rapidpro-v0.4.7",
    "salesforce-",
    "sftp-",
    "surveycto-",
    "telerivet-"
]
```

josephjclark:
## Summary

I've added:

1) A new `@openfn/common/util` package which exposes utility functions (designed for other adaptors for now, but we could expose this from common for job code)
2) A new `expandReferences` utility like this:

```
 const [_resource, _data] = expandReferences(state, resource, data); 
```

It is just an experiment, but actually I think this can be merged to main very quickly.

## Details

The new `expandReferences` function is designed to resolve multiple references in a single line, with a cleaner syntax. It's clean, safe, performant and elegant for every operation factory to call this immediately.

At the time of writing, expandReferences does NOT support the `skip` functionality. That probably requires a `expandReferencesWithSkip(state, skipfn, ...args)` function.

It is up to the adaptor author what names to use on the left hand side of the expansion. However I think we should aim to provide a strong convention here.  Options are:

* `_name` - the underscore looks ugly. Usually underscore means "private method" or "unused variable" in JS convention.
* `$name` - I quite like this, but it does look a bit like a JSON path object. Or angular code.
* `Name` - we could uppercase the resolved reference. This is OK, although in Javascript world an uppercase normally means a class. 

The other approach is to flip this on its head and rename the parameters:
```
/**
 * Create some resource in some system
 * @param {string} resource - The type of entity that will be created
 * @param {object} data - The data to create the new resource
 * @param {function} callback - An optional callback function
 * @returns {Operation}
 */
export function create(_resource, _data, callback) {
  return state => {
    const [resource, data] = expandReferences(state, _resource, _data); 
``` 
This is neat and should not affect JS doc or other tooling (although this needs testing - does it make VSC and Monaco intellisense ugly?)

## Issues

Closes #264

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/common/build.config.js
+++ b/packages/common/build.config.js
@@ -1,4 +1,4 @@
 // override build config with an extra entry point
 export default path => ({
-  entry: [`${path}/src/index.js`, `${path}/src/metadata.js`],
+  entry: [`${path}/src/index.js`, `${path}/src/metadata.js`, `${path}/src/util.js`],
 });
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -24,7 +24,11 @@
     "./metadata": {
       "import": "./dist/metadata.js",
       "require": "./dist/metadata.cjs"
-    }
+    },
+    "./util": {
+      "import": "./dist/util.js",
+      "require": "./dist/util.cjs"
+    } 
   },
   "author": "Open Function Group",
   "license": "LGPL-3.0-or-later",
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -1,0 +1,30 @@
+/**
+ * General-purpose utility functions
+ * 
+ * These are designed more for use in adaptor code than job code
+ * (but we could choose to export util from common)
+ * 
+ * None of these functions are operation factories
+ */
+
+// TODO this doesn't currently support skip
+export function expandReferences(state, ...args) {
+  return args.map((value) => expandReference(state, value));
+}
+
+function expandReference(state, value) {
+  if (Array.isArray(value)) {
+    return value.map(v => expandReference(state, v));
+  }
+
+  if (typeof value == 'object' && !!value) {
+    return Object.keys(value).reduce((acc, key) => {
+      return { ...acc, [key]: expandReference(state, value[key]) };
+    }, {});
+  }
+
+  if (typeof value == 'function') {
+    return expandReference(state, value(state));
+  }
+  return value;
+}
--- a/packages/common/test/util.test.js
+++ b/packages/common/test/util.test.js
@@ -1,0 +1,66 @@
+import { expect } from 'chai';
+import { expandReferences } from '../src/util';
+
+describe('util', () => {
+  it('should not expand references', () => {
+    const name = 'mulder';
+    const state = {};
+
+    const [resolvedName] = expandReferences(state, name)
+
+    expect(resolvedName).to.equal('mulder')
+  });
+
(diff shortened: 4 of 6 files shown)


josephjclark:
One big thing to flag: I've flagged `@openfn/common` as an external dependency, so common will no longer be bundled with each adaptor. This is probably correct as common is listed as a dependency anyway (so right now I think it's redundantly downloaded), but this is potentially a big build change and does need testing.

taylordowns2000:
this looks great, and i'll get it in quickly but there's definitely a conversation about the naming convention to recommend.

right now, if there was a way to ensure it didn’t mess up the intellisense too much, i’d probably vote for:
```js
export function create(_resource, _data, callback) {
  return state => {
    const [resource, data] = expandReferences(state, _resource, _data); 
 ```
I like it because it’s basically our way of saying: **DON’T USE THIS PARAMETER _WITHOUT_ EXPANDING IT!**

Which in turn begs the question… should we enforce parameter expansion in some other way? A non-exhaustive list of hidden contracts we keep at OpenFn:

1. calling each operation with state, one by one, synchronously, and passing the return from the previous into the next
2. ~expanding each parameter as if it might be a function that takes state~
3. other things

???

josephjclark:
> I like it because it’s basically our way of saying: DON’T USE THIS PARAMETER WITHOUT EXPANDING IT!

But it only says that to the person writing the adaptor, who is the same person who wrote the parameter declaration.

To users of the API (ie job writers), we either want to present the parameters verbatim, or indicate that they are expandable.

I quite like `export function create(_resource, _data, callback)` but the more I think about it, it WILL mess up intellisense.

My favourites are:

* Accept the verbose form of `resolvedResource` (which we use today anyway)
* use `$resource` as a convention to indicate a resolved parameter. 

I am actually preferring `$resource` right now. It's overplayed, but it's clear, terse and readable. It won't get confused with json path because it's internal to adaptor code. It doesn't affect jobs and doesn't appear in the operation signature.

taylordowns2000 (approved):

josephjclark:
The last example looks pretty nice to me. We can try rolling it out on the next adaptor we work on to see how it feels.

An immutability layer wouldn't really help. Immutability is more about preventing you from mutating an object or array - whereas the underlying problem here is about writing to a variable held in an outer closure.

Linting and code review are the best defenses.

PR #287, opened 2023-06-22
Extend `parseCsv` to `http` and `bigquery`

mtuchi:
## Summary

Use  `common` `parseCsv` in `http` and `bigquery`

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -554,10 +554,10 @@ export function chunk(array, chunkSize) {
 }
 
 /**
- * The function `parseCsv` takes a CSV file string or stream and parsing options as input, and returns a promise that
- * resolves to the parsed CSV data.
+ * Takes a CSV file string or stream and parsing options as input, and returns a promise that
+ * resolves to the parsed CSV data as an array of objects.
  * Options for `parsingOptions` include:
- * - `delimiter` {string/Buffer/[string/Buffer]} - Defines the character(s) used to delimitate the fields inside a record. Default: `','`
+ * - `delimiter` {string/Buffer/[string/Buffer]} - Defines the character(s) used to delineate the fields inside a record. Default: `','`
  * - `quote` {string/Buffer/[string/Buffer]} - Defines the characters used to surround a field. Default: `'"'`
  * - `escape` {Buffer/string/null/boolean} - Set the escape character as one character/byte only. Default: `"`
  * - `columns` {boolean / array / function} - Generates record in the form of object literals. Default: `true`
@@ -566,11 +566,11 @@ export function chunk(array, chunkSize) {
  * - `ltrim` {boolean} - Ignore whitespace characters from the left side of a CSV field. Default: `true`
  * - `rtrim` {boolean} - Ignore whitespace characters from the right side of a CSV field. Default: `true`
  * - `chunkSize` {number} - The size of each chunk of CSV data. Default: `Infinity`
- * - `skip_empty_lines` {boolean} - The `skip_empty_lines` skips any line which is empty. Default: `true`
+ * - `skip_empty_lines` {boolean} - Ignore empty lines in the CSV file. Default: `true`
  * @public
  * @function
  * @param {String | Stream} csvData - A CSV string or a readable stream
- * @param {Object} [parsingOptions] - Optional. Parsing options for converting CSV to JSON. \n
+ * @param {Object} [parsingOptions] - Optional. Parsing options for converting CSV to JSON.
  * @param {function} [callback] - (Optional) callback function. If used it will be called state and an array of rows.
  * @returns {Operation} The function returns a Promise that resolves to the result of parsing a CSV `stringOrStream`.
  */
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -629,3 +629,19 @@ export function parseCsv(csvData, parsingOptions = {}, callback) {
     return finalState;
   };
 }
+
+// /**
+//  * Returns a unique array of objects by an attribute in those objects
+//  * @public
+//  * @example
+//  * uniqBy([{a:1}, {b:2}, {a:1, b:2}], 'a')
+//  * @function
+//  * @param {Object} array - Array of objects to be deduplicated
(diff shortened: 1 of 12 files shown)


taylordowns2000 (approved):

taylordowns2000:
I've added a broken test in here and want to hold off on the merge until I get the right idea !

@josephjclark , @stuartc ...

Mtuchi and I couldn't _quite_ figure out how to use `parseCsv` in this canonical implementation (see failing test in language-http) where we want to send an http request for each chunk and see the responses from those http requests in state.

If the answer is that we can make the requests (which we can, according to my new test), but that we can't see the responses in state, that might also be ok.

But I want to make sure I understand the limits here before merging. Thank you!

packages/http/test/index.js
```diff
@@ -509,13 +509,54 @@ describe('post', () => {
...
+      async (state, rows) => {
+        // TODO: how do we write these in the real world? Like this?
+        if (rows.length > 0)
+          await post(
+            'https://www.example.com/api/csv-reader',
+            {
+              body: rows,
+            },
+            state => {
+              console.log("I'm a regular callback");
```
josephjclark, line 543: @taylordowns2000 Coming in out of context so I may be missing something, but isn't the http response now on state.data here?

So you can just `state.apiReferences.push(state.data)`

If we're not seeing that at runtime, that feels like a bug

packages/http/test/index.js
```diff
@@ -509,13 +509,54 @@ describe('post', () => {
...
+  it('should make an http request from inside the parseCSV callback', async function () {
+    const csv = 'id,name\n1,taylor\n2,mtuchi\n3,joe\n4,stu\n5,elias';
+    const state = { references: [], data: [], apiResponses: [] };
+
+    const resultingState = await parseCsv(
+      csv,
+      { chunkSize: 2 },
+      async (state, rows) => {
+        // TODO: how do we write these in the real world? Like this?
+        if (rows.length > 0)
```
josephjclark, line 536: If you're looking for a more functional solution, you could declare the callback like this:

```
(state, rows) => rows.length && post(...)
```
You wouldn't need to async/await the post because it'll return a promise.

mtuchi: @josephjclark i tried this approach 
```
(state, rows) =>
  rows.length > 0 &&
  post(
    'https://www.example.com/api/csv-reader',
    {
      body: rows,
    },
    state => {
      console.log("I'm a regular callback");
      return state;
    }
  )
``` 
It did not work. i was getting this error 
```
 AssertionError: expected [Function] to deeply equal { data: { references: [], …(2) } }
 ```
 
 Then tried 
 ```
 (state, rows) =>
  rows.length > 0 &&
  post(
    'https://www.example.com/api/csv-reader',
    {
      body: rows,
    },
    state => {
      console.log("I'm a regular callback");
      return state;
    }
  )(state)
  ```
  Now i am getting this error 
  ```
   TypeError: state.references is not iterable
   ```
   I think we should pair to see pair to discuss this further, You can see the [current implementation of parseCsv here ](https://github.com/OpenFn/adaptors/blob/7034947eded80f6dc69681be3ec61a7dd38dbcd8/packages/common/src/Adaptor.js#L565)

**Comment by mtuchi** [comment]
@josephjclark i have pushed the changes from our pairing today. Can you review this PR and let me know if it needs more adjustment

File: packages/common/src/Adaptor.js [hunk 3]
```diff
@@ -481,6 +481,16 @@ export function humanProper(str) {
...
+ * Splits an object into two objects based on a list of keys.
+ * The first object contains the keys that are not in the list,
+ * and the second contains the keys that are.
+ * @public
+ * @function
+ * @param {Object} obj - The object to split.
+ * @param {string[]} keys - List of keys to split on.
+ * @returns {Object[]} - Tuple of objects, first object contains keys not in list, second contains keys that are.
+ */
 export function splitKeys(obj, keys) {
```
josephjclark, line 494: Is it correct that we're publicly documenting  `splitKeys` now?

I'm not against it but it doesn't feel related to this work. It should really be mentioned in the main PR description.

packages/common/src/Adaptor.js
```diff
@@ -586,62 +598,34 @@ export function parseCsv(csvData, parsingOptions = {}, callback) {
...
+        // eslint-disable-next-line no-param-reassign
+        state = nextState;
+        buffer = nextBuffer;
+      }
+    }
+    const [finalState] = await flushBuffer(state);
+    return finalState;
   };
 }
-// /**
```
josephjclark, line 633: Hmm, this shouldn't be removed here either.

Is this diff just out of date? If we're behind then that's ok :)

packages/http/test/index.js
```diff
@@ -509,13 +509,50 @@ describe('post', () => {
       return body;
     });
 
+    testServer
+      .post('/api/csv-reader')
+      .times(3)
+      .reply(200, function (url, body) {
+        console.log('thanks for the message!', body);
```
josephjclark, line 516: Should probably remove this log

packages/http/test/index.js
```diff
@@ -509,13 +509,50 @@ describe('post', () => {
...
+      });
+
     testServer
       .post('/api/fake-custom-success-codes')
       .reply(302, function (url, body) {
         return { ...body, statusCode: 302 };
       });
   });
 
+  it('should make an http request from inside the parseCSV callback', async function () {
```
josephjclark, line 527: @taylordowns2000 just to bring you up to speed on this, here's what your http test looks like now.

Is this all as you expect?

josephjclark (changes requested):
Dependencies need updating in bigquery and http

packages/common/src/Adaptor.js
stuartc: This was me, I guess we were exporting it and _not_ documenting it - so I thought we should document the interface.

@mtuchi can you update the PR notes to include something like "Added documentation for `splitKeys`.

packages/http/test/index.js
taylordowns2000: @mtuchi is all good with this! thanks joe!

josephjclark (approved):

josephjclark (approved):
Looks good to me now!

PR #296, opened 2023-06-27
Noodle: run actual job code in tests

josephjclark:
Imagine if our unit tests looked like this:
```
  it('executes each operation in sequence', async () => {
    const job = `
      fn(() => ({ counter: 1 }))
      fn(() => ({ counter: 2 }))
      fn(() => ({ counter: 3 }))
    `

    const finalState = await execute(job, Adaptor)
    expect(finalState).to.eql({ counter: 3 });
  });
```
It would be so much easier to write tests and properly compose operations.

I've tested this in `template` just to get a vibe for how it works. So far so good.

This depends on changes in kit. I expect it to fail on local machines right now.

## How it works

I've added a helped function in `tools` which calls `@openfn/compile` and `@openfn/execute`.

Usually the compiler will inject an import statement for the adaptor, and the runtime's linker will load it. But because we want to unit test from source, not the built adaptor, this doesn't really work.

So instead, we inject the adaptor function's exports into the runtime as globals. This is good enough to run adaptor code through the actual runtime pipeline.

I've added support to the runtime to:
a) accept globals injected into the runtime context.
b) use `globals.execute`, if available, instead of the default execute function 

## Still to do

This is just an initial spike/noodle into what a solution might be.

We need to think about and merge the changes in kit. I am having some problems pushing to kit from this laptop (??) so this ma need to wait until I get back.

Aside from adding/rewriting tests, I'm not sure much else is needed here

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,10 +20,11 @@
   "license": "ISC",
   "devDependencies": {
     "@changesets/cli": "2.25.0",
-    "@openfn/buildtools": "workspace:^1.0.2",
-    "@openfn/metadata": "workspace:^1.0.1",
-    "@openfn/parse-jsdoc": "workspace:^1.0.0",
+    "@openfn/buildtools": "workspace:*",
+    "@openfn/metadata": "workspace:*",
+    "@openfn/parse-jsdoc": "workspace:*",
     "@openfn/simple-ast": "0.4.1",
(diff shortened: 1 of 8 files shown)


josephjclark:
Superceded by #482

PR #297, opened 2023-06-27
`msgraph` add sharepoint

mtuchi:
## In summary
TL;DR: We want to be able to download files from a [sharepoint](openfnorg.sharepoint.com) site using the following wishful job code 
```
getDrive({ id: 'openfnorg.sharepoint.com', owner: 'sites' });

getFolder('/BC Actuals Export');

fn(state => {
  const groupOfActuals = [];
  const itemsDownloadUrl = state.data.value
    .filter(value => value.file.mimeType === 'application/vnd.ms-excel')
    .map(value => value['@microsoft.graph.downloadUrl']);
  return { ...state, itemsDownloadUrl, groupOfActuals };
});

each(
  '$.data',
  get(
    () => state.data,
    {},
    state => {
      state.groupOfActuals.push(state.data);
      return {
        ...state,
        references: [],
        data: {},
        response: {},
      };
    }
  )
);
```
### Changes 
Add the following functions

- [x] `getDrive(specifier, name = 'default', callback = s => s)`
- [x] `getFolder(pathOrId, options, callback = s => s)`
- [x] `getFile(pathOrId, options = {}, callback = s => s)`

### Docs & Example
- [x] `getDrive(specifier, name = 'default', callback = s => s)`
- [x] `getFolder(pathOrId, options, callback = s => s)`
- [x] `getFile(pathOrId, options = {}, callback = s => s)`

### Tests
- `getDrive(specifier, name = 'default', callback = s => s)`
  - [x] `should get a drive by id and set it to state`
  - [x] `should get a named drive by id and set it to state`
  - [x] `should get the default drive for a site`
 
- `getFolder(pathOrId, options, callback = s => s)`
  - [x] `should get a folder by id`
  - [x] `should get a folder for a named drive by id`
  - [x] `get folder at path`
  - [x] `throw a helpful error if drive is not defined`
  
- `getFile(pathOrId, options = {}, callback = s => s)`
  - [x] `should get a file by id`
  - [x] `should get a file for a named drive by id`
  - [x] `get a file metadata ({ metadata: true }) shoudl return itemWithDownloadUrl`

### Out of scope (TODO's) #310 
_Open an issue before assign reviewer for TODO's_
- **getDrive()**
```
// TODO: pass a string drive id as the specifier
// TODO: how will we get a list of drives?
// TODO: can we use /me/drive to get the user's default drive?
// TODO maybe error if no id on -> const { id, owner = 'drive' } = resolvedSpecifier;
```
- **getFile**
```
// TODO implement metadata (don't get file content, just metadata)
// TODO implement filter (maps to $filter, must be URI encoded)
// TODO if response is a single file, it needs to be nested in an array
```
## Review Checklist

- [x] I have review my code (remove unused code, console.log, todo's)
- [x] I have open an issue for pending TODO's #310 


packages/msgraph/src/Sharepoint.js
```diff
@@ -1,46 +1,120 @@
...
+ * @param {string} sharepointSite -A SharePoint siteId
+ * @param {function} [callback = state => state] - An optional callback function
+ * @return {Operation}
+ */
+export function getSites(sharepointSite, callback = s => s) {
+  return state => {
+    const { accessToken, apiVersion } = state.configuration;
+    const [resolvedRootSite] = expandReferences(state, sharepointSite);
+
+    const resolvePath = resolvedRootSite
```
josephjclark, line 22: This can be simplified to:

```
const resolvePath = `sites/${resolvedRootSite || 'root}`
```

packages/msgraph/test/mockAgent.js
```diff
@@ -2,42 +2,547 @@ import { MockAgent } from 'undici';
 
 const mockAgent = new MockAgent();
 
-const mockPool = mockAgent.get('https://fake.server.com');
+const mockPool = mockAgent.get('https://graph.microsoft.com');
+
+const driveResponse = {
```
josephjclark, line 7: I would move each fixture data out into its own file, probably in `test/fixtures`. That will really clean up the mock agent

packages/msgraph/test/Adaptor.test.js
```diff
@@ -38,67 +45,286 @@ describe('execute', () => {
   });
 });
 
-describe('create', () => {
-  it.skip('makes a post request to the right endpoint', async () => {
+describe('getSites', () => {
```
josephjclark, line 48: These unit tests are excellent, well done. They help me understand the APIs and their expected usage as well!

josephjclark (changes requested):
Sorry - lots of comments here, some of them on the wider adaptor design.

I don't know the details of msgraph or its sharepoint API (your excellent unit tests help!), but I'm not convinced the adaptor signatures are correct.

When a function takes an object as an argument, I expect one of three things:
a) All the properties of the object must be provided
b) The object represents a list of options and they're all optional
c) The object represents some data and will just be fed/forwarded through the function

The signatures here all take an object but they do not confirm to any of these patterns. Some keys are required, some keys are optional, some are mutually exclusive. It just makes it hard to predict what the API will do if various options are passed in, and makes it hard for me as as user to know what values are important.

I've made suggestions in-line for better signatures for these functions, but they may need a bit more thought yet (and I may be misunderstanding the API)

I've also got some comments about the language around `resolve` variables, and some wider comments on the adaptor. I recommend doing all the work in this PR (and updating the title and description to match)

josephjclark:
Ok after some reflections (and a little help from chatGPT) I'm gonna recommend this as the signature:
```
@param drive {string} - the id of the drive to get
@param drive {object} - an object describing the drive location (site/group/user). Takes a resource and resourceId
function getDrive(drive?, callback?): (state) => state
```
// Get default drive for the current user
// me/drive
getDrive(undefined, callback)

// Get any drive by id
// /drives/{driveId}
getDrive('<drive-id>', callback)

// Get default drive for a site
// /sites/siteId/drive
getDrive({ resourceId: "<sharepoint-site-id>", resource: 'sites' }, callback) 

// get default drive for user
// users/userId/drive
getDrive({ resourceId: "<user-id>", resource: 'users' }, callback) 

// get default drive for a group
// groups/groupId/drive
getDrive({ resourceId: "<group-id>", resource: 'groups' }, callback) 
```

Some notes:
* I am not sure about the semantics around the "resource"
* It's a little bit lumpy but it's the best compromise of devx, documentation and implementation that I've found.
* I prefer the overridden signature with arity 0,1,2 or 3 - that gives us the most control and the best devx, but it's the hardest to document.
* I am using `resourceId` and not `id` because `getDrive({ id })` is a bit misleading and confusing (the id is NOT the drive id)
* I have not added support for an options object. We may want this for `expand` but I suggest we cross that bridge when we come to it.
* I suppose we could model the "resource" as a tuple `['sites', '<site-id>']`, but that's not very Javascripty.

We will later introduce a `listDrives` API which has a similar signature but returns a list of drives (or drive ids?)

File: packages/msgraph/src/Adaptor.js [hunk 4]
```diff
@@ -88,12 +89,57 @@ export function get(path, query, callback = false) {
...
+ * getDrives({siteId: "openfn.sharepoint.com"})
+ * @example <caption>Get a drive for a site</caption>
+ * getDrives({siteId: "openfn.sharepoint.com", defaultDrive: true})
+ * @example <caption>Get a drive by ID</caption>
+ * getDrives({driveId: "YXzpkoLwR06bxC8tNdg71m"})
+ * @param {object} [resource={ driveId: '', siteId: '', groupId: '', defaultDrive: false }] - A resource object containing resource ids and options
+ * @param {function} [callback = s => s] (Optional) Callback function
+ * @return {Operation}
+ */
+export function getDrives(
```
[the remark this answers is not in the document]
mtuchi: Changed the function signature to 
```
// Get default drive for the current user
// me/drive
getDrive(undefined, callback)

// Get any drive by id
// /drives/{driveId}
getDrive('<drive-id>', callback)

// Get default drive for a site
// /sites/siteId/drive
getDrive({ resourceId: "<sharepoint-site-id>", resource: 'sites' }, callback) 

// get default drive for user
// users/userId/drive
getDrive({ resourceId: "<user-id>", resource: 'users' }, callback) 

// get default drive for a group
// groups/groupId/drive
getDrive({ resourceId: "<group-id>", resource: 'groups' }, callback) 
```
see [f3fcc79](https://github.com/OpenFn/adaptors/pull/297/commits/f3fcc79fb2325ceb9b56650a6f6fe4e74cabaf92)

mtuchi:
@josephjclark i thin the first argument for `getDrive()` should be `driveOrResource` and not `drive`

`getDrive(driveOrResource, callback)`
because it doesn't make sense to call `{resourceId : '', resource:''}` these options have nothing to do with drive but have something to do with the drive resource

packages/msgraph/src/Adaptor.js
```diff
@@ -88,12 +89,57 @@ export function get(path, query, callback = false) {
...
+) {
+  return state => {
+    const { accessToken, apiVersion } = state.configuration;
+    const [resolveResource] = expandReferences(state, resource);
+
+    const { siteId, driveId, groupId, defaultDrive } = resolveResource;
 
+    let resolvePath = 'me/drives';
+
+    if (driveId) resolvePath = `drives/${driveId}`;
```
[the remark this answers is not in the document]
mtuchi: Fixed

packages/msgraph/test/mockAgent.js
mtuchi: Fixed here [ee84147](https://github.com/OpenFn/adaptors/pull/297/commits/ee84147293f921c322f5fd27f17746e64cb4a458)

josephjclark:
@mtuchi If you want to get a drive you just do `getDrive("<drive-id>")`. If you want to get the default drive for a resource you do `getDrive({ resourceId: '<site-id>', resource: 'site' })`.

The parameter is called `drive` because you're specifying a drive by either string id or object resource. We could call it `driveIdOrResource` yeah, but it's pretty cumbersome.

I am not keen on the word `resource` as it's a bit generic. They use `library` sometimes in the msgraph docs but I don't know if that's more helpful.

packages/msgraph/src/Sharepoint.js
```diff
@@ -1,46 +1,120 @@
...
+      ? `sites/${resolvedRootSite}`
+      : 'sites/root';
+    const url = setUrl({ apiVersion, resolvePath });
```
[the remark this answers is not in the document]
mtuchi: @josephjclark i have changed the signature to `getUrl(urlPath, apiVersion)` Does this make sense to you now ?

josephjclark:
After a long call with @mtuchi I'm going to recommend more changes API (sorry!)

It's proving really tricky to map the msgraph API to a clean JS interface. We either have a million functions or a very complex, undocumentable list of arguments.

After going through some of the cases I've decided to try and raise the abstraction level. All we really want to do is get (and maybe write) a list of files from drive. The rest is all detail and boilerplate. And I think it's important that job writers can get information out of msgraph without having a PHD in their documentation.

So the new concept is basically this:

* At the top of an `msgraph` job, you 'connect' to a drive (get a reference to it). There are multiple ways you might do this. The drive reference is written to state (with an optional name so that you can connect multiple drives)
* You can use `getFile`, `writeFile` and `getFolder` with a path (and optional drive name). This will use the drive id from your state and pull file/folder contents/metadata down for you.

We're breaking for evening but right now the API looks like this:
```
// Gets a drive reference and writes to state.drives.default
// @param specifier: specify the drive to get. default 'me'.
//         If an id/owner pair, will return the default drive for the specified owner
// @param name: optional name for the drive. Will be written to state.drives[name].
getDrive(specifier, name, callback)

// Examples:
getDrive({}, 'joe');   // get my drive and name it joe
getDrive('me', 'joe');   // get my drive and name it joe
getDrive({ id: 'driveId' }, 'joe'); // Get a drive by Id
getDrive({ id: '<siteId>', owner: 'sites'}, 'joe'); // get default drive for a site
getDrive({ id: '<userId>', owner: 'users' }); // get default drive for a user

// return an array of folder contents
// We should safely URI encode the path
// Path must start with /, else will be treated as an id
// writes to state.data
// uses default drive
getFolder(pathOrId, { options }, (callback = s => s));

// return a single file's content (pass { metadata: true } to get metadata only)
// return an array of folder contents
// We should safely URI encode the path
// Path must start with /, else will be treated as an id
// writes to state.data
// uses default drive
getFile(pathOrId, { options }, callback);

// writes one or more files to the path provided
writeFiles(pathOrId, { options }, data, callback)

// loose spec for getFile/writeFiles options
const options = {
  driveName: 'string', // drive of the file to use (optional)
  metadata: false, // return metadata, not content
  $filter: 'string' // filter on the query. Needs to be URI encoded
};
```

josephjclark:
You can get files on a drive by path [[docs](https://learn.microsoft.com/en-us/graph/onedrive-addressing-driveitems)]. Paths and queries must be URI encoded, single quotes must be double escaped. The path looks a bit like this: `".../root:/<content%20path>:/content"

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +85,195 @@ export function create(resource, data, callback) {
...
+ * @param {object} options - (Optional) Query parametes
+ * @param {function} [callback = s => s] (Optional) Callback function
+ * @return {Operation}
+ */
+export function getFolder(pathOrId, options, callback = s => s) {
+  return async state => {
+    const defaultOptions = {
+      driveName: 'default', // Named drive in state.drives
+      metadata: false, // If false return folder files if true return folder metadata
+      $filter: '', // Eg: "file/mimeType eq \'application/vnd.ms-excel\'"
```
josephjclark, line 162: Did we get filter working? I can't see any logic for it. If not we should probably comment this out and add a TODO.

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +85,195 @@ export function create(resource, data, callback) {
...
+
+    const { id: driveId } = state.drives[driveName];
+
+    let urlPath;
+
+    if (resolvedPathOrId.startsWith('/')) {
+      urlPath = `drives/${driveId}/root:/${encodeURIComponent(
+        resolvedPathOrId
+      )}`;
+      metadata ? urlPath : (urlPath = `${urlPath}:/children`);
```
josephjclark, line 197: This ternary shoud be written as
```
if (metadata) {
   urlPath += "/:children";
}
```
Also it should be done after the if/else block, because it's the same code either way.

Check this pattern across this file.

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +85,195 @@ export function create(resource, data, callback) {
...
+        `Quick Fix: Add getDrive() operation before getFolder()`,
+        `${egJobCode}`,
+      ].join('\n\t∟ ');
+
+      throw new Error(errorString);
+    }
+
+    const { id: driveId } = state.drives[driveName];
+
+    let urlPath;
```
josephjclark, line 191: Can we call this `resource` or something? It's weird how we have a urlPath and then create a URL from it. Better to model this as the resource we are fetching I think.

This change needs applying to a few functions.

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +85,195 @@ export function create(resource, data, callback) {
...
+      pathOrId,
+      options
+    );
+
+    const { driveName, metadata, select } = {
+      ...defaultOptions,
+      ...resolvedOptions,
+    };
+
+    if (!state.drives[driveName]) {
```
josephjclark, line 244: This whole code block is re-used - it should be factored out into a function (probably in Utils.js?)

josephjclark (changes requested):
This is looking great - dumb question but I assume it works for your job? Are YOU happy with it?

I've left a couple of comments which need addressing. I would like to look over the unit tests but I'm out of time right now.

packages/msgraph/src/Adaptor.js
mtuchi: We have an issue that track all todo's https://github.com/OpenFn/adaptors/issues/310, So mean while i have commented out `$filter`

packages/msgraph/src/Adaptor.js
mtuchi: 👍🏽 , i have added `if (!metadata) {
      urlPath += resolvedPathOrId.startsWith('/') ? ':/children' : '/children';
    }` after `if/else` block

packages/msgraph/src/Adaptor.js
mtuchi: Done https://github.com/OpenFn/adaptors/pull/297/commits/76b7d8907a1cfa09b389b6f0e5db8f57c466ad68

packages/msgraph/src/Utils.js
```diff
@@ -1,13 +1,31 @@
+import { fetch } from 'undici';
 import { composeNextState } from '@openfn/language-common';
 
-export function setUrl(urlParams) {
-  const { apiVersion, resolvePath } = urlParams;
+export function ensureDriveExistsOrFail(drive, fileType) {
```
josephjclark, line 4: You can just call this `assertDrive`. The word `assert` means "test something or throw", so the functionality would be well represented.

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +91,205 @@ export function create(resource, data, callback) {
...
+
+    return request(url, { ...auth }).then(response => {
+      state.drives[resolvedName] = response;
+      return callback(state);
+    });
+  };
+}
+
+/**
+ * Get folder metadata or an array of folder contents.
```
josephjclark, line 152: This should say "get the contents or metadata of a folder", because contents have precedence (they are they default)

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +91,205 @@ export function create(resource, data, callback) {
...
+ * @public
+ * @example <caption>Get a folder by ID</caption>
+ * getFolder('01LUM6XOCKDTZKQC7AVZF2VMHE2I3O6OY3')
+ * @example <caption>Get a folder for a named drive by id</caption>
+ * getFolder("01LUM6XOCKDTZKQC7AVZF2VMHE2I3O6OY3",{ driveName: "mydrive"})
+ * @param {string} pathOrId - A path to a folder or folder id
+ * @param {object} options - (Optional) Query parametes
```
josephjclark, line 159: typo on `parameters`

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +91,205 @@ export function create(resource, data, callback) {
...
+      state,
+      pathOrId,
+      options
+    );
+
+    const { driveName, metadata } = { ...defaultOptions, ...resolvedOptions };
+
+    ensureDriveExistsOrFail(state.drives[driveName], 'folder');
+
+    // if (!state.drives[driveName]) {
```
josephjclark, line 181: We can delete these comments now

packages/msgraph/src/Utils.js
josephjclark: Also this should just take `state` and a drive name as parameters. Make it as easy as you can to call it

packages/msgraph/src/Utils.js
```diff
@@ -1,13 +1,31 @@
...
+  if (!drive) {
+    const operation = fileType === 'folder' ? 'getFolder' : 'getFile';
+    const egJobCode = [
+      `Eg: Get a site drive then get "/Sample Data" ${fileType}`,
```
josephjclark, line 8: I really like this level of help but I do think it's a bit over-complicated. That the `fileType` parameter is kinda weird and won't scale. I'd like to remove it to simplify the signature.

I think I'd make the whole error message just say this:
```
Drive is not defined.

At the top of your job you should define all the drives you want to us.

eg: getDrive({ id: "openfn.sharepoint.com", owner: "sites"})
```
if we change the parameters  to `state, driveName`, you should take the drive name into account (which is probably a different example)/

packages/msgraph/src/Adaptor.js
```diff
@@ -76,14 +91,205 @@ export function create(resource, data, callback) {
...
+      options
+    );
+
+    const { driveName, metadata } = {
+      ...defaultOptions,
+      ...resolvedOptions,
+    };
+
+    ensureDriveExistsOrFail(state.drives[driveName], 'file');
+    // if (!state.drives[driveName]) {
```
josephjclark, line 256: These comments can also go

packages/msgraph/test/Adaptor.test.js
```diff
@@ -37,68 +39,439 @@ describe('execute', () => {
...
+          id: 'b!YXzpkoLwR06bxC8tNdg71m_',
+        },
+      },
+    };
+
+    const finalState = await execute(
+      getFolder(
+        '01LUM6XOCKDTZKQC7AVZF2VMHE2I3O6OY3',
+        { metadata: true },
+        state => {
```
josephjclark, line 170: I'm being picky but we don't actually need to use the callback here - let's remove it as it only confuses the test!

packages/msgraph/test/Adaptor.test.js
```diff
@@ -37,68 +39,439 @@ describe('execute', () => {
...
+      },
+      drives: {
+        default: {
+          id: 'b!YXzpkoLwR06bxC8tNdg71m_',
+        },
+      },
+    };
+
+    const finalState = await execute(
+      getFolder('01LUM6XOCKDTZKQC7AVZF2VMHE2I3O6OY3', state => {
```
josephjclark, line 193: Again, no callback needed. Check the other functions, because it's copied and pasted I think we'll see this a lot

josephjclark (changes requested):
This is looking great. I've said it before but I'll say it again, I love the unit tests.

I'm really sorry but I have just enough nitpicks - and they are nitpicks - that I'm going to ask for one more small round of changes. Shouldn't take long, the help message is the hardest bit (we can pair over that if you like)

packages/msgraph/src/Utils.js
mtuchi: @josephjclark the examples are different, that's why the function takes `drive` and `fileType` because i want to be able to show different errors when `assertDrive()` in `getFolder()` and in `getFile()`

packages/msgraph/src/Utils.js
mtuchi: Done here [839975e](https://github.com/OpenFn/adaptors/pull/297/commits/839975e9adb3311d10912eb66c1a419ae55960db)

packages/msgraph/src/Adaptor.js
mtuchi: Done [839975e](https://github.com/OpenFn/adaptors/pull/297/commits/839975e9adb3311d10912eb66c1a419ae55960db)

packages/msgraph/src/Adaptor.js
mtuchi: Done [c6dee2f](https://github.com/OpenFn/adaptors/pull/297/commits/c6dee2f238fd73974bea9557342c399c735e70fd)

packages/msgraph/test/Adaptor.test.js
mtuchi: Done [c7d7243](https://github.com/OpenFn/adaptors/pull/297/commits/c7d72434d44e189403c6b983dfef73da42c21d6d)

packages/msgraph/test/Adaptor.test.js
mtuchi: Done [c7d7243](https://github.com/OpenFn/adaptors/pull/297/commits/c7d72434d44e189403c6b983dfef73da42c21d6d)

josephjclark (approved):

PR #298, opened 2023-06-28
Normalize access token

taylordowns2000:
@josephjclark , this is a bit of a compromise, but as I started to change `accessToken` to `access_token` all over so many Javascript files, I realized that what we actually want is to accept standard JSON credentials (which we now do, see configuration-schemas) and then convert them in a standard way to some JavaScript-sensible configuration.

This is a much lighter change, will handle both snakes and camels, and provides us with an interface for further OAuth2 credential-to-adaptor-auth magic if we need it.

There is a bunch of formatting that got applied (prettier must not have been run on these files before) but the change that's indicative of all this stuff is here: https://github.com/OpenFn/adaptors/pull/298/files#diff-a5033e85ba9e830835a40e0ced8f110c61ba2c8640e42c8c88287932cf542f68R30

Thoughts?

### DIFF
--- a/.changeset/large-monkeys-protect.md
+++ b/.changeset/large-monkeys-protect.md
@@ -1,0 +1,11 @@
+---
+'@openfn/language-common': minor
+'@openfn/language-dynamics': patch
+'@openfn/language-facebook': patch
+'@openfn/language-googlehealthcare': patch
+'@openfn/language-googlesheets': patch
+'@openfn/language-msgraph': patch
+---
+
+Ensure that standard OAuth2 credentials with snake-cased "access_token" keys can
+be used for OAuth2-reliant adaptors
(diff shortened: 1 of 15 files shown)


taylordowns2000:
oooooh and also @josephjclark , how do I introduce a new function to `common` and use it in another adaptor?

this looks like a dependency failure (the tests run OK locally) and I wonder if I need to build and release common first, then update the common version in my other adaptors? (But I thought this was handled already by that `workspace` bit in pnpm.)

image

taylordowns2000:
@josephjclark , i spoke about this with @stuartc and he's alright with the direction. if you're able to weigh in today, great, if not I will merge this (it's not a breaking change anywhere, only adds that deprecation helper/converter to the oauth2 dependent adaptors) so as to unblock the oauth2 token refresh stuff on lightning.

josephjclark:
Hi, sorry for the late response on this, but it looks good to me!

PR #299, opened 2023-07-04
Update `googlehealthcare` configuration-schema

mtuchi:
## Summary

Add a title and description in properties of `googlehealthcare` configuration schema

This should fix the empty input box when adding Google Healthcare credentials on Lighting 
image

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/googlehealthcare/configuration-schema.json
+++ b/packages/googlehealthcare/configuration-schema.json
@@ -22,45 +22,9 @@
         "v1",
         "v1beta1"
       ]
-    },
-    "cloudRegion": {
-      "type": "string",
-      "title": "cloudRegion",
-      "description": "The region Google Cloud region of your dataset",
-      "examples": [
-        "us-central1"
-      ]
-    },
-    "projectId": {
-      "type": "string",
-      "title": "projectId",
-      "description": "The ID of your Google Cloud project",
-      "examples": [
-        "adjective-noun-123"
-      ]
-    },
-    "datasetId": {
-      "type": "string",
-      "title": "datasetId",
-      "description": "The FHIR store's parent dataset",
-      "examples": [
-        "my-dataset"
-      ]
-    },
-    "fhirStoreId": {
-      "type": "string",
-      "title": "fhirStoreId",
-      "description": "An identifier for the FHIR store",
-      "examples": [
-        "my-fhir-store"
-      ]
     }
   },
   "required": [
-    "cloudRegion",
-    "projectId",
-    "datasetId",
-    "fhirStoreId",
     "access_token"
   ]
 }
--- a/packages/googlehealthcare/configuration-schema.json
+++ b/packages/googlehealthcare/configuration-schema.json
@@ -26,7 +26,7 @@
     "cloudRegion": {
       "type": "string",
       "title": "cloudRegion",
-      "description": " The dataset location",
+      "description": "The region Google Cloud region of your dataset",
       "examples": [
         "us-central1"
       ]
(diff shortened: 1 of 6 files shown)


packages/googlehealthcare/configuration-schema.json
```diff
@@ -18,23 +18,42 @@
...
+      "examples": [
+        "v1",
+        "v1beta1"
+      ]
     },
     "cloudRegion": {
       "type": "string",
-      "examples": ["us-central1"]
+      "title": "cloudRegion",
+      "description": " The dataset location",
```
josephjclark, line 29: Is this description correct? Is it not "The region of your Google Cloud project" ?

packages/googlehealthcare/configuration-schema.json
```diff
@@ -18,23 +18,42 @@
...
+      "description": "The ID of your Google Cloud project",
+      "examples": [
+        "adjective-noun-123"
+      ]
     },
     "datasetId": {
       "type": "string",
-      "examples": ["my-dataset"]
+      "title": "datasetId",
+      "description": "The FHIR store's parent dataset",
```
josephjclark, line 45: I honestly don't know, is the word "parent" here correct?

josephjclark (commented):
Looks fine to me although I don't really know the gory details - I left a couple of open questions in comments :)

packages/googlehealthcare/configuration-schema.json
mtuchi: @josephjclark I guess we can say that, I was following this description from Google cloud docs
Screenshot 2023-07-07 at 1 52 27 PM

packages/googlehealthcare/configuration-schema.json
mtuchi: That is a good question, according to the docs it says `the FHIR store's parent dataset` - https://cloud.google.com/healthcare-api/docs/how-tos/fhir-resources#healthcare-create-patient-drest

packages/googlehealthcare/configuration-schema.json
amberrignell: @josephjclark mtuchi and I confirmed that with google healthcare API, you can set a region at the dataset level, so this description does make sense.

amberrignell (approved):

taylordowns2000 (changes requested):
Please move projectId, dataSetId, cloudRegion, and fhirStoreId out of the credential.

taylordowns2000 (changes requested):
I don't love the proposed pattern. The `cloudRegion` (and stuff like that) isn't part of the "resource" if i'm understanding correctly, so having `name` and `gender` in the same object as `projectId` doesn't quite sit right.

I'd suggest that we adopt a signature like `createFhirResource(resource, fhirStore, callback)` where `fhirStore` is whichever one of those attributes is the _most_ granular. (That might be `fhirStoreId`, that might be `dataSetId`.)

mtuchi:
> I don't love the proposed pattern. The `cloudRegion` (and stuff like that) isn't part of the "resource" if i'm understanding correctly, so having `name` and `gender` in the same object as `projectId` doesn't quite sit right.
> 
> I'd suggest that we adopt a signature like `createFhirResource(resource, fhirStore, callback)` where `fhirStore` is whichever one of those attributes is the _most_ granular. (That might be `fhirStoreId`, that might be `dataSetId`.)

Hiya @taylordowns2000 , i have worked on the requested changes with slight adjustment to the signature of the function 
`createFhirResource(fhirStore, resource, callback)`

josephjclark (approved):
Looks good to me.

Issue #283, opened 2023-06-20
new version of `http` that uses `fetch`


taylordowns2000:
build a new fetch-based version:

1. round up all outstanding issues related to http, take them into consideration
2. make sure the new version adheres to the template style
3. take into consideration the `expandReferences` dilemma
4. confirm that we should drop the `common.http` since all users will now have access to `fetch` directly

josephjclark:
There are a few things going on at the same time here

1. Replace `axios` with `unidici` (fetch won't work). This is a non-trivial and breaking change.
2. Review the common http API signatures, and look at how they can be re-used across adaptors.
3. Review the new pattern of adaptors providing their own fetch-based http adaptor in a `utils` file.

Each is a big and breaking update, but it might make sense to consider all of them together.

josephjclark:
This is the expandReferences dilemma:

## expandReferences

expand references find what I call a "deferred value" - a value nested inside a function which must be resolved against the latest state.

`expandReferences(arg)` will recursively find any functions on `arg` and invoke them with state. It will do this for all members of an array, all keys of an object.

So this pattern, with two nested deferred values, should be totally fine:

```get(url, { query: (s) => s.q, headers: () => s.h })```

This pattern defers the whole query object, which is also fine (and for my money, better)

```get(url, (s) => ({ query: s.q, headers: s.h }))```

I do not know why this recursion is needed. Nesting deferred values doesn't  do anything but complicate the code - you don't get a different value for state or anything. I actually think it's an antipattern.

Basically every adaptor supports/depends on this recursive expansion. If we change it, we break every single adaptor in a really subtle way.

## expandRequestReferences

The common.http functions expose all the parameters of an axios request to the `params` argument

Many of axios's options look like deferred values to our recursive expander, so we'll try and invoke them. And almost certainly break something.

So `expandRequestReferences` was introduced to enable safe recursion over axios' options by only trying to expand a small number of keys.

It looks inelegant from the outside, but it's a good and necessary solution.

josephjclark:
## Replacing axios

Remove the dependency on `axios` in `language-common` and `language-http` and use `unidici` instead.

Axios and unidici do not share an API. And unfortunately common exposes axios's options directly to consumers.

Also remember that common exposes the axios object directly. Jobs which depend on this will be incompatible with this version, and authors writing new jobs against the new adaptor may need to retrain themselves.

[...]

PR #302, opened 2023-07-10
new version of http that uses fetch

mtuchi:
## Summary

Migrate from using `axios` to `fetch` in `common.http`

1. we have request()  in `common/util.js`  that uses `fetch()` from `undici`
2. http common to have all http function (`get()`, `post()`, `head()`, `patch()`, etc) and all these function `expandRefences` in their arguments.
3. All other adaptors that need access http functions they should use http common function and not implement their own version of `get()` or `post()` etc
4. If other adaptors can not  use http common function then extend it from common (For example in other adaptors we have create() which is basically a http common `post()` so it make sense to extend post from http common)

## Details

- [ ] buildRequest
- [ ] buildUrl
- [ ] request 
- [ ] tests

## Issues 
Ref #283 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -46,7 +46,8 @@
     "csvtojson": "^2.0.10",
     "date-fns": "^2.25.0",
     "jsonpath-plus": "^4.0.0",
-    "lodash": "^4.17.19"
+    "lodash": "^4.17.19",
+    "undici": "^5.22.1"
   },
   "devDependencies": {
     "@openfn/buildtools": "workspace:^1.0.2",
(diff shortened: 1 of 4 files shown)


josephjclark:
Hi @mtuchi - can we put the breaks on this until I've had a chance to really dig into it?

I keep making snap decisions on things without thinking them through - I think maybe we all do - and basically before taking this any further I want to make sure we have a really tight spec about what we want to do, and why.

As part of the issue I'm going to review the new request pattern laid down in template. I've let it sail over me a little bit and I just want to make sure it's right.

PR #309, opened 2023-07-18
update `semver`

mtuchi:
## Summary

Update `semver` to fix https://github.com/OpenFn/adaptors/security/dependabot/19

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/itchy-suits-travel.md
+++ b/.changeset/itchy-suits-travel.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-commcare': patch
+---
+
+update superagent
--- a/packages/commcare/package.json
+++ b/packages/commcare/package.json
@@ -31,7 +31,7 @@
     "form-data": "^4.0.0",
     "js2xmlparser": "^1.0.0",
     "lodash-fp": "^0.10.4",
-    "superagent": "^6.1.0",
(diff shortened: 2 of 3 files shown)


josephjclark:
@mtuchi how did you do this? Was it `pnpm update` or something?

mtuchi:
> Digging into this a bit, it looks like this PR doesn't affect superagent, which has dependencies on the insecure semver 7.5.0. That's a production dependency.
> 
> I don't really care that devDependencies like mocha or simple-ast use the older version, but `superagent` is a dependency of `commcare` which means its using that code at runtime. A quick look at `pnpm -r why semver` suggests a couple of adaptors have dependencies (`twilio`, `mssql`).
> 
> Probably not a big deal but if we're going to do this, we ought to fix those dependencies.
> 
> pnpm allows us to override dependency versions. Perhaps we could try this? Just for the affected packages, like `superagent`, to bump them up to the next safe patch version (7.5.2).
> 
> Incidentally, `magpi` declares a `dependency` on `superagent` but it's lying - it's only a `devDependency`. It uses `request` in production.

@josephjclark i have updated the `superagent` version to the latest version

josephjclark (approved):

PR #312, opened 2023-07-25
`fhir` improvements

mtuchi:
## Summary
Add `getClaim` and improve `get`. Also for uniti tests to work i had to migrate from `axios` to using `fetch` from `undici` because the unit test only works with `fetch`.

## Issues 

https://github.com/OpenFn/madagascar-dhis2-mtomady/issues/3

## Review Checklist

      migration guide followed?


josephjclark (approved):
This is really bad timing :(

We are just about to write the http stuff with a different pattern to this. I kinda wish we hadn't done this right now.

I'm going to approve it because it looks like an improvement and I assume you need the changes. The code looks fine. I haven't tested anything.

But we will have to revisit this in the next few weeks.

PR #318, opened 2023-07-27
`msgraph` cleanup state before throw

mtuchi:
## Summary

When catching an error on `execute()` i added a cleanup step then throw an error. Before that we were doing cleanup without throwing an error which made job code to pass even if there is a `throw new Error()` in that job code 

Ref #319 
## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/msgraph/test/Adaptor.test.js
+++ b/packages/msgraph/test/Adaptor.test.js
@@ -57,9 +57,12 @@ describe('execute', () => {
       },
     ];
 
-    const finalState = await execute(...operations)(state).catch(e => {
-      expect(e.message).to.contain('Failed operation');
+    let e;
+    const finalState = await execute(...operations)(state).catch(err => {
+      e = err;
     });
+
+    expect(e.message).to.contain('Failed operation');
 
     expect(finalState).to.eql(undefined);
   });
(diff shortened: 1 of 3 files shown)


packages/msgraph/test/Adaptor.test.js
```diff
@@ -38,6 +38,31 @@ describe('execute', () => {
...
+        state.counter++;
+        throw new Error('Failed operation');
+      },
+      state => {
+        return { ...state, counter: 1 };
+      },
+    ];
+
+    const finalState = await execute(...operations)(state).catch(e => {
+      expect(e.message).to.contain('Failed operation');
```
josephjclark, line 61: @mtuchi the fix looks fine but if I comment it out, your test still passes :grimacing: 

If execute doesn't throw, this assertion doesn't run, and so the test passes.

You need to do something like this (this is untested):
```
    let e;
    const finalState = await execute(...operations)(state).catch(err => {
      e = err;
    });

    expect(e.message).to.contain('Failed operation');
```

You probably need to assert that e exists (ie, the catch handler was called) and that it has the right value

mtuchi: Done [cb6adb0](https://github.com/OpenFn/adaptors/pull/318/commits/cb6adb0a0ac66c48f3e66666217edce4bcd61cc0)

josephjclark (approved):

PR #333, opened 2023-08-03
Fix opinionator

### LINKED ISSUE #330: Pin version of opinionator
We use a dependency, `word-wrap`, which allows insecure regular expressions.

This is unlikely to be a concern for us. 

it affects `mailgun` and `salesforce`. It is likely to go away with mailgun when we update to the latest version of [mailgun.js](https://github.com/mailgun/mailgun.js#create).

Salesforce is a bit trickier because the problem comes out jsonpath, which is pretty unresponsive. But we can use pnpm to pin the version of `opinionator` to `0.9.3` - see https://github.com/estools/escodegen/issues/459. That should sort it now and forever.

This closes https://github.com/OpenFn/adaptors/security/dependabot/20

josephjclark:
We've had a security alert from a low level dependency (`word-wrap`), which comes out of `optionator`.

The vulnerability affects two packages: salesforce and mailgun.

Salesforce is actually a red herring because it only affects a dev dependency, which I have corrected here (and has since been fixed on main). I've also removed `yargs` from salesforce deps as its unneeded.

For mailgun I have overridden the version of `optionator`. If you install mailgun now and run `npm why optionator`, you can see that the version is correctly overridden:

image

Note that adding the override in the monorepo root (via pnpm) doesn't actually affect the built packages (disappointing but on reflection perhaps not surprising). So we have to do it in the adaptor itself.

This is a good workaround until we properly update the mailgun package.

Note that this is based on the pnpm bump branch.

Closes #330 (and the associated vulnerabilities)

### DIFF
--- a/packages/salesforce/package.json
+++ b/packages/salesforce/package.json
@@ -32,11 +32,11 @@
   ],
   "dependencies": {
     "@openfn/language-common": "1.10.2",
+    "JSONPath": "^0.10.0",
     "axios": "^0.21.1",
     "jsforce": "^1.11.0",
     "lodash": "^4.17.21",
-    "mustache": "^2.2.0",
-    "yargs": "^3.30.0"
+    "mustache": "^2.2.0"
   },
   "devDependencies": {
     "assertion-error": "1.1.0",
--- a/packages/salesforce/package.json
+++ b/packages/salesforce/package.json
@@ -31,11 +31,10 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "1.10.2",
-    "axios": "^0.21.1",
-    "jsforce": "^1.11.0",
-    "lodash": "^4.17.21",
-    "mustache": "^2.2.0"
+    "@openfn/language-common": "workspace:*",
+    "axios": "^0.21.4",
+    "jsforce": "^1.11.1",
+    "lodash": "^4.17.21"
   },
   "devDependencies": {
     "assertion-error": "1.1.0",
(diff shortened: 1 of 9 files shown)


josephjclark:
@mtuchi I've just rebased this on top of the other changes

packages/salesforce/package.json
```diff
@@ -31,11 +31,11 @@
...
   "dependencies": {
-    "@openfn/language-common": "workspace:*",
-    "axios": "^0.21.4",
-    "jsforce": "^1.11.1",
+    "@openfn/language-common": "1.10.2",
+    "axios": "^0.21.1",
+    "jsforce": "^1.11.0",
     "lodash": "^4.17.21",
-    "yargs": "^3.32.0"
+    "mustache": "^2.2.0"
```
mtuchi, line 38: @josephjclark i don't think this package is being used in salesforce adaptor

packages/salesforce/package.json
```diff
@@ -31,11 +31,11 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "workspace:*",
-    "axios": "^0.21.4",
-    "jsforce": "^1.11.1",
+    "@openfn/language-common": "1.10.2",
+    "axios": "^0.21.1",
```
mtuchi, line 35: Also this was updated here -https://github.com/OpenFn/adaptors/pull/337

josephjclark:
Thanks for being vigilant @mtuchi !

I've checked out the package.json from main and removed yargs.  It should be up to date now.

mtuchi (approved):

mtuchi:
This looks good now, i am going to merge it

PR #334, opened 2023-08-03
Fix medicmobile

### LINKED ISSUE #328: medicmobile: update `query-string`
Update `medicmobile` to either use an updated `query-string` or to pin the version of `decode-uri-component`.

This closes https://github.com/OpenFn/adaptors/security/dependabot/46

josephjclark:
This PR bumps the `query-string` dependency of `medicmobile`.

Although `query-string` goes up several major versions, really its just because of keeping track with node versions. Looking over the history I don't think there's anything for us to worry about - we just get better quality and safer query string parsing.

Note: I originally wanted to just bump the affected library `encode-uri-component` - but adding the override didn't seem to work in my test project :thinking:  Looking over the `query-string` changes I'm sure this is a safe alternative.

This closes #328 and allows us to close the related security advisory.

### DIFF
--- a/.changeset/happy-worms-shake.md
+++ b/.changeset/happy-worms-shake.md
@@ -2,4 +2,4 @@
 '@openfn/language-medicmobile': patch
 ---
 
-Bump decode-uri-component
+Bump query-string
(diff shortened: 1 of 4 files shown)


mtuchi:
This looks good 👍🏽 @josephjclark

PR #338, opened 2023-08-03
Nexmo security patch

### LINKED ISSUE #329: nexmo: update to 3.x
Version 2.x of the nexmo adaptor uses jsonwebtoken 7.4.3, which has as security vulnerability.

We should update to the 3.x version. This is a major version bump so we need to check that signatures are all compatible.

Recommend we make this a minor version bump to the adaptor.

This closes:
* https://github.com/OpenFn/adaptors/security/dependabot/54
* https://github.com/OpenFn/adaptors/security/dependabot/21

mtuchi:
## Summary

Removed unused packages and update `nexmo` package to use the latest version `v2.9.1`. Also add `expandReferences` on `sendSMS` params. Fixes #329 

## Review Checklist

### DIFF
--- a/.changeset/forty-bees-sit.md
+++ b/.changeset/forty-bees-sit.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-nexmo': minor
+---
+
+- update nexmo to `v2.9.1``
+- expandReferences on sendSMS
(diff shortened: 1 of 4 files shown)


josephjclark (approved):
Nice catch on expand references

PR #339, opened 2023-08-04
`common` http helpers

mtuchi:
## Summary

Add new common util helper function for `http`.  Function added are 
- `request(method, fullUrl, options)`
- `get(url, options)  request(method, fullUrl, options)`
- `post(url, body, options) : request(method, fullUrl, options)`
- `del(url, body, options):  request(method, fullUrl, options)` ⚠️  **Could not use delete because it's a reserved word in js**

Related https://github.com/OpenFn/adaptors/pull/302

## Review Checklist

      migration guide followed?


packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+    method: method,
+    headers: headers,
+    errors: errors,
+    body: body ? JSON.stringify(body) : undefined,
+    throwOnError: false,
+  });
+
+  const responseBody = await readResponseBody(response);
+
+  if (response.statusCode >= 400) {
```
josephjclark, line 116: If I wanted to do: `errors: { 204: "content not found" }`, it wouldn't throw. 

You should write a unit test to prove this, then fix it.

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+  };
+}
+
+async function readResponseBody(response) {
+  const contentType = response.headers['content-type'];
+
+  if (contentType?.includes('application/json')) {
+    return await response.body.json();
+  } else {
+    const chunks = [];
```
josephjclark, line 143: We should be able to just `return response.body.text()` here.

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+    code: response.statusCode,
+    headers: response.headers,
+    data: responseBody,
+  };
+}
+
+async function readResponseBody(response) {
+  const contentType = response.headers['content-type'];
+
+  if (contentType?.includes('application/json')) {
```
josephjclark, line 140: We need to support some kind of option to force the response to be parsed as json. At the very least we should have a TODO in the code at this stage, but I'd rather see an API proposal and an implementation from you.

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+    return await response.body.json();
```
josephjclark, line 141: No need to await before returning

packages/common/src/util.js
josephjclark, line 116: I would probably push the error handling stuff out into another function, maybe something `assertResponse(response, options)`. The `assert` implies it'll throw.

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+  } else {
+    const chunks = [];
+    for await (const chunk of response.body) {
+      chunks.push(chunk);
+    }
+    return Buffer.concat(chunks).toString();
+  }
+}
+
+// Functions for different HTTP methods
```
josephjclark, line 151: No need for this comment.

Also I think you can make the signatures more lightweight, like this:
```
export const get = (url, options) => request('GET', url, options)

export const post = (url, body, options) => request('GET', url, { body, ...options })
```

No need to default options here by the way, because the `request` will do that for us

packages/common/test/util.test.js
```diff
@@ -1,5 +1,209 @@
...
+      .intercept({
+        path: '/api',
+        method: 'GET',
+      })
+      .reply(200, r => {
+        request = r;
+        return {};
+      });
+
+    await get('https://www.example.com/api');
```
josephjclark, line 27: I think we should allow the `baseUrl` stuff to be specified in the request options (rather than have to require the whole url to be passed every time)

Because then at the top of our test file we can do:
```
const options = { baseUrl: 'https://www.example.com' };
const client = enableMockClient(options.baseUrl);
```
And then in each test we can do:
```
await get('/api', options)
```

josephjclark (changes requested):
Hi @mtuchi, we're off to a solid start here, but I have a few suggestions for work to be done.

I also think we should move all the http stuff into a file at `src/util/http.js`, and export it from `src/util/index.js`. You'll also have to move the other util functions (either into index or into eg `references.js`). This new file structure should be reflected in the test folder. I think this gives us a much cleaner, less cluttered, more discoverable codebase.

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+
+export async function request(method, fullUrl, options = {}) {
+  const { baseUrl, path } = separateUrl(fullUrl);
+  const defaultOptions = {
+    headers: {},
+    query: undefined,
+    timeout: {},
+    tls: {},
+    body: undefined,
+    errors: {
```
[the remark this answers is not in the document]
mtuchi: Remove the default `errorMap`

packages/common/src/util.js
mtuchi: This was done from our pairing, we added `assertOK(response, errors, fullUrlOrPath);`

packages/common/src/util.js
mtuchi: Thanks 👍🏽 , improve the function signature here [3ffec9a](https://github.com/OpenFn/adaptors/pull/339/commits/3ffec9a5a43d478ac4599488f4a9964a4aa43412)

packages/common/test/util.test.js
mtuchi: Added support for `baseUrl` option

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+function separateUrl(fullUrl) {
+  const urlObject = new URL(fullUrl);
+  const baseUrl = `${urlObject.protocol}//${urlObject.host}`;
+  const path = urlObject.pathname;
+  return { baseUrl, path };
+}
+
+export async function request(method, fullUrl, options = {}) {
+  const { baseUrl, path } = separateUrl(fullUrl);
+  const defaultOptions = {
```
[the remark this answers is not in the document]
mtuchi: Moved default options outside the `request` function

packages/common/src/util.js
mtuchi: @josephjclark some thoughts on this 
How about we only support only 3 type of response `json`, `text` and `stream`
The default being `stream`.  Then we determine if `json` or `text` based on `content-type`

Because from an adaptor level if i need the response to be `json` i will pass in `Content-Type: application/json`. 
For response `text()`. We can have an array of Mime-Types then we can check if content-type contains an mime-type in that array. This can be the automagically part 
```
const contentType = getContentTypeFromHeaders()
const mimeTypes = ['application/vnd.ms-excel', 'text/csv', 'text/html' ...]

if(mimeTypes.contains(contentType)) await response.text()
```

Then if i don't specify `Content-Type: json` or i didn't specify `Content-Type` but the response has the mimeType that we support for `text()` we return `stream`

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+  };
+
+  if (!headers['Content-Type'] && body) {
+    headers['Content-Type'] = 'application/json';
+  }
+
+  const client = getClient(baseUrl);
+
+  const response = await client.request({
+    path: path,
```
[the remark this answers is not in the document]
mtuchi: Resolved here [4d3d643](https://github.com/OpenFn/adaptors/pull/339/commits/4d3d6435ce2b464cced302611f4f9311913f1ce2)

packages/common/src/util.js
```diff
@@ -49,3 +68,99 @@ export function normalizeOauthConfig(configuration) {
...
+      404: 'Not found',
+    },
+  };
+
+  const { headers, query, body, errors, timeout, tls } = {
+    ...defaultOptions,
+    ...options,
+  };
+
+  if (!headers['Content-Type'] && body) {
```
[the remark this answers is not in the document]
mtuchi: Made improvement and added a test here [9bd4192](https://github.com/OpenFn/adaptors/pull/339/commits/9bd4192584b21476286188c0176ff2d2811d3a52)

packages/common/test/util.test.js
```diff
@@ -1,5 +1,365 @@
...
+
+    await del('https://www.example.com/api');
+
+    // test the result
+    expect(request.method).to.eql('DELETE');
+  });
+});
+
+describe('options', () => {
+  it('should not throw 204 if response is okay using errorMap', async () => {
```
josephjclark, line 129: Inside a describe block, you should have a series of tests which tell a story. Starting from the simplest, most general case, each test should build on the last one by adding complexity, or changing an edge-case, or altering the input. Ideally each test should just change a single thing about the input,

Also, because all (almost all?) these tests are about the error map, the describe block should be called `errorMap`.

The tests should be something like this:
```
should not throw for a 200+ error
should not throw for a 300+ error
should throw for a 400+ error
should throw for a 400+ error
should throw for a 200 error if a mapping is provided
should throw a custom error for a 404
should use a function to generate an error message
should ignore mappings that don't apply
```

The nice thing about these tests is that you can see the progression in complexity and it explains the behaviour of the error map, In fact, the tests start to look like documentation!

packages/common/test/util.test.js
```diff
@@ -1,5 +1,365 @@
...
+      await await get('https://www.example.com/api/noAccess');
+    } catch (err) {
+      error = err;
+    }
+    expect(error.message).to.eql(
+      'Request to https://www.example.com/api/noAccess failed with status: 405'
+    );
+  });
+
+  it('should parse a response body as json if content-type is json', async () => {
```
josephjclark, line 335: Note that we'll probably soon need a describe block just for response parsing

packages/common/src/util.js
```diff
@@ -63,3 +82,137 @@ export function normalizeOauthConfig(configuration) {
...
+  try {
+    const { protocol } = new URL(fullUrl);
+    return protocol === 'http:' || protocol === 'https:';
+  } catch (_) {
+    return false;
+  }
+};
+
+/**
+ * The function `request` is an asynchronous function that sends HTTP requests and returns the response
```
josephjclark, line 141: It is nice to see good, thorough documentation here  - but remember this function won't be exposed directly to job writers, it won't be documented on docs.openfn.org

packages/common/src/util.js
```diff
@@ -63,3 +82,137 @@ export function normalizeOauthConfig(configuration) {
...
+const defaultOptions = {
+  timeout: 300, // Default to 300 seconds
+  headers: {},
+  query: undefined,
+  body: undefined,
+  errors: {},
+  tls: {},
+};
+
+const assertOK = (response, errorMap, fullUrl) => {
```
josephjclark, line 113: Maybe this should be called `assertResponseOk`

packages/common/src/util.js
```diff
@@ -63,3 +82,137 @@ export function normalizeOauthConfig(configuration) {
...
+    error.code = response.statusCode;
+    error.url = fullUrl;
+    throw error;
+  }
+};
+
+const isValidHttpUrl = fullUrl => {
+  try {
+    const { protocol } = new URL(fullUrl);
+    return protocol === 'http:' || protocol === 'https:';
```
josephjclark, line 134: Maybe `return protocol.startsWith('http')`

packages/common/src/util.js
```diff
@@ -63,3 +82,137 @@ export function normalizeOauthConfig(configuration) {
...
+
+  const response = await client.request({
+    path,
+    query,
+    method,
+    headers,
+    body: body ? JSON.stringify(body) : undefined,
+    throwOnError: false,
+    bodyTimeout: timeout,
+    connect: tls,
```
josephjclark, line 182: `connect` doesn't exist on `client.request`. The connect/tls stuff happens when we create a new Client.

packages/common/src/util.js
```diff
@@ -63,3 +82,137 @@ export function normalizeOauthConfig(configuration) {
...
+  }
+
+  const client = getClient(baseUrl);
+
+  const response = await client.request({
+    path,
+    query,
+    method,
+    headers,
+    body: body ? JSON.stringify(body) : undefined,
```
josephjclark, line 179: hmm, I think body parsing will be a bit more complicated than this because we may not always be posting json. Maybe its ok to leave it for now and come back to it later.

packages/common/test/util.test.js
```diff
@@ -1,5 +1,365 @@
...
+        return {};
+      });
+
+    const response = await get('/api', { baseUrl: 'https://www.example.com' });
+
+    expect(request.path).to.eql('/api');
+    expect(response.code).to.eql(200);
+  });
+
+  it('should use fullUrlOrPath', async () => {
```
josephjclark, line 49: Not a great test name - maybe `should accept an absolute url`. I would like the name to reference the previous test, because the two go hand-in-hand, so maybe both test names need thinking about together

packages/common/test/util.test.js
```diff
@@ -1,5 +1,365 @@
...
+    // set up mock POST at /
+    let request;
+
+    client
+      .intercept({
+        path: '/api',
+        method: 'POST',
+      })
+      .reply(200, r => {
+        request = r;
```
josephjclark, line 97: I know I drafted this pattern with you but I'm quickly going off it. It's a bit confusing in most cases. Mostly this pattern was created just suggest one way to test the `method` parameter of request.

I think we should set up an intercept a) with a unique URL and b) the things we want to test for in the intercept options.

If the request is correct and has the right headers, body, url, query etc, then the interceptor will match and we can just test the response code. Maybe the body too just to be sure that it's THIS interceptor that fired.

So this test would look more like this:
```
client
      .intercept({
        path: '/api/ali',
        method: 'POST',
      }).
     reply(200, 'ali-1')

const  { code, body } = await post('https://www.example.com/api/ali');

expect(code).to.eql(200)
expect(body).to.eql('ali-1');
```

Let's talk about this a bit more tomorrow.

packages/common/src/util.js
mtuchi: Added `parseAs` support

packages/common/test/util/http.test.js
```diff
@@ -0,0 +1,449 @@
...
+describe('options', () => {
+  it('should not throw 204 if response is okay using errorMap', async () => {
+    client
+      .intercept({
+        path: '/api/content',
+        method: 'GET',
+      })
+      .reply(
+        200,
+        {},
```
josephjclark, line 191: I don't think any of these error map tests need response data or headers? Can we check and remove if possible?

Test code should be as small and accurate as possible!

josephjclark (commented):
Great, nearly there, just a couple of small things :)

packages/common/src/util/http.js
```diff
@@ -0,0 +1,152 @@
...
+  });
+
+  assertOK(response, errors, fullUrlOrPath);
+
+  const responseBody = await readResponseBody(response, parseAs);
+
+  return {
+    code: response.statusCode,
+    headers: response.headers,
+    data: responseBody,
```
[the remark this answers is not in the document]
mtuchi: Body is more standard 👍🏽

packages/common/src/util/http.js
```diff
@@ -0,0 +1,152 @@
...
+
+  switch (parseAs) {
+    case 'json':
+      return response.body.json();
+    case 'text':
+      return response.body.text();
+    case 'stream':
+      return response.body;
+    default:
+      return contentType?.includes('application/json')
```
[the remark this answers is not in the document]
mtuchi: @josephjclark sometimes the `contentType` is `undefined`
But yeah i should use `contentType === 'application/json'` This is better

packages/common/test/util/http.test.js
mtuchi: Removed the headers and the test are still 👍🏽

PR #342, opened 2023-08-12
Fix streaming interface to parseCSV

josephjclark:
## Summary

This PR adds a couple of tiny fixes to support CSV streaming, plus a couple of unit tests to show the expected usage.

## Details

There were a couple of wee problems with the streaming API:

1. Expand references was trying to expand streams. Naughty. 
1. You can't pass `response.body` straight into `csv-parse`. response.body is a browser-compliant ReadableStream, but we need it to be a node Readable. Subtle. All we have to do is call `Readable.from` to convert the stream to a compatible API
1. The callback was always invoked after the final chunk - even if it's passing an empty chunk. Not a big deal but I thought I'd drop a fix.

There are tests in index.test.js which show streaming working from the filesystem and from unidici request. I have also tested this without a mock (ie, host a csv file locally with `npx serve`, pull it down with node native fetch, and parse it) - but obviously there's no use commiting that. I am satisfied that the mock exercises the same API.

@mtuchi I think you should probably merge this into your `msgraph` work for akf. You should be able to stream freely from there.


mtuchi (approved):
Thanks @josephjclark this looks good 👍🏽 , I have reviewed and tested the implementation, 
Tiny changes i added `import {Readable} from 'node:stream'` in `Adaptors.js` Since it's being used in `parseCsv` and the job was failing when i tried to use `parseCsv`
and i have remove `import {Readable} from 'node:stream'` in `test.js` since it was not being used.

PR #345, opened 2023-08-15
add readStream for reading stream response

mtuchi:
## Summary

## Details

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/msgraph/test/Adaptor.test.js
+++ b/packages/msgraph/test/Adaptor.test.js
@@ -1,12 +1,34 @@
 import { expect } from 'chai';
+import { setGlobalDispatcher } from 'undici';
+
 import { execute, getDrive, getFolder, getFile } from '../src/Adaptor.js';
 
 import MockAgent from './mockAgent.js';
 import { fixtures } from './fixtures.js';
 
-import { setGlobalDispatcher } from 'undici';
-
 setGlobalDispatcher(MockAgent);
+
+const readStream = async stream => {
+  const reader = stream.getReader();
+  let result = '';
+  // read() returns a promise that resolves
+  // when a value has been received
+  await reader.read().then(function processText({ done, value }) {
+    // Result objects contain two properties:
+    // done  - true if the stream has already given you all its data.
+    // value - some data. Always undefined when done is true.
+    if (done) {
+      return;
+    }
+
+    const chunk = new TextDecoder().decode(value);
+    result += chunk;
+    // Read some more, and call this function again
+    return reader.read().then(processText);
+  });
+
+  return result;
+};
 
 describe('execute', () => {
   it('executes each operation in sequence', done => {
@@ -359,7 +381,9 @@ describe('getFile', () => {
       getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW')
     )(state);
 
-    expect(finalState.data).to.eql(fixtures.itemContent);
+    const response = await readStream(finalState.data);
+
+    expect(response).to.eql(fixtures.itemContent);
   });
 
   it('should get a file metadata by id', async () => {
@@ -375,7 +399,9 @@ describe('getFile', () => {
     };
 
     const finalState = await execute(
-      getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW', { metadata: true })
+      getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW', {
+        metadata: true,
+      })
     )(state);
 
     expect(finalState.data).to.eql(fixtures.itemWithDownloadUrl);
@@ -397,7 +423,9 @@ describe('getFile', () => {
       getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW', { driveName: 'mydrive' })
     )(state);
 
-    expect(finalState.data).to.eql(fixtures.itemContent);
+    const response = await readStream(finalState.data);
+
+    expect(response).to.eql(fixtures.itemContent);
   });
 
   it('should get a file by path', async () => {
@@ -415,8 +443,9 @@ describe('getFile', () => {
     const finalState = await execute(getFile('/Sample Data/test.csv', {}))(
       state
     );
+    const response = await readStream(finalState.data);
 
-    expect(finalState.data).to.eql(fixtures.itemContent);
+    expect(response).to.eql(fixtures.itemContent);
   });
 
   it('should get a file metadata by path', async () => {


josephjclark:
Ah this is my bad, I didn't realise the tests were failing when I merged. Should have looked more closely!

This doesn't look right to me, let me look into it so we can get this wrapped quickly

packages/msgraph/test/Adaptor.test.js
```diff
@@ -1,13 +1,29 @@
...
 import { execute, getDrive, getFolder, getFile } from '../src/Adaptor.js';
 
 import MockAgent from './mockAgent.js';
 import { fixtures } from './fixtures.js';
 
-import { setGlobalDispatcher } from 'undici';
-
 setGlobalDispatcher(MockAgent);
 
+const readStream = async stream => {
```
josephjclark, line 11: I've got two problems here.

First off this is a really over-complicated way to resolve the stream. There are much easier ways!

Secondly, and more importantly, these tests are showing us that we've broken `getFile` in the adaptor. Rather than coding around the problem in the unit tests, we should address it properly because it's unusable in job code right now.

josephjclark:
So we're in this position because we changed the msgraph `request` to return a stream if it it wasn't requesting json. That was a bit of a stretch but I figured it was OK while it only affects our one workflow.

But I'm very nervous about leaving that functionality so broken.

A better fix for the issue would be to make getFile convert the response stream into text. Convert the return of `getFile` into this:
```
const url = getUrl(resource, apiVersion);
const auth = getAuth(accessToken);
const result = await request(url, { ...auth });
const content = response.pipeTo ? await response.text() : response;
return  handleResponse(content, state, callback);
```

The important bit is that the request function returns a stream, we convert it into text and return the text (like we used to).

After this change you should be able to revert your unit tests and everything will work.

Do we have similar problems with `getDrive` and `getFolder`? Or are they OK because they return json?

josephjclark (changes requested):
Sorry but this doesn't look right to me. See comments in the main thread!

mtuchi:
@josephjclark 
```
const response = await request(url, { ...auth });
const content = response.pipeTo ? await response.text() : response;
```

I am not sure if i understand `response.pipeTo` is this another way of checking if the response is stream ?

josephjclark:
@mtuchi Yes, that's a way to check if the response is a stream. It's more specialised than the `isStream` function, because it'll only ever refer to the ReadableStream returned by fetch. It's all a bit complicated but I think this is the simplest, most honest test.

You're right about the response though - the `text` function is on the response itself, not the body.  Hmm.

mtuchi:
> @mtuchi Yes, that's a way to check if the response is a stream. It's more specialised than the `isStream` function, because it'll only ever refer to the ReadableStream returned by fetch. It's all a bit complicated but I think this is the simplest, most honest test.
> 
> You're right about the response though - the `text` function is on the response itself, not the body. Hmm.

@josephjclark i have got it working with this approach 
```
const content = response.pipeTo
      ? await new Response(response).text()
      : response;
```

This way i don't have to change the response in util `request` function. Is this okay ?

josephjclark:
@mtuchi Alright, I really didn't want to mess with the request function too much - but to get the control we need we're just going to have to. This is a temporary solution until we sort out the http helpers.

First, in `utils/request`, the `params` object needs a key called `parseAs`.
```
const { headers, parseAs  } = params;
```
And we need to delete this key after reading it
```
delete params.parseAs;
delete params.headers;
```

Then, if `parseAs` is set, we use it to decode the response. If `parseAs` isn't set, we keep the old behaviour.
```
let content;
if (parseAs) {
   if (parseAs === 'json') content = await response.json()
   if (parseAs === 'text') content = await response.text()
} else {
  content = contentType?.includes('application/json')
    ? await response.json()
    : response.body
}
```

Now `getFile` needs to set `parseAs` to `text`. The other functions can all stay the same (I think)

None of this needs to be pretty  - it's all going to be removed soon anyway...

packages/msgraph/test/Adaptor.test.js
mtuchi: I have remove this approach in favor of using `parseAs`

mtuchi:
> @mtuchi your solution will break parseCSV!

@josephjclark 
I have pushed a update with the new approach that uses `parseAs` and add that option in `getFile()` only. Should i update the **changeset** to be a minor version bump ?

packages/msgraph/src/Adaptor.js
```diff
@@ -213,21 +213,21 @@ export function getFolder(pathOrId, options, callback = s => s) {
  * @return {Operation}
  */
 export function getFile(pathOrId, options, callback = s => s) {
+  const defaultOptions = {
```
josephjclark, line 216: Oh, I hadn't intended that `getFile` would expose `parseAs` to the job. Interesting, let me think about that!

Files usually have an extension and a mimetype. So when we request a file we should always be able to work out the type and return an appropriate data structure to represent it.

So I really don't think we should expose this option. Instead, we should hard-code `parseAs` into the request (a bit like how we pass auth differently).

Also, we're introducing `parseAs` to `request` as a temporary fix. But if we expose it in `getFile`, we're kinda committed to maintaining it in future. Which I'm not comfortable with.

packages/msgraph/src/Adaptor.js
```diff
@@ -254,9 +254,9 @@ export function getFile(pathOrId, options, callback = s => s) {
 
     const auth = getAuth(accessToken);
 
-    return request(url, { ...auth }).then(response =>
-      handleResponse(response, state, callback)
-    );
+    const response = await request(url, { ...auth, parseAs });
```
josephjclark, line 257: I would ideally like to touch this function as little as possible, can we revert most of the refactoring here and just do:
```
 return request(url, { ...auth, parseAs: 'text' }).then(...)
```

packages/msgraph/test/Adaptor.test.js
```diff
@@ -356,7 +355,7 @@ describe('getFile', () => {
     };
 
     const finalState = await execute(
-      getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW')
+      getFile('01LUM6XOGRONYNTZ26DBBJPTN5IFTQPBIW', { parseAs: 'text' })
```
josephjclark, line 358: This should not be passing `parseText`. I want the unit tests to be unaffected by this change.

josephjclark (changes requested):
Almost there, a couple of other comments. I was surprised to see changes to the `getFile` signature, that's not quite right.

Yes, I agree that this is a minor bump because we've extended the `request` API

packages/msgraph/src/Adaptor.js
mtuchi: @josephjclark `getFile()` can return metadata which is `json` or can return file content which in this case `text`
So if we pass `parseAs: text` the all test that returns metadata will fail unless in those test i pass 
`parseAs: json`. But i have thought of a quick fix which can be `parseAs: metadata ? 'json' : 'text',`
So the response will be like this.
```
const response = await request(url, {
      ...auth,
      parseAs: metadata ? 'json' : 'text',
    });
```

This way the function still behave the same and we don't have to update the tests

packages/msgraph/src/Adaptor.js
mtuchi: I like the idea of file mime type but i think it's going to be at-least a minor version bump and it needs planning

packages/msgraph/src/Adaptor.js
mtuchi: https://github.com/OpenFn/adaptors/pull/345/commits/cb97fff6d74e50192a7fa98fcf9951b3d54850ea

packages/msgraph/src/Adaptor.js
josephjclark: Oh I'm not suggesting we do any of the mime stuff now. I'm just saying we should return files as plain text for now!

packages/msgraph/src/Adaptor.js
```diff
@@ -254,9 +254,12 @@ export function getFile(pathOrId, options, callback = s => s) {
 
     const auth = getAuth(accessToken);
 
-    return request(url, { ...auth }).then(response =>
-      handleResponse(response, state, callback)
-    );
+    const response = await request(url, {
+      ...auth,
+      parseAs: metadata ? 'json' : 'text',
```
josephjclark, line 259: Good catch, thanks!

josephjclark (approved):

josephjclark:
@mtuchi This looks great, thanks for your efforts.

Before I merge, have we tested the workflow against this adaptor build? I just want to be sure we haven't accidentally broken anything

mtuchi:
> @mtuchi This looks great, thanks for your efforts.
> 
> Before I merge, have we tested the workflow against this adaptor build? I just want to be sure we haven't accidentally broken anything

@josephjclark 
Yes i have test my workflow but i am not using `getFile()` in my workflow, I am using `get()` and it's working fine as expected

PR #347, opened 2023-08-16
Bump node to 18.17.1

### LINKED ISSUE #346: Update node version
We need to ensure the node version in this repo is the latest LTS 18.17.1

josephjclark:
* Update node versions in asdf
* Update @types/node to match (only affects tooling)

No changeset as this is not a public facing change

Closes #346


mtuchi (approved):
Tested locally by running `asdf install` and `pnpm i & pnpm build` 
It seems to work as expected

PR #352, opened 2023-08-17
`mailchimp` Add generic helpers

### LINKED ISSUE #349: `mailchimp` generic helpers
### In Summary 
Provide low level HTTP operations for mailchimp.

These allow job writers to use the mailchimp HTTP API directly, rather than using our abstractions

## Operation Signatures

```
get(path, query?, callback): (state) => state | Promise<state>
post(path, data, query?, callback): (state) => state | Promise<state>

request(method, path, { query, data }, callback)
```
Note that path is like `/allowlists/add` - it is not a full URL.

## Details

`request` is the function which does all the work here.

Using `unidici.fetch` it should:
 * Build a URL using the domain in config and hard-coding the API to v3.0
 * Set auth headers (or whatever auth we need)
 * Build a querystring from the query object and append it to the URL (I think we use `querystring` elsewhere for this)
 * Extrac and parse the result (presumanly to json)
 * `composeNextState` on the response data (not the actual response, we ignore that (maybe later we can add it to references)).

Note that we do not expose headers to the job author.

If the response is not a 200, we should throw an error (don't throw the whole response, throw the error code, message and response data).

`get` and `post` are wrapping operations around request. They have no logic of their own, they just call `request`.

## Documentation

In each of these http operations, we should refer to mailchimp's API directly: https://mailchimp.com/developer/transactional/api

## Camelcasing

Note that we should not attempt to camelcase any query parameters (even if this contradicts other parts of the mailchimp API)

mtuchi:
## Summary
Using `undici` client, These changes provide a clean interface for making HTTP request to Mailchimp API. The changes includes: 
- Add `request()`, `get()` and `post()` helper functions. 
- Error handling for `statusCode => 400`
- remove `nock` in favor on `undici`.  So unit test we will be able to use `mockAgent` from `undici` 
- Update `chai` package to the `v4.3.6` for proper import use in test. `import { expect } from 'chai';`.The previous version you had to use `
import chai from 'chai';
const  { expect }  = chai
`

Fixes #349 

## Review Checklist

      migration guide followed?


packages/mailchimp/src/Utils.js
```diff
@@ -8,3 +10,58 @@ export function handleResponse(response, state, callback) {
...
+      method: resolvedMethod,
+      path: `/${apiVersion}${resolvedPath}`,
+      headers,
+      query,
+      body: body ? JSON.stringify(body) : undefined,
+    });
+
+    const responseBody = await response.body.json();
+
+    return handleResponse(responseBody, state, callback);
```
josephjclark, line 59: Actually I don't think we want to use `handleResponse` here.

Looking at the implementation it'll write the body to both `data` and `response`, so you'lll get a state object which looks like this:
```
{
    data: [/* mailchimp data */],
    response: [/* mailchimp data */]
}
```

packages/mailchimp/test/util.test.js
```diff
@@ -0,0 +1,44 @@
+import { expect } from 'chai';
+import { MockAgent, setGlobalDispatcher } from 'undici';
+
+import { request } from '../src/Utils.js';
+
+const mockAgent = new MockAgent({ connections: 1 });
+setGlobalDispatcher(mockAgent);
```
josephjclark, line 7: I suspect this doesn't work because the client doesn't use the global dispatcher.

josephjclark (changes requested):
Pretty close!

The important changes needed are:

* The new operations should live in the main adaptor.js file
* We should use the global `undici.request`, rather than creating our own client
* Don't re-use `handleResponse`, it's not quite right for us

packages/mailchimp/package.json
```diff
@@ -23,13 +23,14 @@
...
-    "md5": "^2.3.0"
+    "md5": "^2.3.0",
+    "undici": "^5.22.1"
   },
   "devDependencies": {
     "@openfn/buildtools": "workspace:^1.0.2",
     "@openfn/simple-ast": "0.4.1",
     "assertion-error": "^1.1.0",
-    "chai": "^3.5.0",
+    "chai": "^4.3.6",
```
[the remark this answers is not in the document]
mtuchi: Yes, the older version aka `v3.5.0`, you have to import chai like this 👇🏽 
```
import chai from 'chai';
const { expect } = chai;
```
Which is quite annoying, but with the latest chai update you can import `expect` like this 👇🏽 
```
```

packages/mailchimp/src/Utils.js
```diff
@@ -8,3 +10,58 @@ export function handleResponse(response, state, callback) {
...
+});
+
+const defaultOptions = {
+  query: {},
+  body: undefined,
+};
+
+export const request = (method, path, options, callback = s => s) => {
+  return async state => {
+    // TODO should we add apiVersion in credentials schema ? the default can be 3.0
```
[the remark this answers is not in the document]
mtuchi: So using the generic helpers, How can we add support to other apiVersions? or it's something we don't want to support for now ?

packages/mailchimp/src/Utils.js
```diff
@@ -8,3 +10,58 @@ export function handleResponse(response, state, callback) {
...
+      path,
+      options
+    );
+    const { query, body } = { ...defaultOptions, ...resolvedOptions };
+
+    const headers = setHeaders({ apiKey });
+
+    const baseUrl = `https://${server}.api.mailchimp.com`;
+
+    const client = createClient(baseUrl);
```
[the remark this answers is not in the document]
mtuchi: `extend execute to create the client before the job starts`
The adaptor is already doing that but it's using mailchimp node client. So adding this means the adaptor will have to support two clients

packages/mailchimp/src/Adaptor.js
```diff
@@ -312,6 +312,8 @@ export function listAudienceInfo(query, callback = s => s) {
   };
 }
 
+export { request, get, post } from './Utils';
```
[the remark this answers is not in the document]
mtuchi: I have added the functions in adaptor.s

packages/mailchimp/src/Utils.js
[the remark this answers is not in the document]
mtuchi: I have added support for `apiClient` and it get's clean-up at the end of execute

packages/mailchimp/src/Utils.js
mtuchi: I removed the handleResponse from `request()` function

packages/mailchimp/test/util.test.js
mtuchi: Yeah your are right, i have removed this

packages/mailchimp/package.json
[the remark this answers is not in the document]
josephjclark: Great, thanks!

packages/mailchimp/src/Utils.js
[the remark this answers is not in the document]
josephjclark: We don't want to support this right now. I think when there's a v4 api, we'll put out a new major adaptor version which supports v4 in the generic getters AND the underlying mail chimp client.

packages/mailchimp/src/Adaptor.js
```diff
@@ -312,6 +316,108 @@ export function listAudienceInfo(query, callback = s => s) {
...
+
+    assertOK(response, `https://${server}.api.mailchimp.com${urlPath}`);
+
+    const responseBody = await response.body.json();
+
+    const nextState = {
+      data: responseBody,
+      response: responseBody,
+    };
+    if (callback) return callback(nextState);
```
josephjclark, line 388: There will always be a callback because you've defaulted it in the paramters

packages/mailchimp/src/Utils.js
```diff
@@ -1,4 +1,21 @@
 import { composeNextState } from '@openfn/language-common';
+import { Client, MockAgent } from 'undici';
+
+const clients = new Map();
```
josephjclark, line 4: Good - but this is actually a bit over-complicated for mailchimp! There is only ever one `baseUrl` in the config so there's no need to store and maintain a map..

You should be able to do this (and your unit tests should continue to pass)
```
let client;

export const getClient = baseUrl => {
    if (client) {
       return client;
    }
    return  new Client(baseUrl);
}

export const enableMockClient = baseUrl => {
  const mockAgent = new MockAgent({ connections: 1 });
  client = mockAgent.get(baseUrl);
  return client;
};
```

It is still a slightly strange pattern and I'm going to have to think about it a bit more in http land. But it's fine for here today

josephjclark (commented):
Excellent - two small comments.

I'm impressed you got some unit tests running.

Have you been able to do any live tests at all? Do you have a mailchimp sandbox key or anything?

packages/mailchimp/src/Adaptor.js
mtuchi: ✅ 👍🏽  [cf1ac78](https://github.com/OpenFn/adaptors/pull/352/commits/cf1ac7886f49a949686770b63a00f4487d175099)

packages/mailchimp/src/Utils.js
mtuchi: ✅ [62ced82](https://github.com/OpenFn/adaptors/pull/352/commits/62ced82eb39655501c2cde9e747c39edfcf9daff)

josephjclark (approved):

Issue #353, opened 2023-08-17
MongoDB clusterUrl should be type hostname not URI


bensonmiller:
# Summary
The MongoDB adaptor (see `Adaptor.js`) expects the value of `credentials.clusterUrl` to be a simple hostname (e.g., `yourCluster-xxxyzzz.mongodb.net`), but the JSON schema that validates this field specifies `"format": "uri"`. In OpenFn Lightning, when attempting to create a MongoDB credential, it's impossible to save a credential that uses the correct hostname format because the schema verification fails, showing  an error: `expected to be a URI`.

# Details
In `packages/mongodb/src/Adaptor.js`, the `connect()` function constructs the correct URI for the MongoDB cluster as follows:

```
  const uri = `mongodb+srv://${encodeURIComponent(
    username
  )}:${encodeURIComponent(
    password
  )}@${clusterUrl}/test?retryWrites=true&w=majority`;
```

It's clear that `clusterUrl` is a simple hostname and not a full URI.

# Proposed fix
I'd propose two changes:

1. Update `configuration-schema.json` to change the format of `clusterUrl` property to `hostname`
2. Change the name of `clusterUrl` in the configuration to `clusterHostname` and update `Adaptor.js` to use the new configuration name.

PR #354, opened 2023-08-18
Update configuration schema for MongoDB adaptor. 

bensonmiller:
Fixes OpenFn/adaptors#353

## Summary
Fixes JSON schema verification for mongodb clusterUrl configuration 

## Details

In packages/mongodb/src/Adaptor.js, the connect() function constructs the URI for the MongoDB cluster as follows:

```
  const uri = `mongodb+srv://${encodeURIComponent(
    username
  )}:${encodeURIComponent(
    password
  )}@${clusterUrl}/test?retryWrites=true&w=majority`;
```

The adaptor expects the `clusterUrl` configuration property be provided as a hostname and not a full URI, but the current JSON schema specifies the format of this property as `uri`, which fails validation when providing a normal hostname. This PR changes the JSON schema to verify this property as the correct format (`hostname`, not `uri`). The PR also renames the property and fixes all references to the property in the adaptor and unit tests.

## Issues

#353 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/mongodb/CHANGELOG.md
+++ b/packages/mongodb/CHANGELOG.md
@@ -1,5 +1,20 @@
 # @openfn/language-mongodb
 
+## 1.2.0
+
+### Minor Changes
+
+- Update configuration schema for MongoDB adaptor:
+  - Rename `clusterUrl` to `clusterHostname`
+  - Change `clusterHostname` format from `uri` to `hostname`
+  - Update `Adaptor.js` and tests to use new name
+
+  The MongoDB adaptor configuration expects a hostname, not a strict
+  URI. The existing configuration prevented a user from saving a 
+  "correct" configuration in Lightning, due to the JSON schema verification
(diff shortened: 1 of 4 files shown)


bensonmiller:
@josephjclark Thank you for the deep dive here on the MongoDB adaptor. None of these changes would block me. I would note that I'm not deeply experienced with MongoDB, so my current uses are relatively unsophisticated. Nevertheless, I've looked at each of the issues you opened and they all look reasonable to me. If I were to rank these issues according to my own interest/priority:

HIGH:
Support databases not hosted on mongodb atlas (356)
Allow a database to be set in the credential (357)

MED:
Update client (359) -- mainly because it's low-effort

LOW:
Expose connection options (360)
Support TLS certs (358)

josephjclark:
Hi @bensonmiller, just to let you know that this should be live on Lightning. We did a major bump on the adaptor so you'll need to be using Mongo 2.0.0 (or latest)

PR #367, opened 2023-08-24
Watch mode

josephjclark:
This PR add a watch mode to build.

It only works for the `src` build (it doesn't work for dts, docs or ast). This is probably the most important as it's what you need when testing with the CLI.

To watch changes in salesforce, you can do:
```
pnpm -C package/salesforce build --watch
```
Or
```
cd packages/salesforce
pnpm build --watch
```

In watch mode, saving a file in `src` will cause the bundle in `dist` to be immediately regenerated.

Note that this won't detect changes in `common` if you're working on another adaptor - if working on common stuff you might want to set up two watches.

Closes #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,10 +277,19 @@ The `build` command accepts a list of build steps as arguments: `ast`, `src`,
 `docs` and `dts`. Calling build on an adaptor with no arguments will build
 everything.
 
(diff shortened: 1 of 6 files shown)


josephjclark:
@mtuchi Can you take a little look at this? I think it'll make your life a lot easier.

PR #373, opened 2023-08-30
rewrite `language-http`

### LINKED ISSUE #316: http: rewrite to use new helper functions
With the introduction of new http helpers based on `unidici`, we need to re-write `language-http` to use the new helpers.

It should also use the new `expandReferences` utility function.

At the time of writing, we will need to handle `expandReferences` if `options.body` is a stream. I think is it reasonable to add logic for this in the new `expandReferences` utility (ie, don't recurse into a stream object, which we should be able to duck type)

This is a major version bump as the `axios` export and `axios` APIs are being removed.

I would like to change the http signatures as little as possible I think - so we should still have `get(path, params, callback)`.

We will have to  go through `mapToAxiosConfig` and decide which properties to maintain and which to drop. I would prefer to support a much smaller API.

It may be helpful to review existing http job code to get a sense of how difficult migration is going to be.

### LINKED ISSUE #377: new common-http: can't post JSON
We've hit a problem in the new common http helpers.

Turns out we can't do `unidici.request({ method: 'POST', body: { hello: 'world' }})`

I though unidici would stringify the body for us, but it does not. We need to intervene.

I suggest we do the following:

* Add a `contentType` to the request options in the common helper. By default `contentType` is `json`. It can also be `form` or 'none',
* If the request has a `content-type` header, we IGNORE our content-type flag and do nothing automatically.
* Otherwise:
* * If contentType is `'json'`, and `body` has as truthy value, we set the content-type header to `application/json` and stringify the body
* * If contentType if `'form'`, and `body` has as truthy value, we set the content-type header to mulipart form (whatever that should be) and encode the form
* * If contentType is `none` (or falsy), we do nothing automatically

All this needs to happen in `common/util/http:request`. Where appropriate, wrapper functions should expose the `contentType` option.

Notes:
* Is `bodyType` better than `contentType`?
* Should it just be type?
* We now have a `contentType` for the request content, and `parseAs` for the response content. This doesn't feel so clean. What if we have `responseContent` and `requestContent` properties?
* If the request has a json content type header, and the body is not a string, we could try to stringify it.

mtuchi:
## Summary

Re-write language-http to use the new common helpers. Drops support for axios.

This introduces many breaking changes.

Closes #316 

Changes to http:

- [x] Remove axios export and axios APIs used
- [x] Use new request from common/util
- [x] Use new expandReference from common/util
- [x] map and fix existing unit tests
- [x] support TLS (cert details on  config)
- [x] Add auto parsing request body on [7463dbf](https://github.com/OpenFn/adaptors/pull/373/commits/7463dbf6d72e5d01242259033ed5f69e2b93a537), Fixes #377 JC WHAT IS THIS
- [x] write clean new response object to state
- [x] throw clean errors  
- [x] drop support for cookie handling with `tough-cookies`
- [x]  drop support for gzip

Also note some changes to common:
- [x] Add a status message along side a status code
- [x] Use `statusMessage` and `statusCode` instead of `message` and `code`

## QA Notes

Things we need to test in real job code:

* Each of the major handlers - including sending json data
* Error handling - logs and error objects 
* Forms and redirects
* Log output generally
* Ensure references are handled correctly

I'd also love to test some datasources which have real auth

## Review Checklist


josephjclark:
Comments from our discussion:

* Ok, let's break everything in favour of a nice clean API
* We need to keep `statusText`, if we can. So we need to change the underlying response to `{ code, headers, body, message }`

josephjclark:
@mtuchi I have spun https://github.com/OpenFn/adaptors/issues/377 out of the post issue. I suggest you discuss with Stu/Taylor next week and raise a fix to the underlying helper (you can do it on this branch)

packages/common/src/util/http.js
```diff
@@ -115,7 +121,7 @@ async function readResponseBody(response, parseAs) {
     case 'stream':
       return response.body;
     default:
-      return contentType === 'application/json'
+      return contentType && contentType.includes('application/json')
```
mtuchi, line 124: @josephjclark i have to use this approach because some of the response content-type are not exactly `application/json` 
For example there are cases you get a content-type `application/json; charset=utf-8`

josephjclark:
statusText: `undici` doesn't support this in `request`. We could use fetch, or we could add our own mappings. I recommend we drop this from the response for now, and add it back in if someone needs it. Note we do need to check error reporting and what kind of messaging we're getting

josephjclark:
maxRedirects: this is undocumented at the moment, but let's maintain the existing `maxRedirects` and `followAllRedirects`.

In http, both of these map to undici's maxRedirections  (maxRedirects just feeds the value through, and followAlllRedirects, maps to infinity) - which isn't documented properly but I think will work.

The common helper should only support `maxRedirects`, not `followAllRedirects`.

We need unit tests in common http, including NOT redirecting

josephjclark:
Re form: we'll take `form` as a json object and encode it - just like we do today in `mapToAxiosConfig`
```
  if (!isEmpty(formData)) {
    form = new FormData();
    Object.entries(formData).forEach(element => {
      form.append(element[0], element[1]);
    });

    const formHeaders = form.getHeaders();

    headers = { ...headers, ...formHeaders };
  }
  ```

**Comment by josephjclark** [comment]
@mtuchi  Well, I'm a bit stuck here.

1) Regarding redirect, it's weird, but so far as I can tell the redirect doesn't work if you use the mock interceptor. I've run a whole bunch of manual tests and the redirect works every time. For example, using the mock server we build, this is fine (and fails if we disable redirect):
```
  it.only('can follow redirects 2', async () => {
    const state = {
      configuration: {},
      data: {},
    };

      get('http://localhost:2222/r', {
        followAllRedirects: true,
      })
    )(state);

    console.log(finalState.headers);
    expect(finalState.data).to.eql('ok');
  });
  ```
I think we're going to have to leave the test alone. Maybe we can check in the test server and skip the test, then come back to it later.

2) Regarding form data, well this is also a weird one. First of all it turns out that if undici detects form data in the body, it'll encode it and update the headers automatically. See  https://github.com/nodejs/undici/pull/1332

So we should be able to just wrap the `form` JSON in FormData and send that straight through (we still need to create a formdata object, but we don't need to interfere with headers).

What I'd like to do is add a unit test which takes `form` input and checks that the appropriate headers are added to the request, without interrogating the response (which looks surprisingly difficult!). But I've had a bit of a hard time reproducing that.

josephjclark:
@mtuchi I've created two new issues here which I'm going to ask you to look into:

https://github.com/OpenFn/adaptors/issues/386
https://github.com/OpenFn/adaptors/issues/388

packages/common/src/util/http.js
josephjclark: That's fine - thanks for calling it out!

josephjclark:
Hi @mtuchi,

I'm afraid the new logging stuff is wrong and not what we agreed.

The common adaptor should return the duration of any request to the caller. It should not log anything.

The HTTP adaptor should then log according to the format we discussed, including the duration. I think for errors we should log in exactly the same way - but we should ALSO throw a nice error object to the console.

It's also really important that the `startTime` you capture is taken at the start of the request itself. At the moment you're recording it when the module is initialised, which means all the request durations are going to be wrong.

packages/http/src/Adaptor.js
```diff
@@ -43,87 +35,69 @@ export function execute(...operations) {
...
+  };
+}
 
-axios.interceptors.request.use(config => {
-  cookiejar?.getCookies(config.url, (err, cookies) => {
-    config.headers.cookie = cookies?.join('; ');
-  });
-  return config;
-});
+function handleCallback(state, callback) {
```
josephjclark, line 45: Here's a tip  - if you default the callback argument, this gets a lot easier
```
function handleCallback(state, callback = (s) => s) {
    return callback(state)
}
```
I actually think modern javascript has a few good approaches around this sort of pattern, including `return callback?.(state) ?? state`

packages/http/src/Adaptor.js
```diff
@@ -43,87 +35,69 @@ export function execute(...operations) {
...
-      data: {
-        ...extendableData,
-        __cookie: keepCookies?.length === 1 ? keepCookies[0] : keepCookies,
-        __headers: response.headers,
-      },
-    };
-  }
+      body = form;
+    } else {
+      body = body ? JSON.stringify(body) : null;
```
josephjclark, line 69: Try using simpler synax:
```
else if (body) {
   body = JSON.stringify(body)
}
```
Possible subtle bug: what if body is 0?

packages/http/src/Adaptor.js
```diff
@@ -43,87 +35,69 @@ export function execute(...operations) {
...
+    let body = resolvedParams?.body;
+    let headers = resolvedParams?.headers;
 
-      headers['set-cookie']?.forEach(c => {
-        cookiejar.setCookie(Cookie.parse(c), config.url, (err, cookie) => {
-          keepCookies?.push(cookie?.cookieString());
-        });
+    if (resolvedParams?.form) {
+      let form = new FormData();
+      Object.entries(resolvedParams?.form).forEach(element => {
```
josephjclark, line 63: I think we can simplify this loop too:
```
 let form = new FormData();
 for (const key of resolvedParams.form) {
     form.append(key, form[key])
 }
```
You shouldn't really need the optional chaining `?` operator on `resolvedParams`, certainly not inside the if block. Consider defaulting `resolvedParams` to {}

packages/http/src/Utils.js
```diff
@@ -1,148 +1,12 @@
-import FormData from 'form-data';
-import _ from 'lodash/fp';
-import safeStringify from 'fast-safe-stringify';
+// TODO add support for tls on config
```
josephjclark, line 1: Remember this TODO

packages/http/src/Adaptor.js
```diff
@@ -43,87 +35,69 @@ export function execute(...operations) {
...
+      baseUrl,
+      body,
+      maxRedirections,
+    };
 
-  return {
-    ...composeNextState(state, respWithCookies.data),
-    response: respWithCookies,
+    return commonRequest(method, resolvedPath, options)
+      .then(response => handleResponse(state, response))
```
josephjclark, line 89: Both handleResponse and handleCallback are so lightweight... it's almost MORE confusing to have the extra `.then()` steps. Unless they're re-used elsewhere,should we not just do it all here?

josephjclark (changes requested):
Some comments here in addition to the logging stuff.

Also, documentation! We've got a bunch of stuff now like `parseAs` and `errorMap` which we feed through to the underlying request. None of that is obvious, it's all new, an we can't just say "oh it's like axios" anymore.

So we're going to have to work out a documentation solution here. I'll have to open an issue for that.

josephjclark (changes requested):

josephjclark:
I've hit upon an annoying issue here. Maybe not substantial but disappointing.

It appears that undici has no API to return the actual API that was sent in a request.

So when we do `get('blah')`, the url in the `response` object is basically useless.

I'll have to dig a bit deeper but if we can't get the actual URL (which is so useful for debugging purposes) we'll have to remove that key from the response.

josephjclark:
Some changes I've instigated in common:

* Remove `url` from the response. We can't get the actual URL used, and us rebuilding what we think the url was is pointless. So I've removed it.
* Rename `status` and `message` on the response to `statusCode` and `statusMessage`, just because `message` is too ambiguous to be useful

Changes instigated in http:

* Restructure of the unit tests
* Move the request helper into Utils. Now `Adaptor.js` only contains operations, which is neat.

josephjclark:
I am thinking about changing the post signature to `post(url, data, options, callback)`. Ditto put and patch. This might be too big a break with legacy though?

packages/common/src/http.js
```diff
@@ -12,6 +12,8 @@ export { axios };
  * @param {object} value - data
  * @returns {Operation}
  */
+// TODO Deprecate or remove expandRequestReferences
```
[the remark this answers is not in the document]
josephjclark: No - it's part of removing common.http

PR #374, opened 2023-08-30
common: JSON schema validation

### LINKED ISSUE #340: Add JSON schema verification to common
**Summary**
I have a use case where I'd like to verify JSON input payload against a predefined JSON schema. Given that schema verification is a common requirement in data processing pipelines, it feels like schema verification might be an appropriate addition in the `common` adaptor. Implementation feels like it is best handled by a dedicated library (i.e., Ajv), so this would imply additional dependencies are added to package.json.  I am not a professional JS developer, but I am open to providing a PR on this extension, conditional on interest from OpenFn core team.

**Implementation concepts**
The simplest "hello world" implementation might be to pass through the `ajv.compile()` and `ajv.validate()` functions directly to the adaptor, but I could envision a single function (perhaps `jsonSchemaValidate(schema,data)`) that would handle the schema compilation and validation in one stroke. This is probably sufficient for most use cases, although it's not very efficient to compile the schema before every validation.  For context, I'm new to OpenFn and I'm not aware if there are opportunities to cache objects, such as compiled schemas; absent such capability, the efficiency point might be irrelevant.

**Request**
I'd appreciate comments on whether the OpenFn maintainers view JSON schema validation as sufficiently attractive to warrant inclusion in the common adaptor, or whether a different (TBD) approach is recommended. Thank you in advance for your thoughts.

josephjclark:
## Summary

This PR add a `validate(schema, data)` function to the common adaptor. It is designed to validate JSON against a JSON schema.

## Details

I won't explain the functionality at length here - docs in the PR should do that for me.

A couple of notes though:

* The function is designed to use `state.data` and `state.schema`, but you can override both values
* Schema can be a string URL to an externally hosted JSON schema
* We use `ajv` do the the actual validation. `ajv` is a code generator: given a schema, it'll generate a function to validate it. This should be safe in both the old and new runtimes. So far as I can tell the only security concern is poorly performing regex, which frankly means you're using a bad schema. `ajv` does not allow arbitrary code injection and does not look like a sandbox escape vector to me.
* The signature is carefully designed to play well with `each`
* There is some complication introduced by different schema versions - apparently we need to import. Here's the closest thing to [docs](https://ajv.js.org/guide/schema-language.html) on this (note the bit about the import for 2019). What is implemented here is about the only thing I could get to work - it's ugly but should be robust, and should support multiple versions running concurrently in the same job, should anyone need that.

## Example usage

Here's a job I've created and run against the CLI:
```
// set up data
fn(state => {
  return {
    ...state,
    data: [
      { firstName: 'joe', lastName: 'clark', age: 'never you mind' },
    ],
  };
});

fn(state => {
  const schema = {
    $id: 'https://example.com/person.schema.json',
    $schema: 'https://json-schema.org/draft/2020-12/schema',
    title: 'Person',
    type: 'object',
    properties: {
      firstName: {
        type: 'string',
        description: "The person's first name.",
      },
      lastName: {
        type: 'string',
        description: "The person's last name.",
      },
      age: {
        description:
          'Age in years which must be equal to or greater than zero.',
        type: 'integer',
        minimum: 0,
      },
    },
  };

  return { ...state, schema };
});

each('data[*]', validate());

```

## Issues

Closes #340

## Review Checklist

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -47,7 +47,8 @@
     "csvtojson": "^2.0.10",
     "date-fns": "^2.25.0",
     "jsonpath-plus": "^4.0.0",
-    "lodash": "^4.17.19"
+    "lodash": "^4.17.19",
+    "undici": "^5.22.1"
   },
   "devDependencies": {
     "@openfn/buildtools": "workspace:^1.0.2",
@@ -57,8 +58,7 @@
     "esno": "^0.16.3",
     "mocha": "9.2.2",
     "nock": "13.2.9",
-    "rimraf": "^3.0.2",
-    "undici": "^5.22.1"
+    "rimraf": "^3.0.2"
   },
   "main": "dist/index.cjs"
 }
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -41,6 +41,7 @@
   "type": "module",
   "types": "types/index.d.ts",
   "dependencies": {
+    "ajv": "^8.12.0",
     "axios": "1.1.3",
     "csv-parse": "^5.4.0",
     "csvtojson": "^2.0.10",
--- a/.changeset/brave-apes-provide.md
+++ b/.changeset/brave-apes-provide.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-common': minor
+---
+
+Add validate function to validate data against a JSON schema
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -5,6 +5,7 @@ import { JSONPath } from 'jsonpath-plus';
 import { parse } from 'csv-parse';
 import { Readable } from 'node:stream';
 import Ajv from 'ajv';
+import { request } from 'undici';
 
 import { expandReferences as newExpandReferences } from './util';
 
@@ -691,7 +692,7 @@ export function validate(schema = 'schema', data = 'data') {
 
     const resolvedData = resolveData();
     const resolvedSchema = await resolveSchema();
-
+    console.log(resolvedSchema);
     // TODO: warn if the schema doesn't have an id? Does it matter? Maybe, if you're using multiple id-less schemas
     const schemaId = resolvedSchema.$id || 'schema';
     if (!schemaCache[schemaId]) {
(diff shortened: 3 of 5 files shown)


josephjclark:
I've hit a tricky problem with schema versions :thinking: 

It looks from my testing like we need to specifically load the correct base schema version to match whatever schema the user wants to use.

For example, the default ajv export doesn't support the 2020-12 schema. I have to use a different export for that - and the 2020 export doesn't seem to support draft 7.

Theoretically this shouldn't really matter, but the structure of ajv is making this quite difficult. I'm going to have to noodle on it.

packages/common/package.json
```diff
@@ -41,12 +41,14 @@
...
   "dependencies": {
+    "ajv": "^8.12.0",
     "axios": "1.1.3",
     "csv-parse": "^5.4.0",
     "csvtojson": "^2.0.10",
     "date-fns": "^2.25.0",
     "jsonpath-plus": "^4.0.0",
-    "lodash": "^4.17.19"
+    "lodash": "^4.17.19",
+    "undici": "^5.22.1"
```
josephjclark, line 51: It's a bit annoying to carry undici and axios at the same time - but as axios will be removed before too long I think this OK to do.

taylordowns2000 (commented):
i think this looks good @josephjclark , but i'm a little thrown off by the idea of an operation that expects you to have previously prepared state in a certain way. seems like the usage (from your PR) is kinda like: 
```js
doThing(arg, arg)
fn(arg, arg) // returning state in some special state?
validate() // requiring state to look a certain way?
doMoreThings(arg, arg)
```
can they also use it in a more classic "helperFunction" way? :
```js
fn(state => {
  const data = { state }
  const schema = { SOMESCHEMA };
  if ( validate(data, schema) ) return state;
  throw "Invalid!"
})
```

josephjclark:
@taylordowns2000  You can do `validate(schema, data)` - it just uses keys on state by default. It's actually really cool to do use in conjunction with `each`, which I expect is a common use-case.

At the moment it's just an operation, so it needs state to be passed in. Your second example wouldn't work. I'm not quite sure I'd want it to.

A wider topic: I have been thinking about breaking down basically every function in to an Operation (used at the top-level of a job expression, expands references, must return state) and a Helper function (any arbitrary JS). That would fix your second use-case: you could choose to use the `validate(schema, data)(state)` Operation or the `validate(schema, data)` function, depending on your precise context. The new HTTP stuff is likely to use this pattern, but I didn't really feel like rolling it out here. 

I could do though, if you think we need to support the second case?

taylordowns2000:
I'm really interested in the "break down each function into an 'operation' and a 'helper function'" bit. But that's a nice for later. This looks great for now. Thank you. @bensonmiller, this should close your issue!

PR #391, opened 2023-09-13
`salesforce` Add `bulkQuery()` function

mtuchi:
## Summary

Add a `bulkQuery(query, callback)`. A function that allows user to query records in bulk  

## Review Checklist

      migration guide followed?


taylordowns2000 (approved):
this is great @mtuchi , thank you. as it's tested locally, working properly, and only introduces a _new_ operation, i'll merge now. (in the future, as you and joe point out, we need a better test strategy for these types of functions.)

PR #393, opened 2023-09-15
refactor `bulkQuery`

### LINKED ISSUE #389: add `bulkQuery` in salesforce
### Description

Using `query()` has a limitation of maximum result of 2,000 records, Since currently query does not support pagination when we have more than 2,000 results, It's very hard to query the remaining results using the same function (`query()`)

Salesforce Jsforce have a [Bulk Query API](https://jsforce.github.io/document/#bulk-query) which can fetches records in bulk in record stream, or CSV stream which can be piped out to a CSV file.

**Request in short summary**
Add `bulkQuery(query,options, callback)`

mtuchi:
## Summary

Improvements for `bulkQuery()` to use REST API which supports bulk v2 

Fixes #389 

## Review Checklist


mtuchi:
This is ready for review and merge, If improvements are needed will be addressed on https://github.com/OpenFn/adaptors/issues/394,  cc @aleksa-krolls @taylordowns2000

taylordowns2000 (approved):

PR #396, opened 2023-09-19
`msgraph` Add `submitXls`

mtuchi:
## Summary

## Details

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/msgraph/src/Adaptor.js
+++ b/packages/msgraph/src/Adaptor.js
@@ -1,4 +1,5 @@
 import xlsx from 'xlsx';
+import FormData from 'form-data';
 import { execute as commonExecute } from '@openfn/language-common';
 import { expandReferences } from '@openfn/language-common/util';
 
@@ -264,6 +265,14 @@ export function getFile(pathOrId, options, callback = s => s) {
   };
 }
 
+const defaultReq = {
+  fileName: `${new Date()}.xlsx`,
+  wsName: 'Sheet',
+  method: 'PUT',
+  contentType: 'application/vnd.ms-excel',
+  type: 'buffer',
+  bookType: 'xlsx',
+};
 /**
  * Convert form data to xls then submit.
  * @public
@@ -281,31 +290,47 @@ export function submitXls(req, data, callback) {
     const { accessToken, apiVersion } = state.configuration;
 
     const [resolvedRequest, resolvedData] = expandReferences(state, req, data);
-    const { method, name, path } = resolvedRequest;
-    // PUT https://graph.microsoft.com/v1.0/sites/{site-id}/drive/items/{parent-item-id}:/{file-name}:/content
+    const { method, wsName, path, fileName, type, bookType, contentType } = {
+      ...defaultReq,
+      ...resolvedRequest,
+    };
+
     const url = getUrl(path, apiVersion);
 
-    console.log(url);
     const auth = getAuth(accessToken);
+
+    console.log(resolvedData.length, 'data');
 
     const workbook = xlsx.utils.book_new();
     const worksheet = xlsx.utils.json_to_sheet(resolvedData);
-    const ws_name = name || 'SheetJS';
 
-    xlsx.utils.book_append_sheet(workbook, worksheet, ws_name);
+    console.log(worksheet, 'ws');
+
+    xlsx.utils.book_append_sheet(workbook, worksheet, wsName);
 
     // Generate buffer
-    const xlsxBlob = xlsx.writeFile(workbook, {
-      type: 'buffer',
-      bookType: 'xls',
-    });
+    const xlsxBlob = xlsx.write(workbook, { type, bookType });
 
+    const body = new FormData();
+    body.append('file', xlsxBlob, { filename: fileName });
+
+    // const blob = new Blob([xlsxBlob], { type: 'application/octet-stream' });
+    // body.append('file', blob, ws_name);
+
+    // body.append('file', xlsxBlob, ws_name);
+    // body.append('data', new File([xlsxBlob], 'sheetjs.xlsx'));
+
+    console.log(body, 'body');
     console.log('Posting to url: '.concat(url));
     // Upload the XLSX file
     return request(
       url,
       {
-        ...auth,
+        headers: {
+          ...auth.headers,
+          // 'Content-Disposition': `attachment; filename="${ws_name}.xlsx"`,
+          'Content-Type': contentType,
+        },
         body: xlsxBlob,
       },
       method || 'PUT'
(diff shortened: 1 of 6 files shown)


taylordowns2000 (changes requested):
hey @mtuchi , i think we should split `createXls` out into common and then import it from common. thoughts?

mtuchi:
> hey @mtuchi , i think we should split `createXls` out into common and then import it from common. thoughts?

@taylordowns2000 
Adding a function in `common` needs lot's of consideration on how that function will be used and must have unit tests, I suggest for now we update the function name to something that will be replaced with the function from common once it's implemented

`createAndUploadXls` seems close to what the function does

taylordowns2000 (changes requested):
@mtuchi , i think _this_ should be in common and called `createXls`.

in msGraph you'd still have your `upload` function.

thoughts?

if it's only used in a single adaptor, adding it to common doesn't require more planning than adding it here but it establishes the correct pattern and opens the door for future refactoring.

mtuchi:
> @mtuchi , i think _this_ should be in common and called `createXls`: https://github.com/OpenFn/adaptors/pull/396/files#diff-a4d87515408a6dcfa3bd136a1edb1ad3ffe97d0570b37c9f1a57bddb0510c9f8R302-R314
> 
> in msGraph you'd still have your `upload` function.
> 
> thoughts?
> 
> if it's only used in a single adaptor, adding it to common doesn't require more planning than adding it here but it establishes the correct pattern and opens the door for future refactoring.

@taylordowns2000 We still need to add unit test for that and properly document supported options for that, 
It's best having this function in `msgraph/Util.js` then later on move it to common with proper design thinking

packages/msgraph/src/Adaptor.js
```diff
@@ -23,8 +29,14 @@ export function execute(...operations) {
   };
 
   const cleanup = finalState => {
-    const { drives, ...rest } = finalState;
-    return rest;
+    if (finalState?.buffer) {
```
josephjclark, line 32: I actually prefer the old pattern here:
```
const { drives, buffer,  ... rest } = finalState;
return rest;
```

Except actually I'd do like this:

```
const cleaup = (state) => {
  const { drives, buffer,  ... finalState } = state;
  return finalState;
}
```

packages/msgraph/src/Utils.js
```diff
@@ -135,3 +131,67 @@ export const request = async (urlString, params = {}, method = 'GET') => {
...
+ * sheetToBuffer('$.data[*]', {
+ *  wsName: 'Invalid Grant Codes',
+ *  bookType: 'xlsx',
+ * });
+ * @param rows - The `rows` parameter is an array of objects representing the data to be written to the
+ * Excel sheet. Each object in the array represents a row in the sheet, and the keys of the object
+ * represent the column headers. The values of the object represent the data in each cell of the row.
+ * @param options - The `options` parameter is an object that contains additional configuration options
+ * @param {String} [options.wsName] - Worksheet name i.e 32 Characters
+ * @param {String} [daoptionsta.bookType] - File format of the exported file, Default is 'xlsx'. See {@link https://docs.sheetjs.com/docs/api/write-options/#supported-output-formats here}
```
josephjclark, line 160: typo here on the param name

packages/msgraph/src/Utils.js
```diff
@@ -135,3 +131,67 @@ export const request = async (urlString, params = {}, method = 'GET') => {
...
+ * and worksheet using the rows, appends the worksheet to the workbook, and returns the workbook as a
+ * buffer.
+ * @public
+ * @example
+ * <caption>Create a buffer containing excel file with `xlsx` output format  </caption>
+ * sheetToBuffer('$.data[*]', {
+ *  wsName: 'Invalid Grant Codes',
+ *  bookType: 'xlsx',
+ * });
+ * @param rows - The `rows` parameter is an array of objects representing the data to be written to the
```
josephjclark, line 155: The docs on this parameter aren't very clear. Also it looks lie rows takes a string with a path to "an array objects...", rather than an array of objects? Can we pass data directly?

I'm happy to wave this through as its just in this adaptor. It's the kind of thing that would get a lot more scrutiny if we were to merge it into common.

josephjclark: Also one of the strange things about this function is that it doesn't really take a sheet. It looks like it just takes an array of json objects, converts them into a sheet (actually in your job aren't you converting into an array?) then turning THAT into a buffer ready for upload.

Again, because it's not going into common I'm happy to not look too closely at this

packages/msgraph/test/Adaptor.test.js
```diff
@@ -452,3 +458,42 @@ describe('getFile', () => {
     });
   });
 });
+
+describe('uploadFile', () => {
+  it.skip('should convert array of object to excel and post to specified path', async () => {
```
josephjclark, line 463: Is this supposed to be skipped?

josephjclark (commented):
This is fine, I've left a couple of comments. The fixes here are very closely bound to your job. I'd like to be a bit more critical but I just don't think we have the time.

Re common vs msgraph - just to be clear, I think Taylor's broadly right in that the xlsx stuff has nothing to do with msgraph and is a candidate for common. But I also think that there's a big design overhead on that and I don't want to take it on now. It's fine for the moment to add a helper function here.

packages/msgraph/src/Adaptor.js
mtuchi: @josephjclark , that looks clean but you will into an error saying `can not destructure undefine`  if `drives` or `buffer` is not in state. That why i had to check if i have the key `buffer` or `drives` in state before removing it

packages/msgraph/src/Utils.js
mtuchi: Your right the function takes an Array of Objects Data then convert it to buffer ready for upload, What shouldbe best name for this function @josephjclark

packages/msgraph/test/Adaptor.test.js
mtuchi: Yes, Since i need to cut a release soon, i have open an issue for it https://github.com/OpenFn/adaptors/issues/397

packages/msgraph/src/Utils.js
josephjclark: Well what it does it `jsonToSheet`, it's just that the sheet is then represented as a buffer.

Let's leave the name for now - this function is really designed just for this one job and I don't know how well it'll scale beyond that.

We do need to fix the docs though - `[daoptionsta.bookType]`

josephjclark: Oh you did that already I see :+1:

PR #399, opened 2023-09-26
Hive

### LINKED ISSUE #398: Add `hive` adaptor
### Summary 
Add `query` function that can be used to execute spark queries on Apache Hive Database

mtuchi:
## Summary

Add a `query` function. The function will execute Spark SQL queries when connected to Apache hive database. Fixes #398 

**Currently unit test are skipped and will be addressed by this issue #400**

## Review Checklist

- [ ] Are there any unit tests? Should there be? (skipped)


packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,155 @@
...
+      cleanupState
+    )({
+      ...initialState,
+      ...state,
+    });
+  };
+}
+
+function connect(state) {
+  //TODO @Mtuchi how to use the databse?
```
taylordowns2000, line 45: looks like this should be removed

taylordowns2000 (changes requested):
looks great. there's a TODO to remove, but otherwise no changes required

packages/hive/src/Adaptor.js
mtuchi: Done and added comment in improvement issue https://github.com/OpenFn/adaptors/issues/400#issuecomment-1735734535

mtuchi:
@taylordowns2000 done ✅ , comment removed

taylordowns2000 (approved):
@mtuchi , i have cleaned up the README and CHANGELOG. also adjusted the query used in the docstring. all good from my side!

packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,154 @@
...
+  composeNextState,
+} from '@openfn/language-common';
+import { expandReferences } from '@openfn/language-common/util';
+
+import hive from 'hive-driver';
+
+const { TCLIService, TCLIService_types } = hive.thrift;
+const utils = new hive.HiveUtils(TCLIService_types);
+
+const logger = message => console.log(message);
```
josephjclark, line 12: Really don't think we should be aliasing `console.log` like this. It doesn't really save much time or typing, it looks weird when you come into the adaptor for the first time, it doesn't match the signature of `@openfn/logger`, and it doesn't behave the same as `console.log` :thinking: 

Sorry to be a pain but can we remove it?

packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,154 @@
...
+      ? new hive.auth.PlainTcpAuthentication({
+          username,
+          password,
+        })
+      : new hive.auth.NoSaslAuthentication();
+
+  const connect = client.connect({ host, port }, connection, auth);
+
+  logger('Connect client');
+  return { ...state, connect };
```
josephjclark, line 61: Let's write `connect` to a var, rather than adding it to state. I think we agreed a while back that writing secret stuff to state was probably a bad idea?

josephjclark: Also, more importantly, we have to wait for the `client.connect` promise to resolve before we can carry on executing. So you have to do 

```
const connect = await client.connect({ host, port }, connection, auth);
 ```

File: packages/hive/src/Adaptor.js [hunk 4]
```diff
@@ -0,0 +1,154 @@
...
+ * @returns {Operation}
+ */
+export function query(qs, options, callback) {
+  return async state => {
+    const [resolvedQs, resolvedOptions] = expandReferences(state, qs, options);
+    const { connect } = state;
+
+    return connect
+      .then(async client => {
+        const session = await client.openSession({
```
josephjclark, line 94: Are we sure we want to open a new session on each query? Should we not start a session in `connect` and re-use it throughout the job?

I don't know much about Hive but my assumption is that the whole point of a session is to make multiple queries more efficient.

That also makes the `query` code simpler, because of lot of the code here really has little to do with querying. The clue is that you offload the query logic to `execSql`...

packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,154 @@
...
+      : new hive.auth.NoSaslAuthentication();
+
+  const connect = client.connect({ host, port }, connection, auth);
+
+  logger('Connect client');
+  return { ...state, connect };
+}
+
+function disconnect(state) {
+  return state.connect.then(async client => {
```
josephjclark, line 65: This looks wrong - we shouldn't wait for the `connect` promise to resolve here.

See other comments, but once the client (and session ?) have been close, this should just be `await client.close`.

Also I don't think you need `disconnect` AND `cleanupState` - it's a bit confusing to have both. Let's just keep `disconnect` so that our execute looks like this:
```
return commonExecute(
      connect,
      ...operations,
      cleanupState
    )
```

packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,154 @@
...
+            TCLIService_types.TProtocolVersion.HIVE_CLI_SERVICE_PROTOCOL_V10,
+        });
+
+        logger(`Executing query: ${resolvedQs}`);
+
+        const result = await execSql(session, resolvedQs, resolvedOptions);
+
+        logger('Success... ✔');
+        logger(
+          `Retrived ${result.length} result${result.length > 1 ? 's' : ''}`
```
josephjclark, line 105: Typo: `Retrieved`.

Also, just for the record, I wouldn't worry about pluralisation in a low-level log output like this. If anyone gets upset that our logger says `Retrieved 1 items`, send them to me. It's not worth removing the logic at this point though, so leave it :)

packages/hive/src/Adaptor.js
```diff
@@ -0,0 +1,154 @@
...
+        const nextState = {
+          ...composeNextState(state, result),
+          result,
+        };
+
+        if (callback) callback(nextState);
+        return nextState;
+      })
+      .catch(error => {
+        console.error(error);
```
josephjclark, line 119: Why are we logging the error twice?

josephjclark (changes requested):
Bit worried about some of the client/connection logic!

packages/hive/src/Adaptor.js
taylordowns2000: no that's a good catch. @mtuchi , what was the intention here?

mtuchi: @josephjclark i was testing `@openfn/logger` then i ran into export errors. So my intention was to keep `logger` as a pattern that can be used to standardize how we log stuff in our adaptors. 

It's fine we can remove it

packages/hive/src/Adaptor.js
mtuchi: Removed the catch statement

packages/hive/src/Adaptor.js
mtuchi: put session as a global variable

packages/hive/src/Adaptor.js
mtuchi: put client as a global variable

josephjclark (approved):
Looks much better, thanks!

Issue #407, opened 2023-10-10
salesforce: Bulk V2 api support for `bulk()` data import function


aleksa-krolls:
## Adaptor
salesforce

## Background
Recently in some Salesforce projects, when we're connecting to sandboxes that are in the latest version of Salesforce... we're finding that bulk import jobs aren't closing when using our adaptors `bulk()` function that connects to the Bulk API V1. We're seeing way more timeouts than we're used to, so we want to upgrade our adaptor to work with the Bulk API V2 to see if that improves job performance. 
<img width="1220" alt="Screen Shot 2023-10-10 at 10 51 33 PM" src="https://github.com/OpenFn/adaptors/assets/16758106/a640a3df-35b0-4170-9fd5-d0f59b862b7c">
^these jobs remain "in progress" for a very long time and take forever to process, and then our OpenFn runs [timeout](https://www.openfn.org/projects/wfwi-committed-giving-staging/runs/06525aa1-78e9-7a57-800a-f79219183b48). 

## Request
Please develop a new `bulk()` function that leverages the Salesforce Bulk V2 api. 
- Users should be able to upsert/insert/update records in bulk with this function.
- I don't think JSForce supports this new API version, so I think we will need to chain a series of http requests to deliver this new function (similar to how we implemented the `bulkQuery()` function. 
- Consider that we do have several existing implementations that use the Bulk API V1 function we currently have, so ideally this new version behaves just like the existing `bulk()` function and has the same options. The only main difference is that we should be leveraging the Bulk V2 api. 

See Salesforce documentation: https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/bulk_common_diff_two_versions.htm

## To test
- You can use [this Message](https://www.openfn.org/projects/wfwi-committed-giving-staging/messages/0651c6d3-9cbc-7c57-b2c7-df753008f6c7) and this job: https://github.com/OpenFn/women-for-women/blob/staging/jobs/committed/upsertCardMasterV2.js
- sandbox: see LP `WFW Full Sandbox Sept 2023`

aleksa-krolls:
@mtuchi we are running into several timeouts and performance issues on WFWI... I think this will help. Can you take this on tomorrow and sync with @daissatou2 if any questions/ or if you need help testing?

josephjclark:
Hi, so this is all a bit complicated (ain't it always)

[...]

PR #411, opened 2023-10-13
`salesforce` BulkV2 support

mtuchi:
## Summary

Add a `bulk2` function that creates and execute a bulk job using Salesforce Bulk API v2. Also updated the `bulkQuery` function to use jsforce `bulk2.query` function

## Details

I have updated the `jsforce` dependency to `v2.0.0-beta.28`. This version supports Salesforce Bulk V2.
For the implementation of `bulk2` behind the scene we are using [bulk2.loadAndWaitForResult](https://jsforce.github.io/jsforce/classes/api_bulk.BulkV2.html#loadAndWaitForResults) function and for `bulkQuery` we are using [bulk2.query()](https://jsforce.github.io/jsforce/classes/api_bulk.BulkV2.html#query)

Ref #407 
## Review Checklist

- [x] Update changelog and version

### DIFF
--- a/packages/salesforce/CHANGELOG.md
+++ b/packages/salesforce/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @openfn/language-salesforce
 
+## 5.0.0-alpha
+
+### Major Changes
+
+- Update jsforce dependency [29cce5b]
+- Add `bulk2` function [97d3729]
+- Update `bulkQuery` function to use jsforce `bulk2.query` [97d3729]
+
 ## 4.2.2
 
 ### Patch Changes
--- a/packages/salesforce/CHANGELOG.md
+++ b/packages/salesforce/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 5.0.0-alpha
 
+> This is an experimental alpha release based on jsforce beta 2. It is not
+> recommended for production use.
+
 ### Major Changes
 
 - Update jsforce dependency [29cce5b]
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -188,16 +188,12 @@ export function query(qs) {
  * @function
  * @param {String} qs - A query string.
  * @param {Object} options - Options passed to the bulk api.
- * @param {integer} [options.pollTimeout] - Polling timeout in milliseconds.
- * @param {integer} [options.pollInterval] - Polling interval in milliseconds.
+ * @param {integer} [options.pollTimeout] - Polling timeout in milliseconds. Default 3000
+ * @param {integer} [options.pollInterval] - Polling interval in milliseconds. Default 9000
  * @param {Function} callback - A callback to execute once the record is retrieved
  * @returns {Operation}
  */
 export function bulkQuery(qs, options, callback) {
-  const defaultOptions = {
-    pollTimeout: 90000, // in ms
-    pollInterval: 3000, // in ms
-  };
   return async state => {
     const { connection } = state;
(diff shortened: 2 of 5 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -395,13 +328,111 @@ export function bulk(sObject, operation, options, fun) {
...
+ * @param {String} sObject - API name of the sObject.
+ * @param {String} operation - The bulk operation to be performed
+ * @param {Object} options - Options passed to the bulk api.
+ * @param {Function} fun - A function which takes state and returns an array.
+ * @returns {Operation}
+ */
+export function bulk2(sObject, operation, options, fun) {
+  return async state => {
+    const { connection } = state;
+    const { failOnError, allowNoOp, pollTimeout, pollInterval } = {
```
josephjclark, line 359: Bit of a wider issue this.... but this defaulting pattern has been spreading through the codebase. I really don't like it :( It always feels like a heavyweight solution to a simple problem.

I think it's time to put a stop to it.

Instead of the separate variable, the destructure and the spread, we can just do this:
```
const {
  failOnError = true,
  allowNoOp = true,
  pollInterval = 6000,
  pollTimeout = 240000,
  }  = options;
```

packages/salesforce/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # @openfn/language-salesforce
 
+## 5.0.0-alpha
```
josephjclark, line 3: Maybe we can add a single sentence here, between the version and `Major Changes`, like this:

> This is an experimental alpha release based on jsforce  beta 2. It is not recommended for production use.

josephjclark (changes requested):

packages/salesforce/src/Adaptor.js
mtuchi: I think what you have laid out is a good pattern, but that doesn't work if in a job code, a user does not use the optional params for example `pollInterval`.

For example my job code 
```
bulk2(
  'Some_sObject'
  'upsert', 
  {
    extIdField: 'External_ID',
    failOnError: true,
    allowNoOp: true,
  },
  state => state.data
);
```
Produce this error 
```
TypeError: Cannot set properties of undefined (setting 'pollInterval')
```

josephjclark: Hmm, pretty suspicious of that. There's no reason it shouldn't work in the adaptor and runtime, and indeed it seems to work for me in CLI. Can we pair on this?

packages/salesforce/src/Adaptor.js
```diff
@@ -395,13 +321,117 @@ export function bulk(sObject, operation, options, fun) {
   };
 }
 
+const defaultBulkOptions = {
```
josephjclark, line 324: This is unused

packages/salesforce/src/Adaptor.js
```diff
@@ -395,13 +321,117 @@ export function bulk(sObject, operation, options, fun) {
...
+                res.unprocessedRecords.length
+              );
+              return res;
+            }
+          });
+      })
+    );
+
+    console.log('Merging results arrays.');
+    const merged = [].concat(...results);
```
josephjclark, line 424: I am not sure what all this merging and concat'ing is trying to achieve!

Try dropping this line and just returning like this:
```
return { ...state, references: merged.concat(state.references) }
```

josephjclark (changes requested):

josephjclark (approved):

PR #413, opened 2023-10-18
Salesforce Experiment

josephjclark:
This is an experimental PR for Salesforce 5.0.0-alpha

The package has been released from this branch with tag @next. But changes should not be merged.

The release is considered unstable and should not be used in production. It is specifically targeted at fixing #407

### DIFF
--- a/packages/salesforce/CHANGELOG.md
+++ b/packages/salesforce/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @openfn/language-salesforce
 
+## 5.0.0-alpha
+
+> This is an experimental alpha release based on jsforce beta 2. It is not
+> recommended for production use.
+
+### Major Changes
+
+- Update jsforce dependency [29cce5b]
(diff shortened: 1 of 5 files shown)


josephjclark:
Hey @mtuchi We need to do something about the version number here.

We've learned a bit more about how platform works and basically if a job runs with @latest, it'll use the _latest locally installed version_. It doesn't respect the @latest tag.

That's a problem because jobs may accidentally drift onto it. This might be fine but I consider this build unstable I don't  want anyone other than WFWI using it.

I think we should pull 5.0 and re-release as `0.0.0` or something crazy like that. No-one would  willingly install a 0.0.0 version.

mtuchi:
@josephjclark sounds good, i can un-publish `v5.0.1` and publish `v0.0.0`. Something to note even if you un-publish  `v5.0.1` You can't release another `v5.0.1`.

Issue #406, opened 2023-10-10
new http: create and test against a https server


josephjclark:
* Create a standalone koa server
* Generate SSL certs https://www.leohuynh.dev/blog/use-https-in-local-development/
* Add tls to it with `koa-sslify` https://www.npmjs.com/package/koa-sslify
* Create a job which calls the server
* The job should fail if certs aren't added to config

mtuchi:
After following the tutorial for generating the SSL certs, here is my directory structure 
```
├── certs
│   ├── server.crt
│   └── server.key
├── index.js
├── node_modules
│   ├── some libraries
├── package-lock.json
├── package.json
├── rootCA.key
├── rootCA.pem
├── rootCA.srl
├── server.csr
├── server.csr.cnf
├── server.js
├── job.js
├── tmp
│   ├── output.json
│   └── state.json
└── v3.ext
```
`server.js`
```
let Koa = require("koa");
let app = new Koa();
const sslify = require("koa-sslify").default;

app.use(sslify());

app.use(async (ctx) => {
  console.log(ctx);
  ctx.body = "Hello World";
});

module.exports = app;
```

`index.js`
```
let app = require("./server");
let https = require("https");
let fs = require("fs");
let path = require("path");

let certOptions = null;
try {
  certOptions = {
    key: fs.readFileSync(path.resolve("certs/server.key")),
    cert: fs.readFileSync(path.resolve("certs/server.crt")),
  };
} catch (err) {
  console.log("No certificate files found!");
}

let host = process.env.APP_URL || "localhost";
let isLocal = host === "localhost";
let enableHTTPSInLocal = Boolean(isLocal && certOptions);

let port = enableHTTPSInLocal ? 443 : process.env.PORT || 3434;
let protocol = isLocal && !certOptions ? "http" : "https";

let url = `${protocol}://${host}${isLocal ? `:${port}` : ""}`;

let callback = () => {
  console.log(`App start successfully at ${url}`);
};

if (enableHTTPSInLocal) {
  https.createServer(certOptions || {}, app.callback()).listen(port, callback);
} else {
  app.listen(port, callback);
}
```

`package.json`
```
{
  "dependencies": {
    "koa": "^2.14.2",
    "koa-multer": "^1.0.2",
    "koa-sslify": "^5.0.1"
  },
  "scripts": {
    "start": "node index.js"
  }
}
```

`job.js`
```
fn((state) => ({
  ...state,
  tls: {
[...]

Issue #402, opened 2023-10-06
http: add integration tests


josephjclark:
We should add a suite of integration tests in the new http adaptor.

This creates a server and uses the http functions to call it.

As well as some very basic get/put/post stuff, it can test redirects, form data, auth and tls security - stuff that's harder to unit test with the undici mocks

--

Here's a unit test for http which uses redirects:
```
 it('can follow redirects', async () => {
    // mock intercept paths
    // note that the mock undici client won't respect the redirect
    testServer
      .intercept({
        path: '/api/fake-endpoint',
        method: 'GET',
      })
      .reply(
        301,
        1,
        {
          headers: {
            Location: 'https://www.example.com/api/fake-endpoint-2',
          },
        }
      );

    testServer
      .intercept({
        path: '/api/fake-endpoint-2',
        method: 'GET',
      })
      .reply(302, 2, {
        headers: {
          Location: 'https://www.example.com/api/fake-endpoint-3',
        },
      });

    testServer.intercept(
      {
        path: '/api/fake-endpoint-3',
        method: 'GET',
      }).reply(200, res => res.path)
    );

    const state = {
      configuration: {},
      data: {},
    };

    const finalState = await execute(
      get('https://www.example.com/api/fake-endpoint', {
        followAllRedirects: true,
      })
    )(state);

    expect(finalState.data.url).to.eql('/api/fake-endpoint-3');
  });
```

This test doesn't work against the undici mock interceptor because the redirect flag is not respected.

To get this test to pass, we should:
* create a lightweight http server in test
* add redirect routes to the server
* call the actual server
* test the result
* tear down the server

It's  kind of a big test and not worth the time right now.

Here's a koa server we can use:
```
import Koa from "koa";

const server = new Koa();

server.use(async (ctx) => {
  if (ctx.url === "/redirect") {
    ctx.response.status = 301;
    ctx.response.set({ Location: "http://localhost:2222/foo" });
  } else {
    ctx.response.status = 200;
    ctx.response.body = "ok";
  }
});

server.listen(2222);
```

PR #415, opened 2023-10-19
integration tests

mtuchi:
## Summary

Add integration tests. These tests are using nodejs http and https modules for creating a server that is used in tests. 
The http server is used for testing redirects, while the https server is used for testing the `tls` support. 
In addition the https server is also using koa and koa-sslify packages.

You will notice a `test/helpers/` directory which contain `certs.js`. These are the certificates that we are using for the https server for testing 

Ref #402 #406 

## Review Checklist


josephjclark (changes requested):
I'm afraid I'm seeing a number of high-level issues here, I'm not quite sure what's happened.

The `test/intergration.js` file needs renaming to `test/integration.js` (there's only one R in integration). This is on the base branch and affects many of our branches, but let's get it sorted out here and merge it down.

I didn't think you wanted to use the koa server, but at the moment the tests seem to rely on it? Also at the moment we're declaring one set of servers in `intergration.js`, but also importing a different set of servers in `test/helpers/server.js`. So I'm a little confused about the intention. This is the sort of thing where a little comment in the PR description can help dispel confusion and make reviewing a little easier.

There is too much logging in the test code. Logging is useful while writing tests but it's really annoying in CI or when doing other work because the test output is spammed with meaningless stuff. So I'd really like to tighten up the use of logging here.

Also, all integration tests are failing locally. CI is failing too.

mtuchi:
@josephjclark I don't know why the CI is failing because locally all test are passing

mtuchi:
It turns out in ubuntu any `port < 1024` needs root permission, So i have updated the `httpsPort` to `1443` cc @josephjclark

josephjclark:
@mtuchi Great catch on the port number, I had no idea!

What about koa and http - do we need both server implementations?

mtuchi:
> What about koa and http - do we need both server implementations?

Hiya @josephjclark i have removed the koa implementation, and use the nodejs https module

josephjclark (approved):
Merci!

PR #417, opened 2023-10-23
`dhis2` remove `Class Log`

mtuchi:
## Summary

remove Class Log and replaced
- `Log.success` with `console.log`
- `Log.warn` with `console.warn`
- `Log.error` with `console.error`

## Review Checklist

### DIFF
--- a/packages/dhis2/src/Adaptor.js
+++ b/packages/dhis2/src/Adaptor.js
@@ -55,7 +55,6 @@ function configMigrationHelper(state) {
   const { hostUrl, apiUrl } = state.configuration;
   if (!hostUrl) {
     console.warn(
-      `⚠ Warning at ${new Date()}:\n∟`,
       'DEPRECATION WARNING: Please migrate instance address from `apiUrl` to `hostUrl`.'
     );
     state.configuration.hostUrl = apiUrl;
@@ -82,7 +81,7 @@ axios.interceptors.response.use(
           responseData: response.data,
         };
 
-        console.error(`✗ Error at ${new Date()}:\n∟`, newError.message);
+        console.error(newError.message);
 
         return Promise.reject(newError);
       }
@@ -96,10 +95,7 @@ axios.interceptors.response.use(
         // eslint-disable-next-line no-param-reassign
         response = { ...response, data: JSON.parse(response.data) };
       } catch (error) {
-        console.warn(
-          `⚠ Warning at ${new Date()}:\n∟`,
-          'Non-JSON response detected, unable to parse.'
-        );
+        console.warn('Non-JSON response detected, unable to parse.');
       }
     }
     return response;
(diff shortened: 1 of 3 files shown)


packages/dhis2/src/Adaptor.js
```diff
@@ -82,7 +82,7 @@ axios.interceptors.response.use(
           responseData: response.data,
         };
 
-        Log.error(newError.message);
+        console.error(`✗ Error at ${new Date()}:\n∟`, newError.message);
```
taylordowns2000, line 85: i think that printing the date is now redundant.

packages/dhis2/src/Adaptor.js
```diff
@@ -96,7 +96,10 @@ axios.interceptors.response.use(
         // eslint-disable-next-line no-param-reassign
         response = { ...response, data: JSON.parse(response.data) };
       } catch (error) {
-        Log.warn('Non-JSON response detected, unable to parse.');
+        console.warn(
+          `⚠ Warning at ${new Date()}:\n∟`,
```
taylordowns2000, line 100: same here

packages/dhis2/src/Adaptor.js
```diff
@@ -107,7 +110,10 @@ axios.interceptors.response.use(
 
     const details = error.response?.data;
 
-    Log.error(error.message || "That didn't work.");
+    console.error(
+      `✗ Error at ${new Date()}:\n∟`,
```
taylordowns2000, line 114: isn't the idea that the runtime prints the data already? (or decides if to print the date?)

taylordowns2000 (changes requested):
need a quick guide from @josephjclark here. i thought that printing the date was not necessary since that would be optionally (and later?) handled by the runtime. is that not the case?

josephjclark:
@taylordowns2000 Yeah, datestamps will be added to all logs by the runtime. 

That's a good catch. I'd have just removed the class wrapper but you're right, no need to include the timestamps. They're actually a bit distracting the CLI.

@mtuchi please go ahead and remove the timestamps

packages/dhis2/src/Adaptor.js
josephjclark: Just to be clear the _date_ is redundant, we should just log the message please

josephjclark: Just to be clear the _date_ is redundant, we should just log the message please

packages/dhis2/src/Adaptor.js
mtuchi: Fixed https://github.com/OpenFn/adaptors/pull/417/commits/6b89e24bb01f87c94ec229072927becfb976d56f

josephjclark (approved):

josephjclark (approved):

Issue #315, opened 2023-07-26
Remove common.http


josephjclark:
The common.http operations, which use axios, are going to be replaced with helper functions based on `unidici`, see #314 .

This means that common.http is no longer needed and can be retired to beach on the Amalfi coast.

These functions, and their documentation, should be removed, and the common adaptor should have a major version bump.

Dependent adaptors should be updated. They may in turn have a major version bump if they happen to export http directly.

# Deprecation

Instead of just removing common.http, we could deprecate it for one major version.

We should mark it deprecated in the docs, and each time a http function is called, we should console.warn a deprecation notice.

I would prefer a clean break, and given the way we're set up, I think this makes more sense. Adaptors can continue to use old, compatible versions of common until they're ready to migrate, in which case there's no real need to use the deprecated API.

It's only if job authors want a softer migration that we should do this. Ie, if Mtuchi is often working across two major versions of common and getting confused about which APIs are available, we should be lenient for a while.

# Workarounds

I understand that many adaptors use `http.get/post` as general catch-all functions. This is useful if an adaptor supports N api calls from the backend, but the n+1th is not supported - you can just  call GET with the full URL in your job code.

Removing common.http will remove this escape hatch and make life harder for job implementers. The new cmmon.util helpers are not designed for job code, they are not like-for-like replacements.

I have a couple of answers to this!

1) If the adaptor wants to expose a general `get`, it _should_ provide its own `get` function which handles auth and stuff anyway. `msgtraph` at least does this. I.e., it should already be wrapping `http.get`, not exposing it directly. In which case the exported function can be updated. 

2) If the adaptor is directly exporting `http` or `http.get` to the job, and they really want to continue doing that, the adaptor could import the new `language-http` and export the relevant functions from there. There is nothing stopping them from doing that.

3) Maybe the code should run in a `language-http` job anyway, utilising the workflow better

4) If we really really really want to expose general http operations to jobs, then we can re-write common.http to wrap the new helpers in operations.

josephjclark:
I would like to be aggressive here and just remove http, leaving is with a nice clean 2.0 adaptor.

Existing adaptors which have a version-locked dependency (ie, `"@openfn/langauge-common: 1.7.2"`) or a compatible version (ie, `"@openfn/langauge-common: ^1.7.2"`) will not be affected by the 2.0 release, and so the jobs attached to them are also unaffacted.

[...]

PR #420, opened 2023-10-23
Remove `common.http`

mtuchi:
## Summary

Version lock `@openfn/language-common` to all adaptors that were using `workspace:^`. And removed `http` in common.

Ref #315 

## Review Checklist


josephjclark:
I'm not sure this is right :thinking:  An adaptor with a common dependency like `^1.8.1` will not auto-update to common 2.0. It'll use the latest compatible version, which means the highest minor.

The only dependencies we need to worry about are `workspace:*` or `*` or something like that.

All we're doing here is restricting adaptors to not use patch or minor bumps of common - which could be important.

packages/bigquery/package.json
```diff
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "@google-cloud/bigquery": "^5.12.0",
-    "@openfn/language-common": "workspace:1.10.3",
+    "@openfn/language-common": "1.11.1",
```
josephjclark, line 24: I this case bigquery is locked to a specific version number. I don't know why (maybe we can do some github archaeology and work it out). 

This PR is updating the dependency to the next minor version. That's _probably_ fine but I wouldn't be comfortable doing it without checking first.

.changeset/lucky-pianos-sing.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Added version locking to adaptors that uses `openfn/language-common`
+  dependency to ensure consistent and predictable behavior across different
+  environments.
+- Removed `http` on `common` to improve compatibility and reduce potential
+  security vulnerabilities.
+
+### Deprecated
+
+- Deprecated the use of _http.method()_ Eg: `http.post()` as it is no longer a
```
josephjclark, line 15: `http.method()` isn't deprecated - it's removed entirely!

.changeset/lucky-pianos-sing.md
```diff
@@ -0,0 +1,20 @@
...
+  part of the `@openfn/language-common`.
+
+### Known Issues
```
josephjclark, line 18: We don't need to include a known issues section if there are no known issues :) To be honest I would prefer not to release software if there are known issues in it...

packages/common/test/http.test.js
```diff
@@ -1,281 +0,0 @@
-import nock from 'nock';
```
josephjclark, line 1: Are we still using `nock`? We should have moved entirely to the `undici` pattern by now

packages/common/src/http.js
```diff
@@ -1,202 +0,0 @@
-import { expandReferences, splitKeys } from './Adaptor';
-import axios from 'axios';
```
josephjclark, line 2: Have we removed the axios dependency?

packages/bigquery/package.json
mtuchi: @josephjclark i am getting this error when i run `pnpm test`  if i use the workspace version 
```
> @openfn/language-bigquery@2.0.2 test /Users/openfn/Workspace/adaptors/packages/bigquery
> mocha --experimental-specifier-resolution=node --no-warnings


file:///Users/openfn/Workspace/adaptors/packages/bigquery/src/Adaptor.js:130
  http,
  ^^^^
SyntaxError: The requested module '@openfn/language-common' does not provide an export named 'http'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
```

mtuchi: That's why i locked the `common` version to `1.11.1` which is the latest version. Should i perhaps use `^1.10.3`

packages/common/test/http.test.js
mtuchi: Removed

packages/common/src/http.js
mtuchi: Removed

packages/bigquery/package.json
josephjclark: That's actually quite an interesting problem. The workspace dependency, the actual adjacent source code, is no longer compatible with some modules.

We haven't yet bumped common and http to 2.0 have we? If we bump  them, we can make sure the dependencies are of the form `^1.10.3`. I don't think we need `workspace:` in front.

I wonder what `workspace:^1...` even means. Will it build a dependency from a tagged source branch? That' some serious muscle under the hood. I suspect we don't want `workspace:` and, strange as it seems, modules with legacy dependencies should download common from npm

mtuchi:
Closing this in favor of #427

Issue #421, opened 2023-10-24
`sftp` disconnect on error


mtuchi:
### Summary 
Currently when there is an error during operation execution, the sftp adaptor does not end the job execution instead it hang on the already established connection, 

### Request in few words
Catch the error on `commonExecute`, log the error then disconnect from sftp

mtuchi:
Fixed #422

PR #422, opened 2023-10-24
`sftp` disconnect on error

mtuchi:
## Summary

handle error on `commonExecute` and improve operation time logs

Ref #421 

## Review Checklist


mtuchi:
I can confirm that i have tested this locally, using a live sftp server 
Merging this to main cc @josephjclark

Issue #321, opened 2023-07-28
Update template to use new common.http


josephjclark:
`template` needs updating to use the new http helpers. This probably means removing the utils file.

But we may want to include an example in template of how to expose a bespoke http helper (ie, re-export http.get, or create a wrapper around a helper). This needs a bit of thinking about once the helpers are done.

PR #425, opened 2023-10-27
rewrite `template` to use common.http

mtuchi:
## Summary
Re-write `template` to use the new common http

Ref #321

## Review Checklist


packages/template/src/Adaptor.js
```diff
@@ -2,9 +2,57 @@ import {
...
+    const options = {
+      ...resolvedParams,
+      headers,
+      baseUrl,
+    };
+
+    return commonRequest(method, resolvedPath, options)
+      .then(response => {
+        const { method, url, body, code, duration } = response;
+        console.log(method, url, '-', code, 'in', duration + 'ms');
```
josephjclark, line 46: This is REALLY interesting.

This is painfully boring and rote code and I do not want adaptor implementors to have to write it.

But we're making them, right, because the underlying common helper doesn't auto log. We're doing it manually in http and we're doing it manually here.

This is just an idea, but what if the response had some kind of `log()` function built into it, which logs the message printed by HTTP. Now any adaptor can easily and painlessly print out a generic HTTP response log.

Actually, we should also consider that most adaptors don't just want to log the http details. Most adaptors would say "Created patient bob (222ms)" and ignore the http stuff.

packages/template/src/Adaptor.js
```diff
@@ -2,9 +2,57 @@ import {
...
-import { request } from './Utils';
 
+import {
+  request as commonRequest,
+  expandReferences,
+} from '@openfn/language-common/util';
+
+import { makeBasicAuth, assertConfiguration } from './Utils';
+
+export function request(method, path, params, callback) {
```
josephjclark, line 13: I'm not sure I like this pattern. I guess we need to be thinking about what we want to do with the template.

I don't like this code because it's full of boiler-plate. Repetitive code that doesn't add value.

I understand why we expand references - that's important and good.

I like the assertConfiguration pattern (although I wonder if `validateConfig` or `checkConfig` would be better names (yeah, I know I've asked you to use `assert` before, but I'm not sure it's so helpful in this context)).

I'm not keen on any of header handling - all it should really need to do is auth. Doesn't the common util handle JSON for us?

I'm also not keen on the abstraction. This adaptor has  `create` and a `request` API, where `create` is just a wrapper which does nothing. It isn't at all clear why both functions need to exist - if this is a complete adaptor, you'd only need one.

I would rather have `create` and `get` functions, designed to get and create resources. Maybe they use a common `request` but I think it would establish a more helpful template pattern if each operation handled its own references, called a helper function to add auth, and then called common request (which then means you don't need to alias common request).

As a philosophical point, I don't mind at all if each operation function handles its own references, even if they're copied and pasted. I might even prefer that pattern because it's clear and explicit: each operation handles its own inputs, then passes them off to internal helpers.

packages/template/test/Adaptor.test.js
```diff
@@ -38,11 +37,67 @@ describe('execute', () => {
   });
 });
 
+describe('request', () => {
+  it('should makea get request', async () => {
```
josephjclark, line 41: typo

packages/template/test/Adaptor.test.js
```diff
@@ -1,10 +1,9 @@
 import { expect } from 'chai';
```
josephjclark, line 1: This file could use some more comments to explain how the tests are designed - especially around the mock stuff, which is neither trivial nor intuitive

josephjclark (changes requested):
Some comments left - I think we need to think and talk about this one some more. I'm still not entirely clear on what the template code should look like.

Also the ground is still moving a bit in the helper functions so this may be the last thing we finalise.

packages/template/src/Adaptor.js
mtuchi: I like this idea, Can i suggest we have `response.data()` and `response.log()` then, It looks more organized 

```
// The new response structure 
return {
  data: () => ({
    method,
    url: responseUrl(fullUrlOrPath, response),
    code: response.statusCode,
    headers: response.headers,
    body: responseBody,
    message: statusText,
    duration,
  }),
  log: (message) => {
    if (message) {
      console.log(message);
    } else {
      console.log(
        method,
        fullUrlOrPath,
        "-",
        response.statusCode,
        "in",
        duration + "ms"
      );
    }
  },
};

```

mtuchi:
```
// response.log()
log: () => console.log(
        method,
        fullUrlOrPath,
        "-",
        response.statusCode,
        "in",
        duration + "ms"
      );```

PR #426, opened 2023-10-30
[Epic] New http

josephjclark:
Overview PR for all the http and common work.


packages/common/src/util/http.js
```diff
@@ -0,0 +1,135 @@
+import { Client, MockAgent } from 'undici';
+
+const clients = new Map();
```
josephjclark, line 3: I may have messed up this pattern :grimacing: 

This assumes that the adaptor is loaded into an isolate context, and that each workflow has its own context.

A fair assumption for CLI and for platform.

But moving forward we are going to be using long-running adaptors. Ignoring the short term details, really the adaptor should be cached in memory and re-used across different workflows. If we assume we can do this securely (TBD), we also need to be able to do it stably.

The common client map is a problem here. It should be initialised once per workflow, and destroyed when the workflow ends.

In an adaptor, we can hook to execute to re-initialise the clientmap. Although if two jobs are running concurrently they will interfere. Alternatively in an adaptor we can use the state object - a pattern I've been against, but would be totally safe. But the request helper doesn't see state.

The best solution I can see right now is for the util to have a `destroyClient(baseUrl, options)` function, which the adaptor MUST call  at the end of the job. That would ensure clients are not re-used (well maybe - actually, if two jobs are running concurrently with the same URL and options, they'll interfere)

Perhaps this is all nonsense and it's up to the runtime engine to ensure that each workflow has a dedicated module instance. Disabling any module re-use is surely the best way to ensure stability, flexibility AND security across workflows, and would allow the adaptor to do whatever the heck it wants (which is ideally the policy we would adopt, rather than being very prescriptive).

Ok, so this is a long comment to say that it is appropriate for the new runtime engine to support this pattern.

packages/common/src/util/http.js
```diff
@@ -0,0 +1,135 @@
...
+  timeout: 300e3, // Default to 300 seconds
+  headers: {},
+  query: undefined,
+  body: undefined,
+  errors: {},
+  tls: {},
+  parseAs: 'auto',
+};
+
+/**
```
josephjclark, line 65: Request is an internal function. We should be describing (and more importantly the types of its arguments) it in typescript, probably in index.ts.

This function will not currently be generated in docs.openfn.org. And it probably shouldn't be because it's not used by job writers (only by adpator writers). So we don't actually need a full JSDoc here.

packages/common/test/util/references.test.js
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { expandReferences, normalizeOauthConfig } from '../src/util';
+import { expandReferences, normalizeOauthConfig } from '../../src/util/index.js';
```
josephjclark, line 2: TODO why does this have `.js`? Does it need it? If so shouldn't everything need `.js`?

josephjclark:
What is going on with this diff :face_exhaling:

josephjclark:
Closing this down as the diff is messed up.

The `316-http-rewrite-2` is the latest stable version anyway so I'm going to repurpose that into a new epic branch.

PR #427, opened 2023-11-02
Remove `common.http`

mtuchi:
## Summary

Version lock `@openfn/language-common` to all adaptors that were using `workspace:^`. And removed `http` in common.

Ref #315 

## Review Checklist


packages/common/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
-v0.4.0
+## 2.0.0
+
+### Major Changes
+
+- Added version locking to adaptors that uses `openfn/language-common`
```
josephjclark, line 5: This isn't a change to common though. Any adaptor who had its dependency changed should have this.

Remember that the changeset is stuff that users need to know. It's not marketing, so we don't need to justify the work. We just need to inform users what's different in this release.

packages/common/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
...
+  dependency to ensure consistent and predictable behavior across different
+  environments.
+- Removed `http` on `common` to improve compatibility and reduce potential
```
josephjclark, line 8: Again we don't need to justify the change here - although as it's a breaking change it's useful to mention the workaround,

I think this diff should be something like `removed the http namespace. Adaptors should use the new request function`.

Oh, and the new request helper should be mentioned in this change!

packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,79 @@
 # @openfn/language-http
 
+## 6.0.0
+
+### Overview
+
+This release marks a major milestone in the development of the HTTP Adaptor. The
```
josephjclark, line 7: I will adjust this changelog, but I'll do it in a different PR

packages/kobotoolbox/package.json
```diff
@@ -25,7 +25,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "workspace:^1.8.1"
+    "@openfn/language-common": "^1.8.1"
```
josephjclark, line 28: This diff isn't necessary - we're going to trigger a patch release for no change to the final package. There's quite a few like this. I think `workspace:^...` dependencies are perfectly safe.

packages/msgraph/package.json
```diff
@@ -27,7 +27,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "workspace:*",
+    "@openfn/language-common": "^1.10.3",
```
josephjclark, line 30: This one IS necessary, because `workspace:*` will drift to common 2.0 (and so break).

We also need a patch bump on the adaptor version.

packages/salesforce/package.json
```diff
@@ -31,7 +31,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "workspace:*",
+    "@openfn/language-common": "^1.10.3",
```
josephjclark, line 34: This one is necessary too (and needs a corresponding version bump)

josephjclark (changes requested):
Couple of comments - I think the diff is still too big, and some affected adaptors will need version bumps.

josephjclark:
Having started to reproduce this on another branch - I finally understand why the diff is so big.

`pnpm changeset version` doesn't respect semver. If we bump common to 2.0, all adaptors will automaticallly update their versions and dependencies to use it.

The only releases from this work should be common and http. Other adaptors can update in time.

I think the only way I can reduce the impact of all this is to deprecate, rather than remove, the old helpers. This is a bit confusing for developers so I'm gonna put a big screaming warning on those files.

PR #430, opened 2023-11-09
OpenSPP release

josephjclark:
Release branch for the new openSPP adaptor.

Tags are pushed, versions are ready, there is no changeset needed.

I suggest we release from here then merge down into main.


nhatnm0612:
wonderful, thanks a lot 😄

Issue #432, opened 2023-11-13
http @5.x: response data is huge


josephjclark:
http 5.x. which uses axios,  writes the a response object to result state. It

This includes 20k of minified JSON.

A lot of the damage seems to be from the `agent` key.

I think this has been happening forever and it's because we just "blindly" write the axios response straight to state.

Discuss.

josephjclark:
CLI 0.1 (5 months old) also exhibits the same behaviour

mtuchi:
After a close investigation with multiple version of http adaptor here is the summary 

```
get("https://jsonplaceholder.typicode.com/todos/1")
```

| platform | version  |
| -------- | -------- |
| core     | v1.5.1   |
| node     | v18.17.1 |
| cli      | v0.4.6   |

| http   | output size | state.response keys                                                                       | state.data keys                          |
| ------ | ----------- | ----------------------------------------------------------------------------------------- | ---------------------------------------- |
| v3.0.0 | 4.0K        | no _state.response_                                                                       | [ 'body', '__cookie', '__headers' ]      |
| v4.0.0 | 124K        | [ 'status', 'statusText','headers', 'config', 'request', 'data', 'httpStatus', 'message'] | [ 'userId', 'id', 'title', 'completed' ] |
| v5.0.0 | 112k        | [ 'status', 'statusText','headers', 'config', 'request', 'data', 'httpStatus', 'message'] | [ 'userId', 'id', 'title', 'completed' ] |

mtuchi:
In `state.response.request` we have a-lot of data that is not useful in most cases, 
<details><summary>Here is the list of the state.response.request</summary>
<p>

```
[
  '_events',
  '_eventsCount',
  '_maxListeners',
  'outputData',
  'outputSize',
  'writable',
  'destroyed',
  '_last',
  'chunkedEncoding',
  'shouldKeepAlive',
  'maxRequestsOnConnectionReached',
  '_defaultKeepAlive',
  'useChunkedEncodingByDefault',
  'sendDate',
  '_removedConnection',
  '_removedContLen',
  '_removedTE',
  'strictContentLength',
  '_contentLength',
  '_hasBody',
  '_trailer',
  'finished',
  '_headerSent',
  '_closed',
  'socket',
  '_header',
  '_keepAliveTimeout',
  '_onPendingData',
  'agent',
  'socketPath',
  'method',
  'maxHeaderSize',
  'insecureHTTPParser',
  'joinDuplicateHeaders',
[...]

PR #433, opened 2023-11-15
`http` lighten `response.request`

mtuchi:
## Summary

Returned only `{ method, path, host, protocol }` for `response.request`. which lighten the final state  

## Issues

Reference #432 

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/brave-carrots-suffer.md
+++ b/.changeset/brave-carrots-suffer.md
@@ -1,6 +1,0 @@
----
-'@openfn/language-http': minor
----
-
-Clean up `state.response.request` by returning only
-`{ method, path, host, protocol, _headers }`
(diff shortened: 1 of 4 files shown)


josephjclark:
This is looking good. Testing this gives me 4kb of pretty-printed JSON, which is a lot better. Looking at the data there's not much more I'd like to remove.

One thing I'd note - we don't have access to the _request_ headers any more. `response.headers` is just the response headers. I think we should still keep `response.request._headers` from the original object.

josephjclark:
Also @mtuchi I think this should be a minor change

mtuchi:
@josephjclark i made the change necessary changes [db6d0b8](https://github.com/OpenFn/adaptors/pull/433/commits/db6d0b81432d9ac34ea2784d715c8574f493681c)

josephjclark (approved):

PR #435, opened 2023-11-16
[UPD] update OpenSPP adaptor with new functions

nhatnm0612:
## Summary

Update OpenSPP adaptor with new functions

## Details

Including these new functions as requested:
- createIndividual: create new individual
- updateIndividual: update existing individual
- createGroup: create new group
- updateGroup: update existing group
- addToGroup: add individual to group with(out) role
- removeFromGroup: end membership of individual to group

## Issues

[Requests From My Company](https://github.com/OpenSPP/openspp-registry/issues/248#issuecomment-1811798050)

## Review Checklist

      migration guide followed?


packages/openspp/src/Adaptor.js
```diff
@@ -660,6 +658,311 @@ export function unenroll(registrant_id, program_id) {
...
+ * create new individual for OpenSPP
+ * @public
+ * @example
+ * createIndividual({ name: "Individual 1" })
+ * @function
+ * @param {object} data - registrant create data
+ * @param {function} callback - An optional callback function
+ * @returns {Operation}
+ */
+export function createIndividual(data, callback=false) {
```
josephjclark, line 671: If you default callback to a function which returns state:
```
export function createIndividual(data, callback=(s) => s) { /* ... */}
```

then later, instead of:

```
if (callback) {
    return callback(nextState);
}
return nextState;
```
You can simply do:
```
return callback(nextState)
```

packages/openspp/src/Adaptor.js
```diff
@@ -660,6 +658,311 @@ export function unenroll(registrant_id, program_id) {
...
+  return async state => {
+    try {
+      if (typeof data !== 'object' || Array.isArray(data) || data === null) {
```
josephjclark, line 674: Do you really need or want to do this kind of validation in the adaptor?

If someone passes invalid data, won't the OpenSPP backend throw a userful error?

You can leave this stuff if you want, but my advice is to leave validation to the backend as much as possible. Adaptor-side validation should probably only focus on very common use-cases.

Otherwise you can end up writing validation code forever and it's all sort of redundant and might not even be useful.

packages/openspp/src/Adaptor.js
```diff
@@ -660,6 +658,311 @@ export function unenroll(registrant_id, program_id) {
...
+      );
+      let individualRegistrantId = res[0].registrant_id;
+      console.log(`ℹ Individual created with registrant ID: ${individualRegistrantId}`);
+      let nextState = composeNextState(state, individualRegistrantId);
+      if (callback) {
+        return callback(nextState);
+      }
+      return nextState;
+    } catch (err) {
+      console.log(`✗ Error: ${err}`);
```
josephjclark, line 697: I think I gave you bad advice before - I'm not sure what I was thinking now! 

You may not always want to return state if there's an error. It depends on the error really. Sometimes it's fine to throw the error and let the job "crash". You should only trap the error if it's non-fatal or the next operation should still run.

In this case, if there's a validation error on the data, the job will carry on running. That might not be what you want - it might be better for the job to die, and alert the team that action needs to be taken.

packages/openspp/src/Adaptor.js
```diff
@@ -660,6 +658,311 @@ export function unenroll(registrant_id, program_id) {
...
+          [['name', '=', role]]
+        );
+        if (roleId.length === 0) {
+          roleId = [await sppConnector.create(
+            'g2p.group.membership.kind',
+            { name: role }
+          )];
+        }
+      }
+      let res = await sppConnector.searchRead(
```
josephjclark, line 860: If you're not going to re-assign a variable (like `res`), you should declare it with `const`. That tells the reader that the type  (and maybe value) of the  variable won't change - which is really useful to know.

The majority of variables in modern Javascript are declared as `const` (I actually use `let` quite rarely!

packages/openspp/src/Adaptor.js
```diff
@@ -660,6 +658,311 @@ export function unenroll(registrant_id, program_id) {
...
+          ['group.registrant_id', '=', group_id],
+          ['individual.registrant_id', '=', individual_id],
+          ['is_ended', '=', false],
+        ],
+        ['id'],
+      );
+      if (res.length > 0) {
+        let groupMembershipIds = res.map( i => i.id );
+        let now = new Date();
+        let sppDateTimeNowString =
```
josephjclark, line 948: I am reluctant to suggest this because I think it's broken in the CLI and it's not really a pattern I'm keen to support... but the `common` adaptor exports a `dateFns` objects. dateFns has a `format` function which is really useful for this stuff: https://date-fns.org/v2.16.1/docs/format

If it works it'll let you do something like this:
```
 let sppDateTimeNowString = format (new Date(), 'y-M-d H:m:s')`
```

josephjclark (commented):
Hi again! I've left a couple of suggestions for you to think about. I would address the const and callback thing for sure, maybe think a little bit more about error handling and validation - that stuff is up to you really.

packages/openspp/src/Adaptor.js
nhatnm0612: @josephjclark done fixing!

packages/openspp/src/Adaptor.js
nhatnm0612: done removing :D

packages/openspp/src/Adaptor.js
nhatnm0612: replaced

packages/openspp/src/Adaptor.js
nhatnm0612: done replacing

packages/openspp/CHANGELOG.md
```diff
@@ -3,3 +3,14 @@
 ## 1.0.0
 
 - Create OpenSPP adaptor
+
+## 1.0.1
```
josephjclark, line 7: First off can you put the most recent changes first in the log?

Secondly, for future reference, next time you want to update the version, please run `pnpm changeset` from the repo root. That'll ask you some questions and generate a changelog for you. There should be more details in the main readme.

josephjclark:
@nhatnm0612 I think this is looking great - can I just check you're happy with the error handling for now? You can always raise a PR to change it later, of course :)

If you're happy we'll take it on and try and get a release out today.

josephjclark (approved):

nhatnm0612:
> @nhatnm0612 I think this is looking great - can I just check you're happy with the error handling for now? You can always raise a PR to change it later, of course :)
> 
> If you're happy we'll take it on and try and get a release out today.

I am please with this error handling for now, we users now are technical people anyway. I will change these error handling in the future if needed @josephjclark 

it is a pleasure working with you, there were lots of things that I learnt from you guys 😄

PR #444, opened 2023-12-06
Add `xml2json` func, expandReference `parseXML(body)`

mtuchi:
## Short Description

Add `xml2json` function and improve `parseXML` function

## Why ?
The implementation team has ongoing project in partnership with vera team. This project will have an automated workflow that will take data from SunSystem which has a SOAP API to Salesforce. To make the transformation process easy, We need an adaptor function that can be used to convert the xml response to a json object. 

## Implementation Notes #446 

Add `xml2json(xmlString, options, callback)` function that uses [xml2js package](https://github.com/Leonidas-from-XIV/node-xml2js/tree/master). This function will take `xml` string and convert it to `json` object.  

This PR also contain improvements for `parseXML` function. The first argument of the `parseXML` function was not using `expandReferences` which makes it hard to access data from `state` in job code.

## Review Checklist

### DIFF
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -41,7 +41,8 @@
     "form-data": "^3.0.0",
     "lodash": "^4.17.19",
     "request": "^2.88.2",
-    "tough-cookie": "^4.0.0"
+    "tough-cookie": "^4.0.0",
+    "xml2js": "0.5.0"
   },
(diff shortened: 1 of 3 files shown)


mtuchi:
Adding `xml2json` in http is not advised for the time being since `parseXML` is already useful enough.

PR #445, opened 2023-12-06
odoo: Implement new adaptor

mtuchi:
## Summary

Implement a new `odoo` adaptor with `create, update, read, and delete` functions

## Details

## Review Checklist

      migration guide followed?


packages/odoo/CHANGELOG.md
```diff
@@ -0,0 +1,104 @@
+# @openfn/language-template
```
josephjclark, line 1: This changelog needs to be reset for a clean 1.0 implementation with no history

packages/odoo/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@openfn/language-odoo",
+  "version": "3.0.1",
```
josephjclark, line 3: Version number need to be 1.0.0

packages/odoo/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "description": "The standard odoo for building OpenFn adaptors",
```
josephjclark, line 4: This description looks off :thinking:

packages/odoo/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "types/",
+    "ast.json",
+    "configuration-schema.json"
+  ],
+  "dependencies": {
+    "@openfn/language-common": "workspace:*",
+    "odoo-await": "^3.4.1"
+  },
+  "devDependencies": {
+    "@openfn/buildtools": "workspace:^1.0.2",
```
josephjclark, line 34: I think these dev dependencies can be updated. I don't think we need buildtools, simple-ast. Maybe check against intuit's dependencies?

packages/odoo/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
+import {
+  execute as commonExecute,
+  composeNextState,
+} from '@openfn/language-common';
+import { expandReferences } from '@openfn/language-common/util';
+
+import pkg from 'odoo-await';
```
josephjclark, line 7: Can we not do `import Odoo from 'odoo-await'` ?

packages/odoo/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+}
+/**
+ * Create a record in Odoo
+ * @public
+ * @example
+ * create("res.partner", { name: "Kool Keith" });
+ * @function
+ * @param {string} model - The specific record model i.e. "res.partner"
+ * @param {object} data - The data to be created in JSON.
+ * @param {string} externalId - Optional external ID for the record.
```
josephjclark, line 67: Maybe we can use an options object here, like:
```
create("res.partner", { name: "Kool Keith" }, { externalId: 22 );
```

The benefits are:
1) we can add more options later without a breaking change
2) the code is more readable when we explicitly pass in the `externalId` key. Otherwise it's just a random value in the signature and it looks a bit wierd

packages/odoo/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+export function create(model, data, externalId) {
+  return async state => {
+    const [resolvedModel, resolvedData, resolvedExternalId] = expandReferences(
+      state,
+      model,
+      data,
+      externalId
+    );
+
+    console.log(resolvedModel, resolvedData, resolvedExternalId);
```
josephjclark, line 79: This log line needs removing (there are others too)

packages/odoo/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Get a record from Odoo
+ * @public
+ * @example
+ * read("res.partner", [1] , [name]);
+ * @function
+ * @param {string} model - The specific record model from i.e. "res.partner"
+ * @param {number[]} recordId - An array of record IDs to read.
+ * @param {string[]} fields - An array of fields to read from the record.
```
josephjclark, line 98: Is this optional? Can we not just read all fields?

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,89 @@
...
+import {
+  execute,
+  create,
+  dataValue,
+  update,
+  read,
+  deleteRecord,
+} from '../src/Adaptor.js';
+
+import MockAgent from './mockAgent.js';
```
josephjclark, line 12: We should be able to remove all this now, as we're not using the HTTP mock?

We should remove mockAgent.js entirely

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,89 @@
...
+  database: 'somedatabase',
+};
+
+const state = { configuration };
+
+describe('read record', () => {
+  it('should read a record', () => {
+    const mock = {
+      read: (model, id, fields) => {
+        expect(model).to.eql('product.product');
```
josephjclark, line 30: Why not also assert on id and fields?

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,89 @@
...
+        return {
```
josephjclark, line 31: If we're passing an array of ids to oodoo, should it not return an array of items back?

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,89 @@
...
+          id: 2,
+          name: 'Saas Product',
+        };
+      },
+    };
+    enableMockClient(mock);
+
+    read('product.product', [2], ['name'])(state);
```
josephjclark, line 39: There is no assertion on what the returned state looks like. How do we know if the adaptor has written to state.data correctly?

josephjclark (commented):
Left a few thoughts!

packages/odoo/test/Adaptor.test.js
hunterachieng: Yes it is an array

packages/odoo/test/Adaptor.test.js
hunterachieng: state only has the config and if I try to log the response, I do not get anything

josephjclark: > state only has the config

That sounds like a bug you need to fix :)

packages/odoo/src/Adaptor.js
josephjclark: Yes, exactly. We should include docs to clarify this (don't do it on the param,  add them to the main docs above @public, like `Returns all fields unless a field list is provided as a third argument`.

Then we need two examples: one with a caption `Download a single item with all fields` and `Download items with select fields`. In the second example, use `$.recordIds` instead of a passing a hardcoded value.

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,98 @@
...
+      'res.partner',
+      { name: 'Jane Doe' },
+      { externalId: 23 }
+    )(state);
+    expect(data).to.eql(15);
+  });
+});
+
+describe('update', () => {
+  it('should update a record', async () => {
```
josephjclark, line 66: What happens if we try to update a record that doesn't exist? Do we throw an exception? Should we?

packages/odoo/test/Adaptor.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+    const { data } = await update('product.product', 4, {
+      name: 'Testing Product',
+    })(state);
+    expect(data).to.eql(true);
+  });
+});
+
+describe('delete', () => {
+  it('should delete a record', async () => {
```
josephjclark, line 85: What happens if we try to delete  a record that doesn't exist? Do we throw an exception? Should we?

hunterachieng: If we try to delete a record that does not exist, it does not throw an error. Instead, it just returns `true` whether the record is there or not

josephjclark: Maybe if the client returns `false`, we should throw?

It can be hard to know whether this is the right behaviour. It's probably fine if you ask to delete an item that's been deleted.

One way we often fix this is to include an `options` object and pass `throwOnError: boolean` to control the behaviour. That feels good here but in the case of `delete` I think we should default to `false`.

For `update`, on the other hand, if the update failed I think that's probably an error because the database will not be in the state we asked for it to be. So we'd default `throwOnErrow` to `true` there.

Ok, this is a 1.0 release. Let's be brave and opinionated. `delete` should not throw, but `update` should. Don't add any options at this stage.

josephjclark: If we can `create` should also throw if the requested thing wasn't created. I don't know if the client gives us feedback on that. Don't spend a lot of time on it but maybe take a look? :)

hunterachieng: @josephjclark  `update` throws an error automatically if the record is not found. I have added a test for the error

hunterachieng: @josephjclark  `update` throws an error automatically if the record is not found. I have added a test for the error

PR #452, opened 2023-12-15
Remove Regex Patterns In Configuration Schemas

### LINKED ISSUE #451: Remove Regex Pattern For API Key In Mailchimp
### Summary
We are now using regex  in configuration-schema for `Mailchimp` and `Salesforce` adaptor and those validation rules and when we tried to turn them on for the first time we found a bunch of incompatibilities that prevented existing credentials from working. I’d recommend removing them now, and (low priority) considering if and how to introduce some stuff like that later.

elias-ba:
## Summary

This PR ships code that removes the regex patterns that were used for validation in the mailchimp and salesforce adaptors configuration-schema.json file. We are removing those because:

- they were introducing severe bugs to Lightning migrations.
- the regex patterns seems to not match as expected

We need to discuss whether we need those patterns, in that case make sure we have the right ones

## Issues

Fixes #451 

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/angry-grapes-cover.md
+++ b/.changeset/angry-grapes-cover.md
@@ -1,6 +1,0 @@
----
-'@openfn/language-mailchimp': patch
-'@openfn/language-salesforce': patch
----
-
-Remove regex pattern for validation and bump minLength to 1
--- a/.changeset/angry-grapes-cover.md
+++ b/.changeset/angry-grapes-cover.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-mailchimp': patch
+'@openfn/language-salesforce': patch
+---
+
+Remove regex pattern for validation and bump minLength to 1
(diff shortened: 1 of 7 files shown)


packages/mailchimp/configuration-schema.json
```diff
@@ -5,23 +5,21 @@
             "title": "Data Center",
             "type": "string",
             "description": "Mailchimp Data Center for your account",
-            "minLength": 3,
+            "minLength": 1,
```
mtuchi, line 8: @elias-ba i am curious to know why `1`, because from my understanding mailchimp data centers have at least `3` characters going up, For example 
`uk8` 3 characters, `us11` 4 characters

.changeset/angry-grapes-cover.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-mailchimp': patch
+'@openfn/language-salesforce': patch
+---
+
+Remove regex pattern for validation and bump minLength to 1
```
mtuchi, line 6: I think instead of **bump**, it should say **changed**

taylordowns2000 (approved):

Issue #320, opened 2023-07-28
HTTP


josephjclark:
Epic for:

* Replacing axios 
* Creating new http helpers
* "rebasing" all adaptors on the new helpers

This has all branched out of #283

Issue #304, opened 2023-07-12
Magic jsdoc annotations should not be written to public documentation


josephjclark:
Whoops:

![image](https://github.com/OpenFn/adaptors/assets/7052509/ed2bf296-f365-4ea7-aaad-466e18e3c6c1)

Can't believe I didn't see this earlier :(

We should chop the `@magic` metadata tags out when generating the documentation, here in kit. I don't think that will be particularly hard to do (and I'm quite surprised that the tags even render)

josephjclark:
Oh I should consider adding this to the documentation improvements in #457

Issue #370, opened 2023-08-29
Streaming Support


josephjclark:
This is a cross-repo epic to introduce better support for streaming APIs into our adaptors and runtime. I don't know how to structure this so here's a brain dump.

## Streaming Design

In order to support streams for more efficient data processing, we need to make a number of changes.

From a very high level, I would like the runtime and adaptors to understand streams as first-class data type. In fact actually, I want arrays and streams to behave exactly the same way, and for the streaming layer to be almost seamless.

There is a whole host of considerations here and for now I'm just going to type them out.

### Streams on state

It would be super helpful to write streams to state.

Obviously a stream doesn't serialize unless we explicitly write it, so by default it breaks the design. Probably right now the new runtime will drop it in `fast-safe-stringify` (I don't actually know).

It should be perfectly safe to pass a stream between jobs in a workflow (so long as the workflow runs in the same process, which so far it is designed to).

If there are any streams on the final state (and they are not exhausted), we should serialise them to text.

Detecting a Stream is a little tricky - we should ensure  that any streams we use match `instanceof Readable` (the `fetch` body does not, for example).

This is all a bit easier if we support LazyArrays (see below).

This would have helped us recently in a case where  we had to stream a large CSV file to JSON, serialize the JSON to a larger file than the original CSV, and then re-load that JSON in the next job, blowing platform memory limits. Forwarding the stream to the next job would have really helped.

### Transformers / parsers

The big problem with a stream is: how do you convert the array buffer into something meaningful? Actually for us it's always: how do we convert the underlying data to JSON?

We probably need a common `pipe` operation. This would not be seamless, but would allow us to convert a raw response into a JSON or CSV stream. Adaptors may choose to provide seamless APIs where possible.

Parsing a http get might look something like this:
```
get(www, { parseAs: ‘stream’ ) // fetches data as a stream, writes to state.data
pipe(state.data, jsonArray) Note that it doesn’t start streaming yet - just adds to the pipeline
each(‘data’, fn(() => { // iterate over the stream
	// this receives json objects
});
```
We would have to provide a bunch of helper transformer functions for common types, and allow users to define their own. Maybe common exports a stream namespace.

### LazyArrays

I have a bit of an idea that a Stream is just a lazy array, where we only load the bit we're using into working memory.

I envision a wrapper around Stream, a LazyArray but probably not actually called that, which presents an array-like interface:
```
const a = new LazyArray(someStream)
const someMappedLazyArray = a.map(() => {... } );
[...]

PR #457, opened 2024-01-11
[Epic] New HTTP Adaptor & Utils

### LINKED ISSUE #316: http: rewrite to use new helper functions

josephjclark:
Epic PR for all the http and common work.

Closes #316 #320 #304

# Changes

* [x] New http helpers in common
* [x] New http adaptor
* [x] deprecated old common.http
* [x] update template to showcase new pattern
* [x] check and maybe improve docs
* [ ] ~~remove axios, superagent and request across the repo~~
* [x] Add linting for undeclared variables (and fix a whole bunch of stuff)

# Overview

This PR has the broad aim of removing axios, superagent and request in favour of a single HTTP client - `undici`.

But we're not quite doing that yet as it's too painful.

As a starting point,  we have added a new set of http helper functions (functions not operations!) into `common`, which you'll find in `common/util/http`. These are wrappers around `undici`. We intend that all new and existing adaptors migrate onto these utils.

The headline feature for this release is HTTP 6.0, which is the existing HTTP adaptor "rebased" onto the common helpers, with axios removed. This comes with loads of breaking changes, so there's a thorough migration guide in the http changelog. HTTP should now be a very lightweight adaptor which just wraps the common helpers.

The template has been changed to show usage of the new common helpers.

Finally, the docs output has been improved (internal links with upper case characters fixed; magic metadata removed; better documentation for http options added through a typedef) and linting has been strengthened (resulting in many small but critical patches on existing adaptors).


josephjclark:
The examples on docs.openfn.org are likely to be way out of date now :( Maybe we ought to go through them

packages/template/package.json
```diff
@@ -27,7 +27,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "^1.8.1"
+    "@openfn/language-common": "^1.12.0"
```
josephjclark, line 30: TODO!! Template needs its version bumping, I guess?

Or maybe we should forever lock it at 1.0.0?

josephjclark:
@taylordowns2000 what is the purpose of the version and changelog in `template`?

We've previously been bumping the version number when we make a change to template, as well as adding a changelog. Like we would for a normal adaptor.

But this is meaningless. If someone copies and pastes the template - as we suggest they do - they have to delete the changelog and reset the version number. It's a waste of time - and if this gets missed, then the new adaptor is published with a confusing history.

Although it's orthogonal to this branch, can I suggest we:
a) Set `template` to v `1.0.0`
b) Remove the `template` changelog and add a  stub entry for the 1.0.0 release ("intial release" would be fine)

josephjclark:
Ah, two things I haven't done here:

* Allow access to the raw request object (response.raw as a dynamic, non-enumerable property should work and not be serialised)
* Allow access to the raw error (again, error.raw should work in the same way as above)

We could release without these - but I probably ought to commit one more day to it.

josephjclark:
I'd also add that we haven't added a steaming API (although you can request a body as a stream). See #370

This is fine and part of the plan. It should not hold up the release.

taylordowns2000:
> Although it's orthogonal to this branch, can I suggest we:
> a) Set `template` to v `1.0.0`
> b) Remove the `template` changelog and add a stub entry for the 1.0.0 release ("intial release" would be fine)

Yes, I agree with this @josephjclark .

josephjclark:
@mtuchi reports an issue if the URL is `/` or ``. Something to do with how the URL is assembled.

josephjclark:
Added a cheeky fix for #304

josephjclark:
@mtuchi caught an issue where parseXML was broken because expandReferences isn't imported.

This wasn't picked up in unit testing or linting, which is a bit scary.

I don't want to to write a parseXML unit test now, so I have added a linting rule for undeclared variables. This triggers a rash of patch fixes to various adaptors.

packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+  ### New features
+
+  - Error mapping. Provide custom errors to response status codes, eg, Eg:
+    `{errors: {404: "You don't have access"}}`
+  - parseAs option. Overrides the default response type. Eg: `{parseAs: "json"}`
+  - Request logging. All requests are now logged to stdout.
+  - tls options. Pass SSL certificates using the `tls` option on any request.
+
+  ### Breaking changes
```
josephjclark, line 17: More migration notes with @mtuchi

* The `params` and `qs` parameters have been removed. Use`query` instead (it's the same)
* `keepCookie`, which basically just pulls cookies out of the header and writes them tostate.data. But not in a nice way IMO. We've removed support for this. We can reimplement later, with a spec, if anyone needs it.
* `successCodes` is now the error map.
* `transformResponse` is just a callback? We've removed it. Use the callback isntead.
* agentOptions - use the TLS option. `rejectUnauthorised` might still work? It's unclear
* `reponseType` is `parseAs`
* `authorisation` we do this for you in config. If you need to set it explicitly, she the header  (use the helper)

josephjclark: `data` is now `body`

packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
...
+### Migration Guide
+
+- Use `query` instead of `params` or `qs` parameters to a request.
+- Convert succcessCodes to the `errorMap`. Ie, `successCodes: [200, 404]`
+  becomes `errorMap: { 404: false }`
+- Instead of `transformResponse`, use the callback parameter.
+- Convert `agentOptions` to `tls`.The values should be the same.
+- Convert `responseType` to `parseAs`. Note that this is really only useful for
+  JSON, and only needed if the server does not include a content-type response
+  heade.r
```
mtuchi, line 36: typo, should be header

packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
...
+- The `authorisation` property is no longer needed, use state.configuration. The
+  `data` paramter is now `body`.
+
```
mtuchi, line 39: I think we should also add 
- Convert `formData` to `form`.The values should be the same.

stuartc:
> If the server returns an error message, we need to include it in the error object

This is great! A lot of people will be very happy. I can't remember why (if any reason) we didn't do this before? I know there has like always been security considerations around this, like someone logging headers that have API keys in them.

Perhaps it's a non-issue now - but perhaps worth asking @taylordowns2000 why we previously cut error responses from HTTP requests to almost nothing.

josephjclark:
@stuartc The error comments are more related to this branch than to main. I think the error reporting used to be mostly fine - the big problem was that the whole response object was logged, leaking all kinds of stuff and spamming the output.

The new adaptor builds an error object and returns hand-picked keys, which is a problem if we return the wrong keys.

I do want to enable access to the raw request, just in case, but there's a separate issue for that. I think we should do it soonish. Part of  the problem is it's hard to test this stuff, and when you have  a test implementation it's tempting to lean into it a bit too hard.

stuartc (approved):

Issue #414, opened 2023-10-19
Commcare: don't use language-http


josephjclark:
commcare currently depends on language http. It also uses common.http, which is a bit mad

We should refactor so that it uses the new http helpers in common

mtuchi:
We will need a commcare sandbox for testing all functions since we don't have unit test for this adaptor

josephjclark:
Even worse, commcare also users superagent.

I'm really keen to tidy this up, but yes we will need a sandbox

aleksa-krolls:
hey @josephjclark we finally got a CommCare sandbox (LP credential `CommCare Demo Project`). 
Is this something Hunter can take on? Or Mtuchi?

josephjclark:
Awesome - it's a nice task for Hunter!

hunterachieng:
@josephjclark  Do I do the same thing as I did with openmrs?

josephjclark:
@hunterachieng yep - it's basically exactly the same thing :)

Now that we've got a sandbox we'll have to get some kind of test job setup to ensure we don't break anything.

aleksa-krolls:
hey @hunterachieng how's it going here? Any questions or blockers? cc @AishaHassen

hunterachieng:
@aleksa-krolls Its going well. will be done by today

hunterachieng:
@aleksa-krolls  Its currently under the second round of reviews. If approved, I will combine it with the next commcare issue

hunterachieng:
@aleksa-krolls  The review is done and the code has been merged to this [`branch`](https://github.com/OpenFn/adaptors/tree/release/commcare). We are going to combine all the commcare related tasks here and then release all of them together

aleksa-krolls:
Noted thanks @hunterachieng. Once released please let me and @AishaHassen know so we can check out the docs and functions and provide feedback... ty!

PR #458, opened 2024-01-12
HTTP: bump versions and deprecate old helpers

josephjclark:
This PR does two closely related things:

* Bumps version numbers of changed packages
* Deprecates the old http helpers

Note that although every adaptor has had its common dependency changes, this PR only generates releases for `common` and `http`. This is the right thing to do - otherwise we're publishing a meaningless patch update to the other adaptors.

## Deprecation vs Removal

I have eventually decided to deprecate, rather than remove, the old http helpers. It just gives us an easier migration.

If we remove the old helpers, then any adaptor which depends on them will break. This won't affect legacy/existing releases because they're all reasonably well version locked.  But it does leave the monorepo in an inconsistent state: an adaptor might now depend on common@1.12, but references functions which no longer exist. The next release of that adaptor will therefore be broken.

What we need to do is go through the adaptors and update them to use new http functions. Once they're all done, we can bump common to 2.0 and remove the old functions.

## Depreciation Details

I've moved the deprecated code into `http.deprecated.js`, which is pretty clear. There's also a massive great comment at the top.

More controversially, the deprecated functions will now log a deprecation notice unless NODE_ENV is prod. I THINK this means that in platform and lightning, users won't see the deprecation notice. But if a developer is locally editing an adaptor, they should see the notices and be inspired to migrate the adaptor.

I don't have a good way to test this so we'll just have to try it out. If it doesn't work I'll remove the log from common, no big deal.


packages/commcare/package.json
```diff
@@ -25,8 +25,8 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "workspace:1.11.1",
-    "@openfn/language-http": "workspace:^5.0.4",
+    "@openfn/language-common": "workspace:1.12.0",
+    "@openfn/language-http": "workspace:^6.0.0",
```
josephjclark, line 29: This could be trouble - commcare is likely to break!

josephjclark: Commcare is actually a nightmare - it uses common.http, http.post, AND superagent!!

Luckily it only uses a very simple version of http.post. So it won't break.

josephjclark: There is an issue for this: https://github.com/OpenFn/adaptors/issues/414

PR #460, opened 2024-01-12
New template using HTTP utils

### LINKED ISSUE #321: Update template to use new common.http

josephjclark:
This PR updates the `template` package to use the new HTTP utils.

* Adaptor logic is updated to show current best pratice
* Tests updated to show some neat new patterns.

To make this a little slicker I've added a  couple more util functions into common.http (inspired by @mtuchi's work in #425)

Closes #321


josephjclark:
@taylordowns2000 can I ask you to ignore the diffs in `common` and `http` and take a look at how `template` looks now?

taylordowns2000 (approved):

taylordowns2000:
@josephjclark , really nice and clean. is commonRequest following the new pattern? (and is it shared _under_ language-http?)

josephjclark:
@taylordowns2000 `commonRequest` comes from `common`. It's the same helper function that the HTTP adaptor uses.

Issue #464, opened 2024-01-18
Salesforce: add function(s) to make a request to a REST API 


daissatou2:
Add functionality to the SF adaptor to make a request to a REST API. Add this to the Salesforce adaptor once you get it working via HTTP. 

Test this using the WaterAid credential in LP named: "Water Aid - API Only Sandbox". If you POST to the following endpoint the system will create a Contact: `wateraiduk--config2.sandbox.my.salesforce.com/v33.0/actions/custom/flow/POC_OpenFN_Test_Flow`

After sending the request confirm with Tom via Slack that the correct resource was created. 

See Slack [thread](https://openfn.slack.com/archives/C068N0ZC2DV/p1705489875769089) with the original request. See helpful Docs below:
1. https://help.salesforce.com/s/articleView?id=sf.flow_distribute_system_rest.htm&type=5
2. https://astreait.com/Calling-REST-API-from-Flows-in-Salesforce/
3. https://www.asagarwal.com/step-by-step-guide-to-calling-flow-from-an-api/
4. https://www.postman.com/salesforce-developers/workspace/salesforce-developers/request/12721794-684df392-aebc-4ff8-9ff7-30f6e4e88b74
5. https://www.salesforceben.com/introduction-salesforce-flow/#:~:text=What%20are%20Salesforce%20Flows%3F,help%20of%20a%20Salesforce%20developer

daissatou2:
Moving to `icebox`. 

Message from Tom: "Let's leave this parked for now. We are currently not able to call the 3rd Party functionality within Salesforce Flow until we get clarification from that side, if we get a reply at all so don't want to currently devote any further time to these efforts."

mtuchi:
This was done and released on https://github.com/OpenFn/adaptors/pull/465

PR #465, opened 2024-01-18
wip: export request

mtuchi:
## Summary

Add `request(path, options, callback)` function in salesforce 

## Details

Using `state.connection.request()` You can now send REST API request with given HTTP request info, with connected session information.

## Issues

Ref #464 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -782,7 +782,52 @@ export function steps(...operations) {
 export function toUTF8(input) {
   return anyAscii(input);
 }
+/**
+ * Send REST API request with given HTTP request info, with connected session information.
+ *
+ * @param {String} path - URL to send HTTP request
+ * @param {Object} options - HTTP API request options
+ * @param {String} options.method - HTTP method URL to send HTTP request
+ * @param {Object} [options.headers] - HTTP request headers in hash object (key-value)
+ * @param {Object} [options.body] - HTTP request body
+ * @param {Function} callback - A callback to execute once the record is retrieved
+ * @returns {Operation}
+ */
 
+export function request(path, options, callback) {
+  return async state => {
+    const { connection } = state;
+    const [resolvedPath, resolvedOptions] = newExpandReferences(
+      state,
+      path,
+      options
+    );
+    const {
+      method = 'GET',
+      body,
+      headers = { 'content-type': 'application/json' },
+    } = resolvedOptions;
+
+    const requestOptions = {
+      url: resolvedPath,
+      method,
+      headers,
+      body: JSON.stringify(body),
+    };
+
+    await connection.request(requestOptions).then(response => {
+      const nextState = {
+        ...composeNextState(state, response),
+        response,
+      };
+
+      if (callback) return callback(nextState);
+      return nextState;
+    });
+
+    return state;
+  };
+}
 // Note that we expose the entire axios package to the user here.
 import axios from 'axios';
 


packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,56 @@ export function toUTF8(input) {
...
+    } = resolvedOptions;
+
+    const requestOptions = {
+      url: resolvedPath,
+      method,
+      headers,
+      body: JSON.stringify(body),
+    };
+
+    await connection.request(requestOptions).then(response => {
```
josephjclark, line 823: You shouldn't `await` AND `then`. Pick one! You'll have cleaner code if you just use `await`

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,56 @@ export function toUTF8(input) {
...
+      method = 'GET',
+      body,
+      headers = { 'content-type': 'application/json' },
+    } = resolvedOptions;
+
+    const requestOptions = {
+      url: resolvedPath,
+      method,
+      headers,
+      body: JSON.stringify(body),
```
josephjclark, line 820: I don't think that automatically stringifying the body is a good idea. Maybe if it's truthy and not a string its ok.

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,56 @@ export function toUTF8(input) {
...
+    };
+
+    await connection.request(requestOptions).then(response => {
+      const nextState = {
+        ...composeNextState(state, response),
+        response,
+      };
+
+      if (callback) return callback(nextState);
```
josephjclark, line 829: If you default the `callback` parameter to `s => s`, you can just do `return callback(nextState)` here

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,56 @@ export function toUTF8(input) {
   return anyAscii(input);
 }
 
+/**
```
josephjclark, line 786: I don't like the salesforce docs. Try this:
``````
/**
 * Send a HTTP request using connected session information.
 *
 * @example
 * request('/actions/custom/flow/POC_OpenFN_Test_Flow', {
 *   method: 'POST',
 *   body: { inputs: [{}] },
 * });
 * @param {String} url - Relative or absolute URL to request from
 * @param {Object} options - Request options
 * @param {String} options.method - HTTP method to use. Defaults to GET
 * @param {Object} [options.headers] - Object of request headers
 * @param {Object} [options.body] - Request body
 * @param {Function} callback - A callback to execute once the request is complete
 * @returns {Operation}
 */
 ```

File: packages/salesforce/src/Adaptor.js [hunk 5]
```diff
@@ -783,6 +783,56 @@ export function toUTF8(input) {
...
+    const requestOptions = {
+      url: resolvedPath,
+      method,
+      headers,
+      body: JSON.stringify(body),
+    };
+
+    await connection.request(requestOptions).then(response => {
+      const nextState = {
+        ...composeNextState(state, response),
```
josephjclark, line 825: This doesn't look right.

The jsdocs aren't helpful, but it looks like the `request` function just returns the response body to us. In which case, there's no point in writing it to state root? So I think 
```
const nextState = composeNextState(state, response)
```

josephjclark: And actually you should rename `response` to `result`

josephjclark: If I'm wrong and the request really does return a response object, then we need to do some work to write `response.body` to `state.data`.

josephjclark (changes requested):

packages/salesforce/src/Adaptor.js
mtuchi: From our last sync, we agreed on `options.json` - A JSON string request body

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,53 @@ export function toUTF8(input) {
...
+ * @example
+ * request('/actions/custom/flow/POC_OpenFN_Test_Flow', {
+ *   method: 'POST',
+ *   json: { inputs: [{}] },
+ * });
+ * @param {String} url - Relative or absolute URL to request from
+ * @param {Object} options - Request options
+ * @param {String} options.method - HTTP method to use. Defaults to GET
+ * @param {Object} [options.headers] - Object of request headers
+ * @param {Object} [options.json] - A JSON string request body
```
josephjclark, line 798: This is not a "JSON string" - it's just an object. JSON object maybe.

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,53 @@ export function toUTF8(input) {
...
+ * Send a HTTP request using connected session information.
+ *
+ * @example
+ * request('/actions/custom/flow/POC_OpenFN_Test_Flow', {
+ *   method: 'POST',
+ *   json: { inputs: [{}] },
+ * });
+ * @param {String} url - Relative or absolute URL to request from
+ * @param {Object} options - Request options
+ * @param {String} options.method - HTTP method to use. Defaults to GET
```
josephjclark, line 796: method is optional so `[options.method]`.

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,53 @@ export function toUTF8(input) {
...
+    const [resolvedPath, resolvedOptions] = newExpandReferences(
+      state,
+      path,
+      options
+    );
+    const {
+      method = 'GET',
+      json,
+      body,
+      headers = { 'content-type': 'application/json' },
```
josephjclark, line 816: We should only set this header if `json` is truthy

packages/salesforce/src/Adaptor.js
```diff
@@ -783,6 +783,53 @@ export function toUTF8(input) {
...
+ * request('/actions/custom/flow/POC_OpenFN_Test_Flow', {
+ *   method: 'POST',
+ *   json: { inputs: [{}] },
+ * });
+ * @param {String} url - Relative or absolute URL to request from
+ * @param {Object} options - Request options
+ * @param {String} options.method - HTTP method to use. Defaults to GET
+ * @param {Object} [options.headers] - Object of request headers
+ * @param {Object} [options.json] - A JSON string request body
+ * @param {Object} [options.body] - Request body
```
josephjclark, line 799: The body should probably have a string type here - although technically it's any body type supported by jsforce.request. So maybe no type is more appropriate

josephjclark (changes requested):
Just a couple of little tweaks please!

packages/salesforce/src/Adaptor.js
mtuchi: @josephjclark from jsforce, the type is `String` - https://github.com/jsforce/jsforce/blob/master/lib/api/chatter.js#L60

josephjclark: Great, thanks

PR #470, opened 2024-01-25
`OAuth2` support in salesforce

mtuchi:
## Summary

Add support for `OAuth2` in salesforce adaptor

## Details

Improved the `createConnection` function to handle both `accessToken` connection and Basic Auth connection. I have created two functions [`createBasicAuthConnection` and `createAccessTokenConnection`] that gets called in `createConnection` depending on configuration-schema. I have also created another configuration-schema for `OAuth` which requires `other_params.instance_url` and `access_token`. 

Also in `configuration-schema` I have removed the `securityToken` from `required` property since it's an optional value

## Review Checklist

### DIFF
--- a/packages/salesforce/configuration-schema.json
+++ b/packages/salesforce/configuration-schema.json
@@ -55,7 +55,6 @@
   "required": [
     "loginUrl",
     "username",
-    "password",
-    "securityToken"
+    "password"
   ]
 }
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -668,66 +668,17 @@ function setApiVersion(apiVersion) {
  * @returns {State}
  */
 function createConnection(state) {
-  const {
-    loginUrl,
-    username,
-    password,
-    apiVersion,
-    instanceUrl,
-    accessToken,
-    clientId,
-    clientSecret,
-    redirectUri,
-  } = state.configuration;
+  const { loginUrl, instanceUrl } = state.configuration;
 
-  if (!loginUrl) {
-    throw new Error('loginUrl missing from configuration.');
+  if (loginUrl) {
+    return createBasicAuthConnection(state);
+  } else if (instanceUrl) {
+    return createAccessTokenConnection(state);
+  } else {
+    throw new Error(
+      'Invalid configuration. Provide either loginUrl for basic auth or instanceUrl for accessToken auth.'
+    );
   }
-
-  const conn = new jsforce.Connection({
-    oauth2: {
-      // you can change loginUrl to connect to sandbox or prerelease env.
-      // loginUrl : 'https://test.salesforce.com',
-      clientId,
-      clientSecret,
-      redirectUri,
-    },
-  });
-
-  conn.login(username, password, function (err, userInfo) {
-    if (err) {
-      console.error(err);
-      return err;
-    }
-    // Now you can get the access token and instance URL information.
-    // Save them to establish connection next time.
-    console.log(conn.accessToken);
-    console.log(conn.instanceUrl);
-    // logged in user property
-    console.log('User ID: ' + userInfo.id);
-    console.log('Org ID: ' + userInfo.organizationId);
-    // ...
-  });
-
-  return {
-    ...state,
-    connection: conn,
-  };
-
-  // return {
-  //   ...state,
-  //   connection: new jsforce.Connection({ instanceUrl, accessToken }),
-  // };
-
-  // return {
-  //   ...state,
-  //   connection: apiVersion
-  //     ? new jsforce.Connection({
-  //         loginUrl,
-  //         version: setApiVersion(apiVersion),
-  //       })
-  //     : new jsforce.Connection({ loginUrl }),
-  // };
 }
 
 /**
@@ -754,6 +705,65 @@ function login(state) {
       // })
       .then(() => state)
   );
+}
+async function createBasicAuthConnection(state) {
+  const { loginUrl, apiVersion, username, password, securityToken } =
+    state.configuration;
+
+  if (!loginUrl || !username || !password) {
+    throw new Error(
+      'Incomplete configuration for basic authentication. Provide loginUrl, username, password, and securityToken if required.'
+    );
+  }
+
+  const connectionOptions = { loginUrl };
+
+  if (apiVersion) {
+    connectionOptions.version = setApiVersion(apiVersion);
+  }
+
+  const connection = new jsforce.Connection(connectionOptions);
+
+  console.info(`Logging in as ${username}.`);
+  await connection.login(
+    username,
+    securityToken ? password + securityToken : password
+  );
+  // NOTE: Uncomment this to debug connection issues.
+  // .then(response => {
+  //   console.log(connection);
+  //   console.log(response);
+  //   return state;
+  // })
+
+  return {
+    ...state,
+    connection,
+  };
+}
+
+function createAccessTokenConnection(state) {
+  const { apiVersion, instanceUrl, accessToken } = state.configuration;
+
+  if (!instanceUrl || !accessToken) {
+    throw new Error(
+      'instanceUrl and accessToken are required for OAuth authentication.'
+    );
+  }
+
+  const connectionOptions = {
+    instanceUrl,
+    accessToken,
+  };
+
+  if (apiVersion) {
+    connectionOptions.version = setApiVersion(apiVersion);
+  }
+
+  return {
+    ...state,
+    connection: new jsforce.Connection(connectionOptions),
+  };
 }
 
 /**
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -646,60 +646,43 @@ export function reference(position) {
   return state => state.references[position].id;
 }
 
-function setApiVersion(apiVersion) {
+function setApiVersion(state, connectionOptions) {
+  const { apiVersion } = state.configuration;
+
   const apiVersionRegex = /^\d{2}\.\d$/;
-  let version = '52.0';
+
   if (apiVersion && apiVersionRegex.test(apiVersion)) {
     console.log('Using Salesforce API version', apiVersion);
-    version = apiVersion;
+    return { ...connectionOptions, version: apiVersion };
   } else {
-    console.log('Invalid salesforce apiVersion', apiVersion);
-    console.log('Using Salesforce API version', version);
+    console.log('apiVersion is not defined');
+    console.log('We recommend using Salesforce API version 52.0 or latest');
+    return connectionOptions;
   }
-
-  return version;
 }
 
 /**
  * Establish a connection to Salesforce using username and password.
- * @example
- * createBasicAuthConnection(state)
- * @function
+ * @function createBasicAuthConnection
+ * @private
  * @param {State} state - Runtime state.
  * @returns {State}
  */
 async function createBasicAuthConnection(state) {
-  console.log('Attempting to connect with Basic Auth');
+  const { loginUrl, username, password, securityToken } = state.configuration;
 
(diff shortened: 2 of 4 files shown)


packages/salesforce/configuration-schema.json
```diff
@@ -55,7 +55,6 @@
   "required": [
     "loginUrl",
     "username",
-    "password",
-    "securityToken"
```
taylordowns2000, line 59: @mtuchi , is `securityToken` not required anymore?

mtuchi: @taylordowns2000 according to jsforce, `securityToken` is optional - https://jsforce.github.io/document/#username-and-password-login

taylordowns2000: ah wow. @aleksa-krolls , can you confirm that we don't _need_ users to provide their security token when creating a new Salesforce credential? (is this an admin setting in a client's salesforce system? something like, "require security token for API access"?)

mtuchi: I think most salesforce instance are configured to require `securityToken` but if a client have a salesforce instance that does not require `securityToken` then the job should pass. 

The change i introduced also helps with credential creation on lighting which make the `Security Token` input optional

packages/salesforce/src/Adaptor.js
```diff
@@ -659,56 +659,103 @@ function setApiVersion(apiVersion) {
...
- * createConnection(state)
+ * createBasicAuthConnection(state)
  * @function
  * @param {State} state - Runtime state.
  * @returns {State}
  */
-function createConnection(state) {
-  const { loginUrl, apiVersion } = state.configuration;
+async function createBasicAuthConnection(state) {
+  console.log('Attempting to connect with Basic Auth');
```
josephjclark, line 672: Do we log anywhere when the connection is successful?

I would much prefer to say:

`Connected to salesforce as ${user}`

Or

`Failed to connected to salesforce as ${user}`

Ie, print the result, not the process.

When  we say "Attempting to..." it sort of shows a lack of confidence on our part, and when there's no log to follow it up it's gonna look pretty flaky.

For something routine, like logging in, we should be super confident in our logging. And when something goes wrong we should be super clear.

packages/salesforce/src/Adaptor.js
```diff
@@ -659,56 +659,103 @@ function setApiVersion(apiVersion) {
...
+
+  const { loginUrl, apiVersion, username, password, securityToken } =
+    state.configuration;
 
-  if (!loginUrl) {
-    throw new Error('loginUrl missing from configuration.');
+  if (!loginUrl || !username || !password) {
+    throw new Error(
+      'Incomplete configuration for basic authentication. Provide loginUrl, username, password, and securityToken if required.'
```
josephjclark, line 679: The `if required` bit is a little confusing - which keys are required? We don't know.  The user probably does.

I would be tempted to just let the login fail and report the error message from salesforce back to the user. Assuming we get a good error.

The argument against this is that if our users regularly are omitting certain information (like the login url, which seems like an obvious thing to forget), we should warn. I presume that's why we did this for loginUrl. I don't think it's useful for username and password though.

packages/salesforce/src/Adaptor.js
```diff
@@ -659,56 +659,103 @@ function setApiVersion(apiVersion) {
...
+    );
   }
 
+  const connectionOptions = { loginUrl };
+
+  if (apiVersion) {
+    connectionOptions.version = setApiVersion(apiVersion);
```
josephjclark, line 686: hmm. `setApiVersion` doesn't actually set anything.

I'm not sure how big the change is, but can we consider either:
a) changing this to `getApiVersion(state)`
b)  changing it to `setApiVersion(state, connectionOptions)` but pass in the `connectionOptions` object. The function will set the `apiVersion` key if it exists on state.config/

packages/salesforce/src/Adaptor.js
```diff
@@ -659,56 +659,103 @@ function setApiVersion(apiVersion) {
...
-          version: setApiVersion(apiVersion),
-        })
-      : new jsforce.Connection({ loginUrl }),
+    connection,
   };
 }
 
 /**
- * Performs a login.
+ * Establish a connection to Salesforce using Access Token.
```
josephjclark, line 711: I don't think that this or `createBasicAuthConnection` should be documented like this. They're both private and internal functions - `createConnection` is the public interface (and even then it's an internal function, it shouldn't be documented for job writers)

packages/salesforce/src/Adaptor.js
mtuchi: I agree on with the `Connected to salesforce as ${username}` but on Failure i don't think we should log `Failed to connect as ${username}` because this suggest that the username is wrong while it could be a number of reasons. 

I would suggest we leave the error message that comes from salesforce which looks like this
```
INVALID_LOGIN: Invalid username, password, security token; or user locked out.
```
cc @josephjclark @taylordowns2000

taylordowns2000: @mtuchi , we added the username there (if i'm remembering this correctly) because it's incredibly helpful for the end-user to know which SF user they have to debug! users _do_ get locked out, tokens become invalid, etc., etc.

printing the username is great for the salesforce admin. i don't think it will make them think that the username is wrong, i think it will help them figure out whether the password, token, or user status in salesforce is wrong.

josephjclark: I don't think that error implies that username is the fault. I agree with taylor that it's a useful bit of information to report.

I also agree that the original salesforce message is fine.

We can do both!

```
console.error(`Failed to connected to salesforce as ${user}`)
console.error(e.message)  // or whatever
```

Actually, if we we throw the error, I think the runtime should report it (and if it doesn't, that's something I need to fix!)

```
console.error(`Failed to connected to salesforce as ${user}`)
throw e
````

File: packages/salesforce/src/Adaptor.js (hunk 4 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js:686]: @josephjclark I choose option `b`. But i think i need to refactor my implementation of `setApiVersion`. Because i don't want to set the api version if i haven't specified the version in `state.configuration`. and currently my implementation of `setApiVersion` defaults to `v52.0`

File: packages/salesforce/src/Adaptor.js (hunk 3 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js:679]: @josephjclark 
The required key are `username, password and loginUrl`. I used the jsforce language `if required` because not all salesforce instance requires `securityToken` - https://jsforce.github.io/document/#username-and-password-login

If a user does not provide those required properties, the salesforce error log will say
```
INVALID_LOGIN: Invalid username, password, security token; or user locked out.
```
To me that error suggests i have entered the wrong information rather than telling me i need to fill in the required values in my `state.configuration` 

So if it's a warning i think it should consider both required values rather than looking for `loginUrl` only

File: packages/salesforce/src/Adaptor.js (hunk 2 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js:672]: Okay sounds good 👍🏽 , I will use this 👇🏽
```
console.error(`Failed to connected to salesforce as ${user}`)
throw e
```

Comment (taylordowns2000) [reply packages/salesforce/src/Adaptor.js]: hrmmm, i think we're not supposed to use `console.error`, right? isn't it a different stream and won't it cause timing/printout issues?

File: packages/salesforce/src/Adaptor.js (hunk 4 above)
Comment (josephjclark) [reply packages/salesforce/src/Adaptor.js]: If it's too complicated you can leave it - it's just something that caught my eye.

File: packages/salesforce/src/Adaptor.js (hunk 5 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js:694]: @josephjclark i have update the docstring and add `@private`, Please take a look

File: packages/salesforce/src/Adaptor.js (hunk 3 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js]: Remove the if statement, salesforce will return `INVALID_LOGIN: Invalid username, password, security token; or user locked out.` If required fields are not specified in `state.configuration`

File: packages/salesforce/src/Adaptor.js (hunk 4 above)
Comment (mtuchi) [reply packages/salesforce/src/Adaptor.js]: It's not a big change, i have pushed the change with `setApiVersino(state, connOpts)` @josephjclark

File: packages/salesforce/src/Adaptor.js [hunk 6]
```diff
@@ -646,69 +646,90 @@ export function reference(position) {
...
+function createAccessTokenConnection(state) {
+  const { other_params, access_token } = state.configuration;
+  const { instance_url } = other_params;
+
+  const connectionOptions = setApiVersion(state, {
+    instanceUrl: instance_url,
+    accessToken: access_token,
+  });
+
+  const connection = new jsforce.Connection(connectionOptions);
```
josephjclark, line 710: Do we not need to do any sort of `login` here?

The connection is synchronous, how do we know we're actually connected?

packages/salesforce/src/Adaptor.js
```diff
@@ -646,69 +646,90 @@ export function reference(position) {
...
-  const { loginUrl, apiVersion } = state.configuration;
 
-  if (!loginUrl) {
-    throw new Error('loginUrl missing from configuration.');
-  }
+function createAccessTokenConnection(state) {
+  const { other_params, access_token } = state.configuration;
+  const { instance_url } = other_params;
+
+  const connectionOptions = setApiVersion(state, {
```
josephjclark, line 705: Hmm. I'm sorry but this is still just a bit too quirky. I feel like we're making an easy thing very difficult.

How about we do this?

```
function getConnection(state, options) {
	const { apiVersion } = state.configuration;
	
	const apiVersionRegex = /^\d{2}\.\d$/;

 	if (apiVersion && apiVersionRegex.test(apiVersion)) {
    	console.log('Using Salesforce API version', apiVersion);
    	options.version = apiVersion;
    } else {
     	console.log('apiVersion is not defined');
    	console.log('We recommend using Salesforce API version 52.0 or latest');
    }

	return new jsforce.Connection(options);
 }
 

 async function createBasicAuthConnection(state) {
  const { loginUrl, username, password, securityToken } = state.configuration;

  const connection = getConnection(state, { loginUrl });

  await connection
    .login(username, securityToken ? password + securityToken : password)
    .catch(e => {
      console.error(`Failed to connect to salesforce as ${username}`);
      throw e;
    });

  console.info(`Connected to salesforce as ${username}.`);

  return {
    ...state,
    connection,
  };
}

function createAccessTokenConnection(state) {
  const { other_params, access_token } = state.configuration;
  const { instance_url } = other_params;

  const connection = getConnection(state, {
    instanceUrl: instance_url,
    accessToken: access_token,
  });

  console.log(`Connected with ${connection._sessionType} session type`);

  return {
    ...state,
    connection,
  };
}
```

packages/salesforce/src/Adaptor.js
mtuchi: No login there, It's a different strategy based on auth type, 
So for basic-auth you will setup the connection then do `connection.login(connectionOptions)`. And for accessToken auth, `getConnection(state, connOpts)` will have the connection with session type `oauth`

For Basic Auth - https://jsforce.github.io/document/#username-and-password-login
For OAuth - https://jsforce.github.io/document/#access-token

packages/salesforce/src/Adaptor.js
josephjclark: @taylordowns2000 I was going to say: `console.error`should be fine - we redirect it to stdout anyway.

...  but then I remembered that this doesn't use our logger, so yes it will go to a different stream and cause timing problems.

I don't think that's the fault of the adaptor though. Let me just dig into something here.

josephjclark (approved):
Approved subject to something in `console.error` that I just want to look into.

josephjclark:
I don't have a quick solution to the console.error thing. For the moment I've just renamed it to console.log.

PR #471, opened 2024-01-30
Release: salesforce 4.4.0

josephjclark:
* Request function #465 
* OAuth #470 

I've also merged in #469

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,6 +1,10 @@
 name: Build & deploy adaptor documentation
 
 on:
+  workflow_run:
+    workflows: [Publish]
+    types:
+      - completed
   push:
     branches: [main]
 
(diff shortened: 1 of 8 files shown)


.github/workflows/docs.yaml
```diff
@@ -1,6 +1,10 @@
 name: Build & deploy adaptor documentation
 
 on:
+  workflow_run:
```
mtuchi, line 4: @josephjclark what does this logic do ?

mtuchi (approved):
@josephjclark this looks good to me 👍🏽 
Just one comment, i don't fully understand the changes in `docs.yaml`
If you're happy with the changes go ahead and merge

.github/workflows/docs.yaml
josephjclark: Oh don't worry about this, it's entirely unrelated.

It just ensures that docs don't get build until after a release is published

Issue #472, opened 2024-02-01
Update OAuth Salesforce Configuration 


mtuchi:
### Description

Update the `oauth-configuration-schema.json` to 
```
{
   "access_token": "",
   "instance_url": ""
}
```
and `createAccessTokenConnection` function 
https://github.com/OpenFn/adaptors/blob/main/packages/salesforce/src/Adaptor.js#L687-L702

PR #473, opened 2024-02-01
Put `instance_url` at the root of configuration schema for Salesforce OAuth

elias-ba:
## Summary
Update `instance_url` in `oauth-configration-schema.json` to be in the root of configuration schema for Salesforce OAuth. And update `createAccessTokenConnection` in `adaptors.js`

## Issues 
Ref #472 

## Review Checklist

### DIFF
--- a/.changeset/itchy-spiders-scream.md
+++ b/.changeset/itchy-spiders-scream.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-salesforce': major
+---
+
+Remove instance_url under other_params and put it at the root level of the
+configuration schema
--- a/.changeset/itchy-spiders-scream.md
+++ b/.changeset/itchy-spiders-scream.md
@@ -1,6 +1,0 @@
----
-'@openfn/language-salesforce': major
----
-
-Remove instance_url under other_params and put it at the root level of the
-configuration schema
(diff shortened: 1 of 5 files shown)


.changeset/itchy-spiders-scream.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-salesforce': major
```
mtuchi, line 2: I don't think we want this change to be a major version bump. Ideally this should be a patch. It make sense to be a major version bump because of breaking changes but the latest release is not used on any project yet. 

@josephjclark Do you have any opinions on this ?

josephjclark: Agreed - I think minor is fine

josephjclark:
We should:
1) Deprecate the current latest version on npm, because it won't work with Lightning
2) In Lightning, find a way to automatically ignore/block deprecated adaptor versions

mtuchi:
> 1. Deprecate the current latest version on npm, because it won't work with Lightning ✅

I have deprecate `@openfn/language-salesforce@v4.4.0` cc @elias-ba 
Screenshot 2024-02-02 at 12 36 49 PM

josephjclark (approved):

Issue #474, opened 2024-02-14
Update openmrs to use the new http module


taylordowns2000:
Based on the patterns specified in https://github.com/OpenFn/adaptors/issues/283, https://github.com/OpenFn/adaptors/issues/320 , and https://github.com/OpenFn/adaptors/issues/315 , migrate [language-openmrs](https://github.com/OpenFn/adaptors/tree/main/packages/openmrs) off of request and superagent.

The new version should:
1. Use the standard format for http requests defined in https://github.com/OpenFn/adaptors/blob/main/packages/template/src/Utils.js
2. Have no reference to `axios`, `superagent` or the old npm `request` module. (Should be handled in the point above.)
3. Remove the [custom logger](https://github.com/OpenFn/adaptors/blob/main/packages/openmrs/src/Utils.js#L3-L8) - this is a pattern we want to avoid/remove wherever we see it
4. Be tested against the [OpenMRS demo](https://openmrs.org/demo/)

@mtuchi will be able to provide additional guidance here

PR #475, opened 2024-02-20
feat:add new request utils

hunterachieng:
Signed-off-by: Hunter Achieng (achienghunter@gmail.com)

## Summary

Changing `openmrs` implementation to use the same one as `http`

## Details

The current implementation of `openmrs` contains depreciated library integrations and we are migrating to a new system which is to be adopted by all adaptor modules. 

## Issues

#474 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/openmrs/README.md
+++ b/packages/openmrs/README.md
@@ -3,6 +3,8 @@
 Language Pack for building expressions and operations for working with the
 [OpenMRS API](https://wiki.openmrs.org/display/docs/API).
 
+[OpenMRS Rest API](https://rest.openmrs.org/?javascript#openmrs-rest-api)
+
 [OpenMRS Data Model Explorer](http://burkeware.com/openmrs-data-model/openmrs-data-model-1.11.html#)
 
 ## Documentation
--- a/packages/openmrs/src/Adaptor.js
+++ b/packages/openmrs/src/Adaptor.js
@@ -5,6 +5,7 @@ import {
 import request from 'superagent';
 import { Log, handleError, handleResponse } from './Utils';
 
+
 let agent = null;
 
 /**
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -1,4 +1,10 @@
 import { composeNextState } from '@openfn/language-common';
+import {
+  request as commonRequest,
+  expandReferences,
+  logResponse,
+  makeBasicAuthHeader,
+} from '@openfn/language-common/util';
 
 export const Log = {
   success: message => console.log(`✓ Success at ${new Date()}:\n∟ ${message}`),
@@ -42,3 +48,84 @@ const isArray = variable => !!variable && variable.constructor === Array;
 export function nestArray(data, key) {
   return isArray(data) ? { [key]: data } : data;
 }
+
+export function addBasicAuth(configuration = {}, headers) {
+  const { username, password } = configuration;
+  if (username && password) {
+    Object.assign(headers, makeBasicAuthHeader(username, password));
+  }
+}
+
+function encodeFormBody(data) {
+  const form = new FormData();
+  for (const [key, value] of Object.entries(data)) {
+    form.append(key, value);
+  }
+  return form;
+}
+
+export function request(method, path, params, callback = s => s) {
+  return state => {
+    const [resolvedPath, resolvedParams = {}] = expandReferences(
+      state,
+      path,
+      params
+    );
+
+    let { body, headers = {} } = resolvedParams;
+
+    if (resolvedParams.json) {
+      console.warn(
+        'WARNING: The `json` option has been deprecated. Use `body` instead'
+      );
+      body = resolvedParams.json;
+    }
+
+    if (resolvedParams.form) {
+      body = encodeFormBody(resolvedParams.form);
+    }
+
+    const baseUrl = state.configuration?.baseUrl;
+
+    addBasicAuth(state.configuration, headers);
+
+    const maxRedirections =
+      resolvedParams.maxRedirections ??
+      (resolvedParams.followAllRedirects === false ? 0 : 5);
+
+    const tls = resolvedParams.tls ?? resolvedParams.agentOptions;
+
+    if (resolvedParams.agentOptions) {
+      console.warn(
+        'WARNING: The `agentOptions` option has been deprecated. Use `tls` instead'
+      );
+    }
+
+    const options = {
+      ...resolvedParams,
+      headers,
+      baseUrl,
+      body,
+      tls,
+      maxRedirections,
+    };
+
+    return commonRequest(method, resolvedPath, options)
+      .then(response => {
+        logResponse(response);
+
+        return {
+          ...composeNextState(state, response.body),
+          response,
+        };
+      })
+      .then(callback)
+      .catch(err => {
+        logResponse(err);
+
+        throw err;
+      });
+  };
+}
+
+
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -1,10 +1,4 @@
 import { composeNextState } from '@openfn/language-common';
-import {
-  request as commonRequest,
-  expandReferences,
-  logResponse,
-  makeBasicAuthHeader,
-} from '@openfn/language-common/util';
 
 export const Log = {
   success: message => console.log(`✓ Success at ${new Date()}:\n∟ ${message}`),
@@ -47,83 +41,4 @@ const isArray = variable => !!variable && variable.constructor === Array;
 
 export function nestArray(data, key) {
   return isArray(data) ? { [key]: data } : data;
-}
-
-export function addBasicAuth(configuration = {}, headers) {
-  const { username, password } = configuration;
-  if (username && password) {
-    Object.assign(headers, makeBasicAuthHeader(username, password));
-  }
-}
-
-function encodeFormBody(data) {
-  const form = new FormData();
-  for (const [key, value] of Object.entries(data)) {
-    form.append(key, value);
-  }
-  return form;
-}
-
-export function request(method, path, params, callback = s => s) {
-  return state => {
-    const [resolvedPath, resolvedParams = {}] = expandReferences(
-      state,
-      path,
-      params
-    );
-
-    let { body, headers = {} } = resolvedParams;
-
-    if (resolvedParams.json) {
-      console.warn(
-        'WARNING: The `json` option has been deprecated. Use `body` instead'
-      );
-      body = resolvedParams.json;
-    }
-
-    if (resolvedParams.form) {
-      body = encodeFormBody(resolvedParams.form);
-    }
-
-    const baseUrl = state.configuration?.baseUrl;
-
-    addBasicAuth(state.configuration, headers);
-
-    const maxRedirections =
-      resolvedParams.maxRedirections ??
-      (resolvedParams.followAllRedirects === false ? 0 : 5);
-
-    const tls = resolvedParams.tls ?? resolvedParams.agentOptions;
-
-    if (resolvedParams.agentOptions) {
-      console.warn(
-        'WARNING: The `agentOptions` option has been deprecated. Use `tls` instead'
-      );
-    }
-
-    const options = {
-      ...resolvedParams,
-      headers,
-      baseUrl,
-      body,
-      tls,
-      maxRedirections,
-    };
-
-    return commonRequest(method, resolvedPath, options)
-      .then(response => {
-        logResponse(response);
-
-        return {
-          ...composeNextState(state, response.body),
-          response,
-        };
-      })
-      .then(callback)
-      .catch(err => {
-        logResponse(err);
-
-        throw err;
-      });
-  };
 }
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -1,4 +1,10 @@
 import { composeNextState } from '@openfn/language-common';
+import {
+  request as commonRequest,
+  expandReferences,
+  logResponse,
+  makeBasicAuthHeader,
+} from '@openfn/language-common/util';
 
 export const Log = {
   success: message => console.log(`✓ Success at ${new Date()}:\n∟ ${message}`),
@@ -41,4 +47,83 @@ const isArray = variable => !!variable && variable.constructor === Array;
 
 export function nestArray(data, key) {
   return isArray(data) ? { [key]: data } : data;
+}
+
+export function addBasicAuth(configuration = {}, headers) {
+  const { username, password } = configuration;
+  if (username && password) {
+    Object.assign(headers, makeBasicAuthHeader(username, password));
+  }
+}
+
+function encodeFormBody(data) {
+  const form = new FormData();
+  for (const [key, value] of Object.entries(data)) {
+    form.append(key, value);
+  }
+  return form;
+}
+
+export function request(method, path, params, callback = s => s) {
+  return state => {
+    const [resolvedPath, resolvedParams = {}] = expandReferences(
+      state,
+      path,
+      params
+    );
+
+    let { body, headers = {} } = resolvedParams;
+
+    if (resolvedParams.json) {
+      console.warn(
+        'WARNING: The `json` option has been deprecated. Use `body` instead'
+      );
+      body = resolvedParams.json;
+    }
+
+    if (resolvedParams.form) {
+      body = encodeFormBody(resolvedParams.form);
+    }
+
+    const baseUrl = state.configuration?.baseUrl;
+
+    addBasicAuth(state.configuration, headers);
+
+    const maxRedirections =
+      resolvedParams.maxRedirections ??
+      (resolvedParams.followAllRedirects === false ? 0 : 5);
+
+    const tls = resolvedParams.tls ?? resolvedParams.agentOptions;
+
+    if (resolvedParams.agentOptions) {
+      console.warn(
+        'WARNING: The `agentOptions` option has been deprecated. Use `tls` instead'
+      );
+    }
+
+    const options = {
+      ...resolvedParams,
+      headers,
+      baseUrl,
+      body,
+      tls,
+      maxRedirections,
+    };
+
+    return commonRequest(method, resolvedPath, options)
+      .then(response => {
+        logResponse(response);
+
+        return {
+          ...composeNextState(state, response.body),
+          response,
+        };
+      })
+      .then(callback)
+      .catch(err => {
+        logResponse(err);
+
+        throw err;
+      });
+  };
 }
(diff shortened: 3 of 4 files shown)


pnpm-lock.yaml
mtuchi, line 1: Hey @hunterachieng i think you have changed the `lockVersion` unintentionally. Please revert the changes in this file and then remove the unused packages. Make sure your using the latest version of pnpm `8.x`, It's possible your using the old version of `pnpm` that is why we have this huge diff in this file.

mtuchi: Make sure you run `pnpm install` after revert the `lockVersion` and removing the unused packages.

I think the `lockVersion` changes is what causing this error 
https://github.com/OpenFn/adaptors/actions/runs/7970961682/job/21759652298?pr=475

packages/mailgun/test/index.js
```diff
@@ -5,7 +5,7 @@ const { execute, send } = Adaptor;
 import { fields, field } from '@openfn/language-common';
 
 describe('execute', () => {
-  it('sends an Email and expects a confirmation response', done => {
+  it.skip('sends an Email and expects a confirmation response', done => {
```
taylordowns2000, line 8: please open the issue to fix this before we merge this PR in

packages/openmrs/README.md
```diff
@@ -3,6 +3,8 @@
 Language Pack for building expressions and operations for working with the
 [OpenMRS API](https://wiki.openmrs.org/display/docs/API).
 
+[OpenMRS Rest API](https://rest.openmrs.org/?javascript#openmrs-rest-api)
```
taylordowns2000, line 6: Is the link on line 4 still valid? Seems like this is the current API docs site and should replace the link on line 4. Thoughts?

packages/openmrs/src/Adaptor.js
```diff
@@ -60,23 +36,7 @@ async function login(state) {
  * @returns {Operation}
  */
 export function getPatient(uuid, callback = false) {
-  return state => {
-    Log.info(`Searching for patient with uuid: ${uuid}`);
```
taylordowns2000, line 64: I like this level of logging. My sense is that if this was here before, we should keep it. Before `return sendRequest` you could add:

```js
console.log(`Searching for patient with uuid: ${uuid}`);
```

packages/openmrs/src/Adaptor.js
```diff
@@ -101,24 +61,12 @@ export function getPatient(uuid, callback = false) {
  * @returns {Operation}
  */
 export function createEncounter(data, callback = false) {
-  return state => {
-    const { instanceUrl } = state.configuration;
-    const body = expandReferences(data)(state);
-    const url = `${instanceUrl}/ws/rest/v1/encounter`;
-
-    Log.info(`Creating an encounter.`);
```
taylordowns2000, line 109: Same as above... let's keep this logging but use `console.log`, rather than that custom Logger.

packages/openmrs/src/Adaptor.js
```diff
@@ -60,23 +36,7 @@ async function login(state) {
...
-    const { instanceUrl } = state.configuration;
-    const defaultQuery = { v: 'full', limit: 1 };
-    const url = `${instanceUrl}/ws/rest/v1/patient/${uuid}`;
-
-    return agent
-      .get(url)
-      .accept('json')
-      .query(defaultQuery)
-      .then(response => {
-        Log.success(`Found patient.`);
```
taylordowns2000, line 74: Can you preserve this with `console.log` too?

packages/openmrs/src/Adaptor.js
```diff
@@ -101,24 +61,12 @@ export function getPatient(uuid, callback = false) {
...
-    const url = `${instanceUrl}/ws/rest/v1/encounter`;
-
-    Log.info(`Creating an encounter.`);
-
-    return agent
-      .post(url)
-      .type('json')
-      .send(body)
-      .then(response => {
-        Log.success(`Created an encounter.`);
```
taylordowns2000, line 116: Same as above

packages/openmrs/src/Adaptor.js
```diff
@@ -189,34 +118,7 @@ export function post(path, data, callback = false) {
  * @returns {Operation}
  */
 export function searchPatient(query, callback = false) {
-  return state => {
-    const qs = expandReferences(query)(state);
-    Log.info(`Searching for patient with name: ${qs.q}`);
```
taylordowns2000, line 194: Same here

packages/openmrs/src/Adaptor.js
```diff
@@ -189,34 +118,7 @@ export function post(path, data, callback = false) {
...
-    return agent
-      .get(url)
-      .accept('json')
-      .query(qs)
-      .then(response => {
-        const data = response.body;
-        const count = data.results.length;
-
-        if (count > 0) {
-          Log.success(
```
taylordowns2000, line 208: And here

packages/openmrs/src/Adaptor.js
```diff
@@ -189,34 +118,7 @@ export function post(path, data, callback = false) {
...
-            `Search successful. Returned ${count} patient${
-              count > 1 ? 's' : ''
-            }.`
-          );
-          return handleResponse(response, state, callback);
-        } else {
-          Log.warn(`${count} records were found.`);
```
taylordowns2000, line 215: This feels like an important one too!

taylordowns2000 (changes requested):
This looks really good! please note the logging comments... i'd like to preserve the logging _logic_ from the old version, but use `console.log` instead of `Logger.blah`.

Other than that I think we're good to go. Would love to see a live demo of this tomorrow!

packages/openmrs/README.md
hunterachieng: > Is the link on line 4 still valid? Seems like this is the current API docs site and should replace the link on line 4. Thoughts?

I think line 6 should replace line 4 as it directly takes the users to the API docs therefore reducing the number of clicks they will have to do to get to the actual API documentation

packages/mailgun/test/index.js
josephjclark: This was actually on me to do! 

I'm doing that now. I just stopped by this PR to reference it :)

packages/openmrs/src/Adaptor.js
hunterachieng: Adding this log requires me to use this method
``` (state)=>sendRequest('GET','patient').(state).then()....```
Using this does not allow me to get the response needed when run this job but the logs before calling sendRequests are working.
@taylordowns2000 @mtuchi  What do you advice?

File: packages/openmrs/src/Adaptor.js [hunk 10]
```diff
@@ -59,23 +36,18 @@ async function login(state) {
...
+  return async state => {
+    const [resolvedUuid] = expandReferences(state, uuid);
+    console.log(`Searching for patient with uuid: ${resolvedUuid}`);
+
+    const response = await request(
+      state,
+      'GET',
+      `/ws/rest/v1/patient/${resolvedUuid}`
+    );
+
```
taylordowns2000, line 49: add `console.log('Found patient.')` if we are preserving all logs

packages/openmrs/src/Adaptor.js
```diff
@@ -100,24 +72,20 @@ export function getPatient(uuid, callback = false) {
...
-      .then(response => {
-        Log.success(`Created an encounter.`);
-
-        return handleResponse(response, state, callback);
-      })
-      .catch(handleError);
+export function createEncounter(data, callback = s => s) {
+  return async state => {
+    const [resolvedData] = expandReferences(state, data);
+    // console.log({resolvedResource});
```
taylordowns2000, line 78: remove commented code

packages/openmrs/src/Adaptor.js
```diff
@@ -100,24 +72,20 @@ export function getPatient(uuid, callback = false) {
...
+    console.log(`Creating an encounter.`);
+    const response = await request(
+      state,
+      'POST',
+      '/ws/rest/v1/encounter',
+      resolvedData
+    );
+
+    console.log(`Created encounter with new UUID: ${response.body.id}`);
```
taylordowns2000, line 87: great!!!

packages/openmrs/src/Adaptor.js
```diff
@@ -134,18 +102,29 @@ export function createEncounter(data, callback = false) {
...
+      state,
+      'GET',
+      `/ws/rest/v1/${resolvedResource}`,
+      {},
+      resolvedQuery
+    );
+    if (response.statusCode === 200) {
+      console.log(`Get operation successful for ${resolvedResource}...`);
+    } else {
+      console.log(`Get operation failed for ${resolvedResource} ...`);
```
taylordowns2000, line 124: what happens on error? i'd expect that this function should make the job exit with an error status if it's not a 200.

packages/openmrs/src/Adaptor.js
```diff
@@ -162,20 +141,28 @@ export function get(path, query, callback = false) {
...
+      state,
+      'POST',
+      `/ws/rest/v1/${resolvedResource}`,
+      resolvedData
+    );
+
+    if (response.statusCode === 201) {
+      console.log(`Successfully posted operation for ${resolvedResource}...`);
+    } else {
+      console.log(`Posting operation failed for ${resolvedResource} ...`);
```
taylordowns2000, line 162: same Q here... is this exiting with an error status? what was `handleError` doing in the previous version? we'll likely need to preserve that functionality.

packages/openmrs/src/Adaptor.js
```diff
@@ -188,34 +175,28 @@ export function post(path, data, callback = false) {
...
+      '/ws/rest/v1/patient',
+      {},
+      resolvedQuery
+    );
+
+    if (response.statusCode === 200) {
+      console.log(`Successfully got patient with name ${resolvedQuery?.q}...`);
+    } else {
+      console.log(
+        `Search operation failed for patient with name ${resolvedQuery?.q} ...`
```
taylordowns2000, line 195: same here. what happens on error?

packages/openmrs/src/Adaptor.js
```diff
@@ -499,33 +477,35 @@ export function upsert(
...
+      {},
+      resolvedQuery
+    )
       .then(resp => {
-        const resources = resp.data.body.results;
-        if (resources.length > 1) {
+        const resource = resp.body.results;
+        if (resource.length > 1) {
           throw new RangeError(
             `Cannot upsert on Non-unique attribute. The operation found more than one records for your request.`
```
taylordowns2000, line 497: make "Non-unique" lowercase: "non-unique"

taylordowns2000 (changes requested):
this is great! i've added a few comments... small things to address with logging.

the one big question I've still got is about how we're handling errors. if a get returns anything but a 200, it seems like you're _logging_ "get failed" but maybe not actually throwing any sort of error. we'll want errors to appear so that we see "red Xs" on the platform and so that users can build subsequent "on failure" paths into their workflows.

packages/openmrs/src/Adaptor.js
```diff
@@ -319,52 +291,54 @@ export function createPatient(data, callback = false) {
...
+    const response = await request(
+      state,
+      'GET',
+      `/ws/rest/v1/encounter/`,
+      {},
+      resolvedQuery
+    );
+
+    // TODO - we could also argue that this should NOT throw.
+    if (response.body.results.length == 0) {
```
josephjclark, line 336: Yeah I don't think we should throw here. Zero results is an empty array, not an exception.

Also this is a change in behavior from the latest release, which we don't really want in this PR.

josephjclark (commented):
Hi Hunter, this looks lovely.

The re-implementation is nice and clean. I've included a few nitpicks but we're nearly there.

Just to be clear - the aim of this PR should be to switch out the backing http functions without changing the adaptor's behaviour or documentation (or at least, as little as possible).

We've already slipped away from that brief with logging changes. I'm happy enough with it, but generally we should keep the scope on any piece of work as tight as possible.

So some of my comments are around unncessary user-facing changes.

There are a couple more changes that we need as well:

- Remove `src/Client.js` and any related tests. I don't think there's any other reference to it.
- Update the `Make a request to any OpenMRS endpoint` section of the readme. This example should use your get or post functions

When everything else is done, please run `pnpm changeset` from the monorepo and add a release note to communicate your changes to end users. I'm happy to help you with this. If you like, you can just generate a verse brief release note and I can edit it later (if you haven't used `changeset` before it is extremely cool and really helps our releases)

packages/openmrs/src/Adaptor.js
josephjclark: Looking at this now, the request helper should be throwing errors for any 400+ requests, which will bubble up and kill the job. So I think we're looking good right now!

packages/openmrs/src/Utils.js
```diff
@@ -1,44 +1,36 @@
...
+
+    parseAs: 'json',
+  };
 
-const isArray = variable => !!variable && variable.constructor === Array;
+  const url = `${instanceUrl}${path}`;
 
-export function nestArray(data, key) {
-  return isArray(data) ? { [key]: data } : data;
+  return commonRequest(method, url, options);
```
josephjclark, line 35: One more thought here.

If you take a look at the HTTP adaptor, it calls `logResponse` from `common` so that every single http request will log the URL and status. 

Working from something Taylor said, maybe we can copy that code out and `console.debug` with similar output? Logging it to debug de-emphasises it (users will be able to hide debug logs one day).

packages/openmrs/src/Adaptor.js
```diff
@@ -162,101 +150,86 @@ export function get(path, query, callback = false) {
...
+
+    return prepareNextState(state, response, callback);
   };
 }
 
 /**
  * Fetch all non-retired persons that match any specified parameters
  * @example
- * searchPerson({ q: Sarah })
+ * searchPerson({ q: Hunter, limit: 3 })
```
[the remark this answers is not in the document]
hunterachieng: @josephjclark  from your comment, does this mean I should remove the limit?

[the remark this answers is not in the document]
josephjclark: Yah let's remove the limit, thanks. I appreciate the idea but it exposes quite a lot of complicated :sweat_smile: 

Let's put the toothpaste back in the tube for now. We'll fix the docs later!

packages/openmrs/src/Adaptor.js
```diff
@@ -162,101 +150,86 @@ export function get(path, query, callback = false) {
...
+      'GET',
+      '/ws/rest/v1/person',
+      {},
+      resolvedQuery
+    );
+
+    if (response.body.results.length == 0) {
+      throw `Search operation failed to find person with query string: ${resolvedQuery?.q}`;
+    } else {
+      console.log(`Found person with query string: ${resolvedQuery?.q}.`);
```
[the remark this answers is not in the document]
josephjclark: Maybe just ``console.log(`Found ${response.body.results.length} people`)``;

packages/openmrs/src/Adaptor.js
josephjclark: Well technically that's a change in behaviour so it's scope creep :thinking: 

Looking at the code before and after though, I think you're right. We used to say `found an encounter` (which wouldn't be true if `results.length === 0`).

So yes let's say `Found ${response.body.results.length}} results` - that's an acceptable change in the spirit of this PR.

packages/openmrs/src/Adaptor.js
```diff
@@ -59,23 +36,24 @@ async function login(state) {
...
+    const response = await request(
+      state,
+      'GET',
+      `/ws/rest/v1/patient/${resolvedUuid}`
+    );
+
+    console.log(response);
+
+    if (response.statusCode === 200) {
+      console.log(`Retrieved patient with uuid: ${resolvedUuid}...`);
```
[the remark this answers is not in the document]
josephjclark: I think this is probably fine and we won't hear back from AK for a few days at least. Let's just assume this is OK to do.

packages/openmrs/src/Adaptor.js
```diff
@@ -162,101 +150,86 @@ export function get(path, query, callback = false) {
...
-      .post(urlPath)
-      .type('json')
-      .send(resolvedData)
-      .then(response => handleResponse(response, state, callback))
-      .catch(handleError);
+export function post(path, data, callback = s => s) {
+  return async state => {
+    const [resolvedPath, resolvedData] = expandReferences(state, path, data);
+
+    console.log(`Sending an HTTP post to ${resolvedPath}`);
```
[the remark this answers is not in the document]
hunterachieng: @josephjclark  Should I display the whole URL?

packages/openmrs/src/Utils.js
josephjclark: Of course - I'll get back to you a bit later

packages/openmrs/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Ooh good question. Probably just the path but maybe not 🤔

Pause on this bit for now, because it's affected by my other comment  (the one I'll get back to you for)

josephjclark:
Ok @hunterachieng this looks great - just one more suggestion from me (expanding on the remaining comments).

I would like the adaptor to log _every_ HTTP request with the method, url and status.

We will use whatever URL we get back from the response object - I think it'll just be the path, but whatever. We'll put it out and wait for feedback if we want to change it.

Here's how we'll do it:
* Take a look at  the [HTTP adaptor's request helper](**url**)
* After it calls `commonRequest`, it calls this thing called `logResponse`
* Here is the [logResponse](https://github.com/OpenFn/adaptors/blob/f66b8ba574ef717fb2cecc2f9c17bfd2a2be3c23/packages/common/src/util/http.js#L13) code in common. It basically just does ``console.log(`${method} ${url} - ${statusCode} in ${duration}ms`)``
* I think you should just call `logResponse` in your request helper in util. You need to do something like:
```
return commonRequest(method, url, options).then(logResponse);
```
* Also, I think you should go into `common.logResponse` and make it return the `response` object, so that its easy to do this sort of chaining in other adaptors. Otherwise the example code I've given you will break!

I hope that's clear? We can talk about it if not. This replaces all the other basic HTTP logging you've added in this adaptor - it gives us better logging at the cost of less code.

hunterachieng:
@josephjclark  I have added the `logResponse` but I do not get the statement where I am to return the response object in ```common.logResponse```

hunterachieng:
Screenshot 2024-04-03 at 12 22 32
Screenshot 2024-04-03 at 12 21 23

hunterachieng:
@josephjclark  I have added the logResponse

josephjclark (approved):

josephjclark:
This is almost ready to go - I'm just investing something on the resulting state from `searchPatient`. The result data is double-buffered on the state object, resulting in needlessly large dataclips.

I'm going to take a quick look at it. If it looks hairy I'll raise an issue for later.

taylordowns2000 (approved):

Issue #394, opened 2023-09-18
 `salesforce` Expose request function from `state.connection.request`


mtuchi:
### Summary 
Expose `state.connection.request(request, options,callback)` function in salesforce adaptor 

### References
- the REQUEST options https://github.com/jsforce/jsforce/blob/master/lib/connection.js#L309-L324
- the httpApi options https://github.com/jsforce/jsforce/blob/master/lib/http-api.js#L8-L19

**Sample code**
```
**
 * Send REST API request with given HTTP request info, with connected session information.
 *
 * Endpoint URL can be absolute URL ('https://na1.salesforce.com/services/data/v52.0/sobjects/Account/describe')
 * , relative path from root ('/services/data/v52.0/sobjects/Account/describe')
 * , or relative path from version root ('/sobjects/Account/describe').
 *
 * @param {String|Object} request - HTTP request object or URL to GET request
 * @param {String} request.method - HTTP method URL to send HTTP request
 * @param {String} request.url - URL to send HTTP request
 * @param {Object} [request.headers] - HTTP request headers in hash object (key-value)
 * @param {Object} [options] - HTTP API request options
 * @param {String} [options.responseType] - Overriding content mime-type in response
 * @param {Transport} [options.transport] - Transport for http api
 * @param {Object} [options.noContentResponse] - Alternative response when no content returned in response (= HTTP 204)
 * @param {Callback.<Object>} [callback] - Callback function
 * @returns {Promise.<Object>}
 */
export function request(request, options, callback) {
  return async state => {
    const [resolvedRequest, resolvedOptions] = newExpandReferences(
      state,
      request,
      options
    );

    await state.connection.request(resolvedRequest, resolvedOptions, callback);
  };
}
```

mtuchi:
Also this is super cool to have available in salesforce adaptor, Basically adding the ability to query more data by auto fetching next records, It's a wish function for how the `query` function should work 
```
const defaultOpts = {
  autoFetch: false,
  maxFetch: 4000, // TODO @Mtuchi support for maxFetch
};

const queryMore = async (state, qs, opts, cb) => {
  let done = false;
  let qResult = null;
  const { connection } = state;
  const { autoFetch } = { ...defaultOpts, ...opts };

  console.log(`Executing query: ${qs}`);
  qResult = await connection.query(qs);

  if (qResult.totalSize > 0) {
    console.log('Total records', qResult.totalSize);

    while (!done) {
      state = {
        ...state,
[...]

Issue #447, opened 2023-12-08
Improve `parseXML` to use `expandRefenrece`


mtuchi:
### Summary
Modify `parseXML` arguments to use `expandReference` 

### Todo's
- [x] Use `expandReference` on `parseXML`  arguments. Done https://github.com/OpenFn/adaptors/pull/448
- [ ] Add example of traversing the DOM

aleksa-krolls:
@mtuchi how do you think this should be prioritized? Is there still adaptor work to be done here before we can work on development for a new Water Aid workflow?

mtuchi:
@aleksa-krolls the important part of this issue is already done in https://github.com/OpenFn/adaptors/pull/448. And we already cut a new release for http adaptor `v5.1.1` which unblock the testing for water-aid. The remaining task `Add example of traversing the DOM` i would say it's low priority. I think it can be part of the new http adaptor once when we cut a release for the new http adaptor.

Issue #477, opened 2024-02-23
Mailgun: Test is failing


josephjclark:
We've randomly got a failing test on Mailgun, which has been skipped in #475

The problem is that the unit test uses a live credential to contact mailgun, and I'm pretty sure the token has expired or something.

I don't really want to keep the live test running, certainly not in CI. So we need to work out what to do here

PR #478, opened 2024-02-27
`Http` & `Salesforce` improvements

mtuchi:
## Summary

Add callback support for `parseXML` in `http` and add `options` and `callback` support for `query` in `salesforce`

## Details

Callback is very useful especially when it's used in `each()` function. The callback support in `parseXML` will make it easier to pluck out data from `state.data.body`. See example below
```
each(
  "$.response[*]",
  parseXML(
    (s) => s.data,
    ($) => $,
    (next) => ({ ...next, results: next.data.body })
  )
);
```

For `salesforce` i made some improvement for the query function to use `query(qs, options, callback)`. This will allow me to specify options like `{ autoFetch: true }` which will make it easy to query more records using only the `query()` instead of `bulkQuery()`
Eg: 
```
query('SELECT Id, Name from Account, { autoFetch: true });
```
With the given example above ⬆️ , If the results is more `2000` records. Then the function will auto-fetch the remaining records until we have all records 

Closes https://github.com/OpenFn/adaptors/issues/394, https://github.com/OpenFn/adaptors/issues/447
## Review Checklist

### DIFF
--- a/.changeset/fresh-zoos-type.md
+++ b/.changeset/fresh-zoos-type.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-salesforce': minor
+---
+
+Add options and callback params in query function
--- a/.changeset/fresh-zoos-type.md
+++ b/.changeset/fresh-zoos-type.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-salesforce': minor
----
-
-Add options and callback params in query function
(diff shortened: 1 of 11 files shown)


mtuchi:
The failing tests are related to mailgun, See https://github.com/OpenFn/adaptors/issues/477 for more information

josephjclark (approved):

josephjclark:
@mtuchi this looks great. Can you please rebase (or merge) against main and run `pnpm changeset version` to push version changes? Then we can merge this in.

josephjclark (approved):

PR #482, opened 2024-03-13
Fancy mocks experiements

josephjclark:
Experimenting with:

 * Alternative adaptor stucture
 * Mockable functions for unit and integration tests
 * job code as unit tests and example documentation

See  `tutorial.md` for a guide to the changes.


josephjclark:
One possible benefit of the operation/impl split is that you could call the impl directly from job code, allowing you to do nested operations more easily. Although even as I type this it feels like bad practice. Would it make [women-for-women](https://github.com/OpenFn/women-for-women/blob/master/jobs/committed/upsertCustomCCV2.js) any cleaner?

PR #480, opened 2024-03-06
feat: Implement ability to generate template with the cli

hunterachieng:
Generate a new adaptor template with `pnpm generate <name>`
## Issues

Closes #19

## Review Checklist

Before merging, the reviewer should check the following items:

- [ ] Does the PR do what it claims to do?
- [ ] If this is a new adaptor, has the migration tool been run and the
      migration guide followed?
- [ ] Are there any unit tests? Should there be?
- [ ] Is there a changeset associated with this PR? Should there be? Note that
      dev only changes don't need a changeset.

PR #487, opened 2024-03-19
`salesforce` close bulk job when finish

mtuchi:
## Details

It turned out that `job.close()` is a promise function and we were calling it without `await`. This led to bulk operation stuck in progress with the status of `Open`. This PR also remove `@openfn/language-template` in `.changeset/config.json`. `@openfn/language-template` is removed on #480 

## Issues

Reference #407 

## Review Checklist


.tool-versions
```diff
@@ -1 +1,2 @@
 nodejs 18.17.1
+pnpm 8.6.12
```
josephjclark, line 2: Does this work? I had trouble getting the pnpm plugin working.

I'll check out the branch and test it out.

josephjclark: @mtuchi can we take this out of here and open it up in a new PR?

Looking at the docs we need some better install instructions (ie, add links to the node and pnpm asdf plugins)

It's also a big-ish change and I'd like it merged as a clean PR for the sake of auditing.

mtuchi: Certainly

josephjclark (approved):

josephjclark:
Great @mtuchi - let's bump versions and add release notes, then we can merge!

packages/salesforce/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @openfn/language-salesforce
 
+## 4.5.2
+
+### Patch Changes
+
+- 2006e88: await `job.close()`
```
josephjclark, line 7: Maybe we can say "fix an issue with bulk jobs not closing"

Also the PR needs to come out of draft so I can merge.

Thanks!

PR #491, opened 2024-03-20
Cursor Helper

### LINKED ISSUE #486: common: create a cursor function / cron job helper
Create a helper function in common to make it easier to support custom cursors in cron jobs.

I think this is mostly fixed by using input state properly, but a helper function can do stuff like standardises state keys and convert natural language strings like `today` into dates at run-time.

TODO: discuss design with Mutchi

josephjclark:
This PR adds a `cursor()` function designed to make the timestamp and pagination cursor pattern easier.

Closes #486

This approach leans heavily on input state and works best in v2.

See the docs PR at https://github.com/OpenFn/docs/pull/465

## General approach

Unlike v1, Lightning (and the CLI) give us good control of the input state. We need to use that.

So, if you want to run a job with the cursor in a specific place, just set the input to `{ cursor: '<date>' }` and that will be respected by the job.

To make date management easier, you pass human friendly strings `now`, `today`, `yesterday`, `n hours ago`, `n days ago`, and `start` (the time at which the run started). Note that these will parse relative to system time - which in lightning's case probably means UTC.

## Usage

Use the `cursor` function from common to set a cursor value on state.

```
cursor($.cursor, { defaultValue:  'today' })
```

This pattern will allow you to use the value on state if provided, but if no cursor is passed will default to "today".

You can update the cursor within the job by setting `state.cursor`. Or, if you want to use a natural language timestamp, use the `cursor` operation again.

```
cursor($.cursor, { defaultValue:  'today' })
fn(/* do something */)
cursor('now')
```

The value that gets written to `state.cursor` will be an actual ISO date, so when the next job runs, the cursor will be set to this start time.

Note that adaptors will need to be updated to export `cursor` from common.

## Options

The second argument accepts options:

* `options.defaultValue` is the value that will be used if the value argument is falsy
* `options.key` lets you change the state key used (ie, use `page` instead of `cursor`)

Later we might introduce a timezone and a flag to disable natural language date parsing.

## Usage examples

Basic usage: a cursor which starts from the start of the current day
```js
cursor('today')
fn(() => { /* do the thing using state.cursor */ })
```

Typical usage: use the input state if provided, or else default to 'now'. At the end of the job, set the cursor to the start time, ready for the next run.
```js
cursor($.cursor, { defaultValue: 'now' })
fn(() => { /* do the thing, use and set state.cursor */ })
cursor('start')
```

Pagination cursor which defaults to page 0
```js
cursor($.cursor, { defaultValue: 0, key: 'page' })
fn(() => { /* do the thing, use and set state.page */ })
```

### DIFF
--- a/.changeset/odd-apples-wait.md
+++ b/.changeset/odd-apples-wait.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-common': minor
+---
+
+Added cursor() helper
--- a/packages/common/ast.json
+++ b/packages/common/ast.json
@@ -1313,6 +1313,83 @@
       "valid": true
     },
     {
+      "name": "cursor",
+      "params": [
+        "value",
+        "options"
+      ],
+      "docs": {
+        "description": "Sets a cursor property on state.\nSupports natural language dates like `now`, `today`, `yesterday`, `n hours ago`, `n days ago`, and `start`,\nwhich will be converted relative to the environment (ie, the Lightning or CLI locale). Custom timezones \nare not yet supported.\nSee the usage guide at @{link https://docs.openfn.org/documentation/jobs/job-writing-guide#using-cursors}",
+        "tags": [
+          {
+            "title": "public",
+            "description": null,
+            "type": null
+          },
+          {
+            "title": "example",
+            "description": "cursor($.cursor, { defaultValue: 'today' })",
+            "caption": "Use a cursor from state if present, or else use the default value"
+          },
+          {
+            "title": "example",
+            "description": "cursor(22)",
+            "caption": "Use a pagination cursor"
+          },
+          {
+            "title": "function",
+            "description": null,
+            "name": null
+          },
+          {
+            "title": "param",
+            "description": "the cursor value. Usually an ISO date, natural language date, or page number",
+            "type": {
+              "type": "NameExpression",
+              "name": "any"
+            },
+            "name": "value"
+          },
+          {
+            "title": "param",
+            "description": "options to control the cursor.",
+            "type": {
+              "type": "NameExpression",
+              "name": "object"
+            },
+            "name": "options"
+          },
+          {
+            "title": "param",
+            "description": "set the cursor key. Will persist through the whole run.",
+            "type": {
+              "type": "NameExpression",
+              "name": "string"
+            },
+            "name": "options.key"
+          },
+          {
+            "title": "param",
+            "description": "the value to use if value is falsy",
+            "type": {
+              "type": "NameExpression",
+              "name": "any"
+            },
+            "name": "options.defaultValue"
+          },
+          {
+            "title": "returns",
+            "description": null,
+            "type": {
+              "type": "NameExpression",
+              "name": "Operation"
+            }
+          }
+        ]
+      },
+      "valid": false
+    },
+    {
       "name": "map",
       "params": [
         "path",
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -827,7 +827,7 @@ export function cursor(value, options = {}) {
       }
     }
     state[cursorKey] = cursor;
-    console.log(`Setting cursor to ${cursor}`);
+    console.log('Setting cursor to:', cursor);
 
     return state;
   }
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -778,26 +778,19 @@ export function validate(schema = 'schema', data = 'data') {
   };
 }
 
-// Note that we don't write cursorStart to state,
-// because it should not affect the next job
-
 let cursorStart = undefined;
 let cursorKey = 'cursor';
 
-// TODO timezones
-// All times should use UTC internally. But you can have strings parsed in a different time zone
-// Also,when we log a cursor time, we should log it in the timezone provided
 /**
- * Sets a cursor value on state (writes to `state.cursor`, or whatever value you set on options.key).
- * The first time this is called, a `cursorStart` key will be set on state,
- * which you can use to set the final state.cursor ready for the next run with `cursor('start')`
+ * Sets a cursor property on state.
  * Supports natural language dates like `now`, `today`, `yesterday`, `n hours ago`, `n days ago`, and `start`,
  * which will be converted relative to the environment (ie, the Lightning or CLI locale). Custom timezones 
  * are not yet supported.
+ * See the usage guide at @{link https://docs.openfn.org/documentation/jobs/job-writing-guide#using-cursors}
  * @public
- * @example <caption>use a cursor from state if present, or else use the default value</caption>
+ * @example <caption>Use a cursor from state if present, or else use the default value</caption>
  * cursor($.cursor, { defaultValue: 'today' })
- * @example <caption>cursor for pagination</caption>
+ * @example <caption>Use a pagination cursor</caption>
  * cursor(22)
  * @function
  * @param {any} value - the cursor value. Usually an ISO date, natural language date, or page number
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -783,32 +783,47 @@ export function validate(schema = 'schema', data = 'data') {
 
 let cursorStart = undefined;
 
+// TODO timezones
+// All times should use UTC internally. But you can have strings parsed in a different time zone
+// Also,when we log a cursor time, we should log it in the timezone provided
 /**
- * Sets a cursor value on state (writes to `state.cursor`).
+ * Sets a cursor value on state (writes to `state.cursor`, or whatever value you set on options.key).
  * The first time this is called, a `cursorStart` key will be set on state,
- * which you can use to set the final state.cursor ready for the next run.
+ * which you can use to set the final state.cursor ready for the next run with `cursor('start')`
  * Supports natural language dates like `now`, `today`, `yesterday`, `n hours ago`, `n days ago`, and `start`
  * @public
- * @example
- * cursor((state) => state.cursor || 'today')
- * @example
+ * @example <caption>use a cursor from state if present, or else use the default value</caption>
+ * cursor($.cursor, { defaultValue: 'today' })
+ * @example <caption>cursor for pagination</caption>
  * cursor(22)
  * @function
  * @param {any} value - the cursor value. Usually an ISO date, natural language date, or page number
+ * @param {any} options - options to control the cursor. Options will persist after the first call.
  * @returns {Operation}
  */
-export function cursor(value) {
+export function cursor(value, options) {
   return (state) => {
+
+    const {
+      defaultValue, // if there is no cursor on state, this will be used
+      key = 'cursor', // the key to use on state. idk why you want to change this, but sure
+      
+      // timezone, // time zone as in UTC, GMT, PST? Or locale?
+    } = options;
+
     if (!cursorStart) {
-      cursorStart = new Date().toISOString()
+      cursorStart = converter(new Date().toISOString())
     }
 
-    if (typeof value === 'string') {
-      state.cursor = parseDate(value, cursorStart)
-      console.log(`Setting cursor "${value}" to ${state.cursor}`)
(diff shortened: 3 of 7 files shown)


josephjclark:
@mtuchi Would you mind taking a look at this, see if it'll work?

I think it can be this simple, I just don't have good access to test cases :(

mtuchi (commented):
@josephjclark `pnpm i & pnpm build` is failing for `common`

<details><summary>I am getting this error </summary>
<p>

```
Building AST

AST build failed! /Users/openfn/Workspace/adaptors/packages/common/ast.json
Error: Command failed: pnpm exec simple-ast  --adaptor /Users/openfn/Workspace/adaptors/packages/common/src/Adaptor.js --output /Users/openfn/Workspace/adaptors/packages/common/ast.json
/Users/openfn/Workspace/adaptors/node_modules/.pnpm/@openfn+simple-ast@0.4.1/node_modules/@openfn/simple-ast/lib/generate.js:209
    if (item.declarations[0].init.callee) {
                                  ^

TypeError: Cannot read properties of null (reading 'callee')
    at /Users/openfn/Workspace/adaptors/node_modules/.pnpm/@openfn+simple-ast@0.4.1/node_modules/@openfn/simple-ast/lib/generate.js:209:35
    at Array.filter (<anonymous>)
    at Object.<anonymous> (/Users/openfn/Workspace/adaptors/node_modules/.pnpm/@openfn+simple-ast@0.4.1/node_modules/@openfn/simple-ast/lib/generate.js:208:4)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.17.1

    at __node_internal_genericNodeError (node:internal/errors:865:15)
    at ChildProcess.exithandler (node:child_process:419:12)
    at ChildProcess.emit (node:events:514:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1091:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:514:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:323:12) {
  code: 1,
  killed: false,
  signal: null,
  cmd: 'pnpm exec simple-ast  --adaptor /Users/openfn/Workspace/adaptors/packages/common/src/Adaptor.js --output /Users/openfn/Workspace/adaptors/packages/common/ast.json'
}
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

Node.js v18.17.1
```

</p>
</details>

josephjclark:
@mtuchi Ack, sorry, silly bug - try again!

mtuchi:
Hiya @josephjclark here is my feedback 

#### 1st. You need to expandReference on cursor value
`cursor(state=> state.cursor || 'start');` Does not work because the `value` does not use expandReferences.

**My quick fix was** 
```
 return state => {
    const [resolvedValue] = newExpandReferences(state, value);
    if (!cursorStart) {
      cursorStart = new Date().toISOString();
    }

    if (typeof resolvedValue === 'string') {
      state.cursor = parseDate(resolvedValue, cursorStart);
      console.log(`Setting cursor "${resolvedValue}" to ${state.cursor}`);
    } else {
      state.cursor = resolvedValue;
      console.log(`Setting cursor to ${resolvedValue}`);
    }

    return state;
  };
  ```
  
#### 2nd. In accurate cursor dates for some values
**My test code** 
```
cursor('start');
cursor('now');
cursor('today');
cursor('yesterday');
cursor('1 hour ago');
cursor('2 hours ago');
cursor('3 hours ago');
cursor('72 hours ago');
cursor('1 day ago');
cursor('2 days ago');
cursor('3 days ago');
cursor('end');
```
**CLI logs** 
```
Setting cursor "start" to 2024-04-04T07:54:46.484Z
Setting cursor "now" to 2024-04-04T07:54:46.484Z
Setting cursor "today" to 2024-04-03T21:00:00.000Z
Setting cursor "yesterday" to 2024-04-02T21:00:00.000Z ❌ 
Setting cursor "1 hour ago" to 2024-04-04T06:54:46.485Z
Setting cursor "2 hours ago" to 2024-04-04T05:54:46.485Z
Setting cursor "3 hours ago" to 2024-04-04T04:54:46.485Z
Setting cursor "72 hours ago" to 2024-04-01T07:54:46.485Z
Setting cursor "1 day ago" to 2024-04-03T21:00:00.000Z
Setting cursor "2 days ago" to 2024-04-03T21:00:00.000Z ❌ 
Setting cursor "3 days ago" to 2024-04-03T21:00:00.000Z ❌ 
Setting cursor "end" to 2024-04-04T07:54:46.485Z
```
### 3rd. Timezone ?
This is more of a questions + suggestion 
- What timezone is used ?
- Can i change a timezone ?  or more important one do we want this ability and why ?

>It will be best to log the timezone as well because the timestamp does not match my local time so i am not sure which timezone the cursor is using

**Comment by mtuchi** [comment]
For some reason i am tempted to ask for support for a callback in a cursor function. I think there was a case in AKF workflow were i need to create a cursor at the end of upsert operation then post it to inbox. The triggered workflow was using that cursor to clear some data in salesforce. 

I guess my point here is the cursor that was created after the upsert operation is not related to that workflow. So the final state `state.cursor` should not use that cursor. I should be able to specify something like `state.clearRecordsCursor` when i am setting the cursor for the clear records workflow

**Comment by mtuchi** [comment]
@josephjclark  I think it's useful because it just cut down these lines of code 👇🏽  into one line of code 
```
fn(state => {
  const manualCursor = '2023-01-15';
  const cursor = manualCursor ? new Date(manualCursor) : new Date();

  console.log(cursor, 'Cursor');

  return { ...state, cursor };
});
```
To
```
cursor(state => state.cursor || '2023-01-15');
```
Which can even be something like this 👇🏽 if initial input is `{ "manualCursor": "2023-01-15"}`
```
cursor(state => state.cursor || state.manualCursor);
```

**Comment by mtuchi** [comment]
I just got another light bulk moment 💡, Sometimes i need to set the cursor to todays date or a specific date but i don't need the timestamp. So i usually have this pattern `new Date().toISOString().slice(0, 10);`. This will give me today's date in a format of `YYYY-MM-DD`

I was wondering if the `cursor()` function have an options params `cursor(value, options. callback= s=> s)` This way i can do something like this `cursor('today', {format: 'YYY-MM-DD'})`

**Comment by josephjclark** [comment]
@mtuchi I think you're starting to get it! :bulb: 

I'm adding an options object shortly, so you'll be able to do this (which is super readable)
```
cursor($.cursor, { defaultValue: '2023-01-15' });
```

I really don't think you need to explicitly use a manual cursor key. But sure, if you absolutely need it, that pattern works.

tbh I can't see why you'd need a callback - it's a synchronous function after all. But you can set or clear it at the end of the job:
```
cursor()
```

Instead of `new Date().toISOString().slice(0, 10)`, why woudn;t you just do `setCirsor('today')` or `setCursor('2024-03-12')`? Both should work (all Date.parse() formats should work)

**Comment by josephjclark** [comment]
Regarding timezones, well, you've hurt my brain :exploding_head: 

For now, all timezones will be in UTC, and I think we'll release in that state.

I would like to support this: `cursor('today', { timezone: 'EST' })`, which would set the timestamp to a UTC date representing midnight on the current day in US Eastern time.

But I have two problems with this:
1) I have no idea how to do it (I've been looking at libraries a bit bit it feels like a very hard problem!)
2) I'll bet most databases are in UTC anyway, so maybe a UTC timezone probably makes most sense.

We will need to document that if you `cursor('today')`, we'll take the current time in UTC and go back to the start of that day - which could be a different calendar day if you're in the East

I suppose this is an argument for `cursor('24h')`

**Comment by mtuchi** [comment]
On timezone, i think we should lock it to UTC or (which ever timezone lighting is using). And add in CLI logs the timezone. 
Eg: `Setting cursor "now" to 2024-04-04T12:07:08.775Z UTC Timezone`

**Comment by mtuchi** [comment]
@josephjclark by the way,  not to add more complexity i saw this library that is related to `date-fns` timezone - https://www.npmjs.com/package/date-fns-tz i don't know if it will help or makes things more complicated

**Comment by josephjclark** [comment]
Ok @mtuchi can you do a retest?

- When you set a cursor like "today", the CLI should print the time in your local time zone, so the dates should be accurate now
- There's an options argument now, so you can do `cursor($.state.cursor, { defaultValue: 'today', key: 'startDate' })`

I'm going to update the main text in the PR to reflect the most recent styles. I'll tweak docs too.

**Comment by mtuchi** [comment]
Hiya @josephjclark, Loving the `$.cursor` 🕺🏽. Here is my feedback 

#### 1. `Invalid Date` Error in console.log
I was getting this error `Setting cursor "end" to Invalid Date` when setting cursor like this `cursor($.cursor, { defaultValue: 'today', key: 'lastRunDate' })` 

Updating the console.log to use `state[cursorKey]` instead of `state.cursor` should solve the problem
```
Setting cursor "${cursor}" to ${new Date(state[cursorKey] ).toLocaleString()
```
#### 2. Timestamp/Timezone
The timestamp logs now reflect my local timezone Eg: `Setting cursor "end" to 4/5/2024, 11:31:20 AM`

If it's not too much to ask can we print the timezone as well ? 
Just modifying console to use`.toLocaleString('en-US', { timeZoneName: 'short' })}` which will print `Setting cursor "end" to 4/5/2024, 11:31:20 AM GMT+3`

#### 3. n days ago not working 
Specifying the number of days ago still is not working. It always set the cursor to todays date 
Eg: 
```
cursor('2 days ago');
cursor('22 days ago');

// Will set the cursor to todays 
Setting cursor "2 days ago" to 4/5/2024, 12:00:00 AM 
Setting cursor "22 days ago" to 4/5/2024, 12:00:00 AM 
```

#### 4. expandReference on options 
I think we should expand reference on `options` as well. It will allow me todo something like this 
```
 cursor($.cursor, { defaultValue: $.manualCursor })
```
And i can easily change the initial state. `state.manualCursor: '72 hours ago'`

### 5. Pagination cursor
In most cases pagination cursor will be an object. Eg

```json
{
  "_metadata": 
  {
      "page": 5,
      "per_page": 20,
      "page_count": 20,
      "total_count": 521,
      "Links": [
        {"self": "/products?page=5&per_page=20"},
        {"first": "/products?page=0&per_page=20"},
        {"previous": "/products?page=4&per_page=20"},
        {"next": "/products?page=6&per_page=20"},
        {"last": "/products?page=26&per_page=20"},
      ]
  },
  "records": [
    {
      "id": 1,
      "name": "Widget #1",
      "uri": "/products/1"
    },
  ]
}
```
So i would assume the way i would set the `cursor($.response._metadata)` or `cursor($.response._metadata.next)` [assumes `state.response._metadata.next`]

In that case the `console.log` should also print the JSON object. 
Right now if you set a cursor to an object. Eg: `cursor({next: 22})` It will log `Setting cursor to [object Object]`

Please consider changing the console.log to `console.log('Setting cursor to ', cursor);` So that the log will print `Setting cursor to  { next: 22 }`

mtuchi:
Also don't forget to build to generate the latest `ast.json` file.

josephjclark:
@mtuchi this is amazing feedback - you've caught some brilliant issues there!

It's all been addressed. I've also added a seperate PR for docs - we should merge both at about the same time.

I've still got to do versioning stuff - I'll create a release branch for that

josephjclark:
@mtuchi thanks - I think if we're gonna do that we'll just use date.format, or maybe date-fns.format. But I agree and I'll make that change now!

PR #493, opened 2024-03-22
List common exports in all docs

josephjclark:
## Summary

This PR adds a list of common exports to the docs markdown generated by the build.

image

## Details

The task here is to get the list of exports from common for each adaptor and add them to the page somewhere.

jsdoc basically can't do it. It seems to have no means to list documentation for export aliases from another module. You can document the export statement and jsdoc will respect it. I considered adding build tooling to generate docs for each export.

But why? The result is a long list of funcitons and when looking at the actual docs, you can't tell the main adaptor functions from the common ones. I mean common mostly introduces a load of noise to the documentation - the most valuable docs are the ones native to this adaptor.

So I took a different approach. 
- Using `acorn`, parse the JS files for the adaptor
- Extract all the exported names from common
- List those exported values as hyperlinks at the top of the docs, but don't add any other documentation.

Now the common functionality is visible, but it's visibly a second-class citizen.

## Issues

Closes https://github.com/OpenFn/docs/issues/440

### DIFF
--- a/tools/build/src/commands/docs.ts
+++ b/tools/build/src/commands/docs.ts
@@ -41,7 +41,6 @@ export default async (lang: string) => {
   await fileSet.add([glob])
 
   // Extract exports from common and add them to the template data as externals
-  let exports = [];
   for (const f of fileSet.files) {
     const src = await fs.readFile(f, 'utf8')
     const exports = extractExports(src).map((e) => ({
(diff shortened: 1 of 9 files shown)


tools/build/src/commands/docs.ts
```diff
@@ -1,22 +1,25 @@
-import jsdoc2md from 'jsdoc-to-markdown';
-import fs from 'node:fs/promises';
+import fs, { writeFile, mkdir } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
-import { writeFile, mkdir } from 'node:fs/promises';
+import jsdoc2md from 'jsdoc-to-markdown';
+import FileSet from 'file-set'
```
josephjclark, line 5: I chose fileset because it's the same library used by jsdoc2md, so we know the glob will be treated the same way

taylordowns2000 (approved):
I'm happy with the user side, have not reviewed the code but if it does what's shown in the image then I'm good

aleksa-krolls (approved):

Issue #301, opened 2023-07-10
`googlesheets` improvements


mtuchi:
#### Summary
Currently the `appendValues` log the response which exposes the accessToken in run logs. 

#### Tasks
- Remove the `console.log(response)` when the request is successful, [See L87](https://github.com/OpenFn/adaptors/blob/7034947eded80f6dc69681be3ec61a7dd38dbcd8/packages/googlesheets/src/Adaptor.js#L87)
- Add support for `callback()` in `appendValues()`
- Add `readValues()` function that takes `params` and `callback`

#### Lighting Run Logs Screenshot 
![Capture-2023-07-10-091927](https://github.com/OpenFn/adaptors/assets/6592749/ca8e2f38-a80f-45fd-bf6d-5934e65796fa)

mtuchi:
Related to #497

mtuchi:
Fixed on #495

PR #495, opened 2024-03-26
`Asana` & `GoogleSheets` improvements

### LINKED ISSUE #497: googlesheets: add getValues()
New a new function to get values for a specific range from a target Google Sheet. Needed to implement the workflow for [ConSoSci #207](https://github.com/OpenFn/ConSoSci/issues/207).

mtuchi:
## Summary
Improvements for Asana and GoogleSheet adaptor driven from https://github.com/OpenFn/ConSoSci/issues/204 and https://github.com/OpenFn/ConSoSci/issues/206

#### Asana
- add `createTaskStory()` func
- use new http from common/util 

#### GoogleSheets
- add `getValues()` function
- add `batchUpdateValues()` function
- use `expandReference()` from common/util
- improve the code to use a google client before the begin of operations
- add `logError` function for in-depth error logs 

## Issues 

Closes #497,  #301 
## Review Checklist

### DIFF
--- a/packages/asana/src/Adaptor.js
+++ b/packages/asana/src/Adaptor.js
@@ -200,25 +200,24 @@ export function upsertTask(projectGid, params, callback) {
 /**
  * Options provided to the createTaskStory request
  * @typedef {Object} StoryOptions
- * @property {object} story - The story to create, It represents an activity associated with an object in the Asana system.
- * @property {string} [story.text] - The plain text of the comment to add. Cannot be used with html_text.
- * @property {string} [story.html_text] - Opt In. HTML formatted text for a comment. This will not include the name of the creator.
- * @property {boolean} [story.is_pinned] - Conditional. Whether the story should be pinned on the resource.
- * @property {string} [story.sticker_naem] - The name of the sticker in this story. `null` if there is no sticker.
- * @property {array} [opt_fields] - Opt In. This endpoint returns a compact resource, which excludes some properties by default. To include those optional properties, set this query parameter to a comma-separated list of the properties you wish to include.
- * @property {boolean} [opt_pretty] - Opt In. Provides the response in a “pretty” format. In the case of JSON this means doing proper line breaking and indentation to make it readable. This will take extra time and increase the response size so it is advisable only to use this during debugging.
+ * @property {string} text - The plain text of the comment to add. Cannot be used with html_text.
+ * @property {string} html_text - Opt In. HTML formatted text for a comment. This will not include the name of the creator.
+ * @property {boolean} is_pinned - Default to `false`. Whether the story should be pinned on the resource.
+ * @property {string} sticker_name - The name of the sticker in this story. `null` if there is no sticker.
+ * @property {array} opt_fields - Opt In. This endpoint returns a compact resource, which excludes some properties by default. To include those optional properties, set this query parameter to a comma-separated list of the properties you wish to include.
+ * @property {boolean} opt_pretty - Defaults to `false`. Provides the response in a “pretty” format. In the case of JSON this means doing proper line breaking and indentation to make it readable. This will take extra time and increase the response size so it is advisable only to use this during debugging.
  */
 
 /**
  * Create a story to a specific task.
  * @public
- * @example
+ * @example <caption>Create a plain text comment</caption>
  * createTaskStory("1206933955023739", {
- *   story: {
- *     text: "This is a comment",
- *     html_text: "<body>This is a comment</body>",
- *     is_pinned: false,
- *   },
+ *   text: "This is a comment",
+ * });
+ * @example <caption>Create a HTML formatted text comment</caption>
+ * createTaskStory("1206933955023739", {
+ *   html_text: "<body>This is a comment</body>",
  * });
  * @function
  * @param {string} taskGid - Globally unique identifier for the task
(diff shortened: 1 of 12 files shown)


packages/asana/src/Adaptor.js
```diff
@@ -1,9 +1,6 @@
-import {
-  execute as commonExecute,
-  composeNextState,
-  expandReferences,
-  http,
-} from '@openfn/language-common';
+import { execute as commonExecute } from '@openfn/language-common';
+import { expandReferences } from '@openfn/language-common/util';
+import { request as commonRequest } from './Utils';
```
josephjclark, line 3: This shouldn't be called `commonRequest` - it's very confusing! Because it's NOT common.request.

Maybe call it `requestHelper`?

packages/asana/src/Utils.js
```diff
@@ -0,0 +1,44 @@
...
+    body,
+  };
+
+  return commonRequest(method, path, options)
+    .then(response => {
+      logResponse(response);
+
+      return {
+        ...composeNextState(state, response.body?.data),
+        response,
```
josephjclark, line 34: Something I've just done in `openmrs` is to remove the `body` from the response object,  because otherwise it just spams up state (the response data will be duplicated on `state.response` and `state.data`). 

Admittedly it's a drop in the ocean compared to references, but it does bother me a bit. 

Should we do the same here? 

See here: https://github.com/OpenFn/adaptors/blob/904e434cbb5f3e388715a2cbac2564700f732ea2/packages/openmrs/src/Utils.js#L9

packages/asana/src/Adaptor.js
```diff
@@ -233,39 +168,122 @@ export function createTask(params, callback) {
...
-    return http
-      .get(config)(state)
-      .then(response => {
-        const matchingTask = response.data.data.find(
+    const [resolvedProjectGid, resolvedParams] = expandReferences(
+      state,
+      projectGid,
+      params
+    );
+    const { externalId, data } = resolvedParams;
```
josephjclark, line 176: Little tip - you should be able to deconstruct these in-line in your expand statement:
```
 const [resolvedProjectGid, { externalId, data }] = expandReferences(
      state,
      projectGid,
      params
    );
```

packages/asana/src/Adaptor.js
```diff
@@ -233,39 +168,122 @@ export function createTask(params, callback) {
...
           return createTask(data, callback)(state);
         }
-      });
+      }
+    );
+  };
+}
+
+/**
+ * Options provided to the createTaskStory request
```
josephjclark, line 201: Great docs!

packages/asana/src/Adaptor.js
```diff
@@ -233,39 +168,122 @@ export function createTask(params, callback) {
...
+ * @typedef {Object} StoryOptions
+ * @property {object} story - The story to create, It represents an activity associated with an object in the Asana system.
+ * @property {string} [story.text] - The plain text of the comment to add. Cannot be used with html_text.
```
josephjclark, line 204: I don't think you need the object name in `[story.text]` though. Look at HTTP, where we just list the property name.

I also don't think we should put all these in square brackets. In an object like this I'd assume options are optional unless otherwise stated. Also, nowhere in the docs do we really explain that `[x]` means "optional", so it's not a terrible useful notation.

packages/asana/src/Adaptor.js
```diff
@@ -233,39 +168,122 @@ export function createTask(params, callback) {
...
+    );
+  };
+}
+
+/**
+ * Options provided to the Asana API request
+ * @typedef {Object} RequestOptions
+ * @property {object} body - Body data to append to the request.
+ * @property {object} query - An object of query parameters to be encoded into the URL.
+ * @property {string} method - The HTTP method to use
```
josephjclark, line 256: We should say this defaults to get

packages/googlesheets/src/Adaptor.js
```diff
@@ -79,20 +112,113 @@ export function appendValues(params) {
...
+            );
           }
         }
       );
     });
   };
 }
 
+/**
+ * Batch Updates values in a Spreadsheet.
```
josephjclark, line 134: Should read `Batch update values in a Spreadsheet`

packages/googlesheets/src/Adaptor.js
```diff
@@ -79,20 +112,113 @@ export function appendValues(params) {
...
+
+    try {
+      const response = await client.spreadsheets.values.get({
+        spreadsheetId: resolvedSheetId,
+        range: resolvedRange,
+      });
+      const numRows = response?.data?.values.length ?? 0;
+      console.log(`${numRows} rows retrieved.`);
+
+      const nextState = { ...composeNextState(state, response.data), response };
```
josephjclark, line 212: I also wonder if large data arrays are double-buffered here in `state.data` and `state.response`.

This might be a difficult situation I've stumbled into :grimacing:

josephjclark (commented):
Couple of suggestions!

packages/asana/src/Adaptor.js
mtuchi: Thanks, this is good 👍🏽

packages/asana/src/Adaptor.js
mtuchi: Thanks

packages/asana/src/Adaptor.js
mtuchi: Update the method to `@property {string} method - The HTTP method to use. Defaults to 'GET'`

packages/googlesheets/src/Adaptor.js
mtuchi: Do you mean duplicates ? or what do you mean by double-buffered ? @josephjclark

packages/asana/src/Adaptor.js
mtuchi: I have made some improvements and remove `params.story` in `createTaskStory()`.
I have grouped all params for story in a variable called story like this `const story = { text, html_text, is_pinned, sticker_name };`.

To create a comment using text 
`createTaskStory("128320", {text: "my comment"})`


To create a comment using html format text  
`createTaskStory("128320", {text: "<body>my comment</body>"})`

josephjclark: Looks good to me!

josephjclark (approved):

PR #504, opened 2024-04-05
Release cursor

josephjclark:
Release branch for the new cursor function

Must be released in tandem with the docs at  https://github.com/OpenFn/docs/pull/465

This branch will include minor releases for:
- common
- msgraph
- asana
- kobotoolbox
- googlesheet

There are patch releases for basically every other adaptor, since they depend on common. These are not terribly helpful but it's quite hard to ignore them - something I think changesets doesn't handle terribly well.

### DIFF
--- a/.changeset/odd-apples-wait.md
+++ b/.changeset/odd-apples-wait.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-common': minor
+---
+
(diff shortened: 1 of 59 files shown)


josephjclark:
This should be ready for release now - we just merge to main and we done!

Issue #481, opened 2024-03-12
Salesforce `SOQL` limits docs


mtuchi:
### Description
They query functions has couple of limits but to name a few important limits 
- 2,000 results per request (API version 28.0 and later),
- 4,000 characters for each string within a WHERE clause.

### Request
Update the documentation for `query()` function to include the limits highlighted above, Learn more about [SOQL Limits here](https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_soslsoql.htm)

aleksa-krolls:
@mtuchi we can close this now, yes? Please do, if yes

mtuchi:
@aleksa-krolls i working on joe's feedback on this pr #501

christad92:
@mtuchi can this be closed now?

mtuchi:
@christad92 yes closing this as it was addressed in #506

PR #506, opened 2024-04-08
`salesforce` API limits

### LINKED ISSUE #501: salesforce: add docs on limits
- Consider updating `Common Errors` [Overview page](https://docs.openfn.org/adaptors/salesforce#common-errors) to add errors that might come up related to limits
- For `query()`, `bulk()`, and `bulkQuery()` --> link to SF documentation on these limits (would be helpful for users to know the specific APIs being used under the hood)

mtuchi:
## Summary

Add `jsdoc` for search query API limits and improved `bulk()` and `bulkQuery()` jsdocs 

## Details

Using `@throws` to describe deferent API limits

## Issues 
Closes #501 #481 

## Review Checklist

### DIFF
--- a/.changeset/honest-glasses-drum.md
+++ b/.changeset/honest-glasses-drum.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-salesforce': patch
+---
+
+- add `query()` limits
+- improve `bulk()` and `bulkQuery()` function
--- a/.changeset/honest-glasses-drum.md
+++ b/.changeset/honest-glasses-drum.md
@@ -2,5 +2,4 @@
 '@openfn/language-salesforce': patch
 ---
 
-- add `query()` limits
-- improve `bulk()` and `bulkQuery()` function
+- Add documentation about Salesforce API limits to query and bulkQuery
--- a/.changeset/honest-glasses-drum.md
+++ b/.changeset/honest-glasses-drum.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-salesforce': patch
----
-
-- Add documentation about Salesforce API limits to query and bulkQuery
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -282,6 +282,10 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
   }
 }
 
+const defaultOptions = {
+  pollTimeout: 90000, // in ms
+  pollInterval: 3000, // in ms
+};
 /**
  * Execute an SOQL Bulk Query.
  * This function uses bulk query to efficiently query large data sets and reduce the number of API requests.
@@ -308,9 +312,17 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
 export function bulkQuery(qs, options, callback) {
   return async state => {
     const { connection } = state;
-    const [resolvedQs, { pollTimeout = 90000, pollInterval = 3000 }] =
-      newExpandReferences(state, qs, options);
+    const [resolvedQs, resolvedOptions] = newExpandReferences(
+      state,
+      qs,
+      options
+    );
     const apiVersion = connection.version;
+
+    const { pollTimeout, pollInterval } = {
+      ...defaultOptions,
+      ...resolvedOptions,
+    };
 
     console.log(`Executing query: ${resolvedQs}`);
 
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -285,7 +285,8 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
 /**
  * Execute an SOQL Bulk Query.
  * This function uses bulk query to efficiently query large data sets and reduce the number of API requests.
- * The `bulkQuery()` uses {@link https://sforce.co/4b6kn6z Bulk API v.2.0} which is available in API version 41.0 and later.
+ * `bulkQuery()` uses {@link https://sforce.co/3y9phlc Bulk API v.2.0} which is available in API version 41.0 and later.
+ * This API is subject to {@link https://sforce.co/4b6kn6z rate limits}.
  * @public
  * @example
  * <caption>The results will be available on `state.data`</caption>
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -148,16 +148,23 @@ export function retrieve(sObject, id, callback) {
  * Execute an SOQL query.
  * Note that in an event of a query error,
  * error logs will be printed but the operation will not throw the error.
+ *
+ * This function makes an API call to retrieve query results.
+ * It is subject to rate limits imposed by the Salesforce API.
+ * `Maximum rows returned in SOQL query results`: </br>
+ * This function will return `2,000` results per request (API v28.0+),
+ * `200` results (Previous API versions)
  * @public
  * @example
  * query(state=> `SELECT Id FROM Patient__c WHERE Health_ID__c = '${state.data.field1}'`);
  * @example <caption>Query more records if next records are available</caption>
  * query(state=> `SELECT Id FROM Patient__c WHERE Health_ID__c = '${state.data.field1}'`, { autoFetch: true });
  * @function
- * @param {String} qs - A query string.
+ * @param {String} qs - A query string. Must be less than `4000` characters in WHERE clause
  * @param {Object} options - Options passed to the bulk api.
- * @param {boolean} [options.autoFetch] - Fetch next records if available.
+ * @param {boolean} [options.autoFetch=false] - Fetch next records if available.
  * @param {Function} callback - A callback to execute once the record is retrieved
+ * @throws {Error} Throws an error if characters in SOQL WHERE clauses exceed `4,000` characters
  * @returns {Operation}
  */
 export function query(qs, options, callback = s => s) {
@@ -280,11 +287,6 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
   }
 }
 
-const defaultOptions = {
-  pollTimeout: 90000, // in ms
-  pollInterval: 3000, // in ms
-};
-
 /**
  * Execute an SOQL Bulk Query.
  * This function uses bulk query to efficiently query large data sets and reduce the number of API requests.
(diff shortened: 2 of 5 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -148,16 +148,23 @@ export function retrieve(sObject, id, callback) {
  * Execute an SOQL query.
  * Note that in an event of a query error,
  * error logs will be printed but the operation will not throw the error.
+ *
+ * This function makes an API call to retrieve query results.
```
josephjclark, line 152: Can we simplify to:

>This function is subject to rate limits imposed by the Salesforce API This function will return `2,000` results per request in API versions since 28.0, or 200 results in older versions.

josephjclark: Having said that, I'd rather that we say as little as possible (`this function is subject to rate limits by the Salesforce API`), and then link to an SF doc page. But only if there's a good place to link to.

packages/salesforce/src/Adaptor.js
```diff
@@ -148,16 +148,23 @@ export function retrieve(sObject, id, callback) {
...
+ * This function will return `2,000` results per request (API v28.0+),
+ * `200` results (Previous API versions)
  * @public
  * @example
  * query(state=> `SELECT Id FROM Patient__c WHERE Health_ID__c = '${state.data.field1}'`);
  * @example <caption>Query more records if next records are available</caption>
  * query(state=> `SELECT Id FROM Patient__c WHERE Health_ID__c = '${state.data.field1}'`, { autoFetch: true });
  * @function
- * @param {String} qs - A query string.
+ * @param {String} qs - A query string. Must be less than `4000` characters in WHERE clause
```
josephjclark, line 163: Just the where clause? Not the whole query?

I think this might be buried out of sight here - would a dev ever see it? Maybe there's not much more we can do, let me think on it

josephjclark (commented):
@mtuchi what do the errors look like when we hit these limits? Do we have any easy way to test?

I'm interested in a) what messaging comes  out of Salesforce and b) how well the runtime prints it to the log.

mtuchi:
@josephjclark i can the `query()` character limits, And i have update the error message on docs overview 
see https://github.com/OpenFn/docs/pull/466/files. 

For the `bulkQuery()` we throw an error if we exceed `pollTimeout`

josephjclark:
@mtuchi We simplify our docs here by pointing directly to the salesforce limits page here?.https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_soslsoql.htm

There are actually quite a lot of limits and our docs only touch on a couple. I'd rather we just handed responsibility to salesforce rather than duplicate their own API docs.

We can say something like "The Salesforce query API is subject to rate limits, see <here> for more details" in the JS docs.

I'd still like to see what the CLI does and what the message looks like when a rate limit occurs. The docs are a bit vague but I think the rate limits will return a 503 message? We could consider adding a link to this page in the error message itself

```
console.error("Received a 503 error from salesforce. This is typically associated with rate limits. See <here> for more details")
```

mtuchi:
@josephjclark We only support `SOQL` query when using `query()` function. The Salesforce documentation for rate limits shows limits for both `SOQL and SOSL`. If we want to share the link directly from our jsdoc then we need to explicitly state that. Another thing about the link, It's too long and i feel like it's polluting the jsdoc

As for the error message. If you hit length limits using `query()` you will get 
```
Error Code 414. Response content: <h1>Bad Message 414</h1><pre>reason: URI Too Long</pre>
```

josephjclark:
@mtuchi 

1) I don't agree with either point. We can simply  say "The query() operation is subject to the SOQL limits described here: ".  And I don't really think we should worry a bout polluting the jsdoc? I don't know who that hurts :thinking: We could use a URL shortener if it's really a problem for you.

2) Do we see that error in the CLI?

The thing with all this is that its way more important to report a really clear error than it for us to document the edges of someone else's system.

mtuchi:
Hiya @josephjclark i have updated the docs to include a link 

Regarding the CLI error, this is what i get 👇🏽 
```
Error ERROR_HTTP_414: <h1>Bad Message 414</h1><pre>reason: URI Too Long</pre>
[R/T] ✘ Failed step queryFinancialsSf after 4.244s
[R/T] ✘ AdaptorError: <h1>Bad Message 414</h1><pre>reason: URI Too Long</pre>
[R/T] ✘ Check state.errors.queryFinancialsSf for details.
[CLI] ✔ State written to tmp/output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 50.924s
```

josephjclark:
Thank you @mtuchi, taht output is really helpful. Good enough for now, I think!

packages/salesforce/src/Adaptor.js
```diff
@@ -303,26 +300,19 @@ const defaultOptions = {
  * @function
  * @param {String} qs - A query string.
  * @param {Object} options - Options passed to the bulk api.
- * @param {integer} [options.pollTimeout] - Polling timeout in milliseconds.
- * @param {integer} [options.pollInterval] - Polling interval in milliseconds.
+ * @param {integer} [options.pollTimeout=90000] - Polling timeout in milliseconds.
+ * @param {integer} [options.pollInterva=3000] - Polling interval in milliseconds.
```
josephjclark, line 304: typo on pollInterval

packages/salesforce/src/Adaptor.js
```diff
@@ -303,26 +300,19 @@ const defaultOptions = {
...
  * @param {Function} callback - A callback to execute once the record is retrieved
+ * @throws {Error} Throws an error if the execution exceeds the specified `pollTimeout` duration.
```
josephjclark, line 306: Again this is fine but I bet there are more things to make this throw, so it's misleading

packages/salesforce/src/Adaptor.js
```diff
@@ -148,16 +148,19 @@ export function retrieve(sObject, id, callback) {
...
  * @example <caption>Query more records if next records are available</caption>
  * query(state=> `SELECT Id FROM Patient__c WHERE Health_ID__c = '${state.data.field1}'`, { autoFetch: true });
  * @function
- * @param {String} qs - A query string.
+ * @param {String} qs - A query string. Must be less than `4000` characters in WHERE clause
  * @param {Object} options - Options passed to the bulk api.
- * @param {boolean} [options.autoFetch] - Fetch next records if available.
+ * @param {boolean} [options.autoFetch=false] - Fetch next records if available.
  * @param {Function} callback - A callback to execute once the record is retrieved
+ * @throws {Error} Throws an error if characters in SOQL WHERE clauses exceed `4,000` characters
```
josephjclark, line 163: This is fine but there  are other limits, right? So I'm a bit concerned about calling this one out in particular

For example the SOQL statement in total can only be 100k characters. We're not talking about that here.

I'm not even sure it's accurate to say `throws an error of rate limits are exceeded`, because I guess it'll also throw if unauthorised, or the query syntax is invalid, or the query is null.

I still  think we've done enough by adding a link to the limits.  By mentioning some limits explicitly, and not others, we're only adding confusion.

If you want to help developers, you need to catch errors in the adaptor and include useful error messages, like. Like catching a `bad message 414` and linking to the rate limit docs.

packages/salesforce/src/Adaptor.js
```diff
@@ -280,16 +283,10 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
...
-  pollTimeout: 90000, // in ms
-  pollInterval: 3000, // in ms
-};
-
 /**
  * Execute an SOQL Bulk Query.
  * This function uses bulk query to efficiently query large data sets and reduce the number of API requests.
- * Note that in an event of a query error,
- * error logs will be printed but the operation will not throw the error.
+ * The `bulkQuery()` uses {@link https://sforce.co/4b6kn6z Bulk API v.2.0} which is available in API version 41.0 and later.
```
josephjclark, line 289: This links to API limits, which is useful but surprising :thinking:

josephjclark (dismissed):
Hmm,sorry Mtuchi, we're moving in the right direction but I'm still a bit unsure about the accuracy and usefulness of these docs.

See comments about stuff which I think could be dangerously misleading. The more specific we get about stuff, the more responsibility we take on to get it right.

packages/salesforce/src/Adaptor.js
mtuchi: Okay i am going to remove all `@throws` for now

packages/salesforce/src/Adaptor.js
mtuchi: Okay got you, i have opened up an issue to track this suggestion https://github.com/OpenFn/adaptors/issues/531

packages/salesforce/src/Adaptor.js
mtuchi: I have updated the links @josephjclark

.changeset/honest-glasses-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-salesforce': patch
+---
+
+- add `query()` limits
```
josephjclark, line 5: We need to do a version bump in a minute. Before we do, can I picky and say that we haven't added any limits - we've just documented existing ones.

Maybe we can simply say:  

> Add documentation about Salesforce API limits to query and bulkQuery

.changeset/honest-glasses-drum.md
```diff
@@ -0,0 +1,6 @@
...
+- improve `bulk()` and `bulkQuery()` function
```
josephjclark, line 6: How have we improved these functions? If AK read these release notes, would she understand the benefits of upgrading to this adaptor version?

josephjclark (approved):
Approved! This final PR is nice and simple. I think it pushes users in the right direction while avoiding a lot of the complexities of rate limiting.

Next step is to update the release notes in the changeset andbump the versions.

.changeset/honest-glasses-drum.md
mtuchi: i just removed the `improve bulk() and bulkQuery() function` comment

packages/salesforce/src/Adaptor.js
```diff
@@ -308,10 +312,18 @@ async function pollJobResult(conn, job, pollInterval, pollTimeout) {
...
-    const [resolvedQs, { pollTimeout = 90000, pollInterval = 3000 }] =
-      newExpandReferences(state, qs, options);
+    const [resolvedQs, resolvedOptions] = newExpandReferences(
+      state,
+      qs,
+      options
+    );
     const apiVersion = connection.version;
 
+    const { pollTimeout, pollInterval } = {
```
josephjclark: Hey @mtuchi what's up with this change? I much prefer the old syntax?

mtuchi: It's buggy, I ran into some problem before. So i think i shouldn't make a change in this PR for now

josephjclark: hmm :thinking: 

OK, at least it's consistent with main, so I'll let this pass

Just version bumps next please! You may need to merge or rebase against main after the surveycto stuff

josephjclark:
Thank you @mtuchi :pray:

Issue #511, opened 2024-04-15
Database configuration schemas


mtuchi:
### Description
 
 Most of database adaptors are using `type: uri` for `hostUrl` which forces the input to contain a protocol. This is not accurate in most cases because a database `hostUrl` can be `hostname`, `ipaddress` or `uri` 
 
 ### Request in few words
Update the `hostUrl` to use the following schema 
```
 "host": {
      "oneOf": [
        {
          "type": "string",
          "format": "hostname"
        },
        {
          "type": "string",
          "format": "ipv4"
        },
        {
          "type": "string",
          "format": "uri"
        }
      ]
    },
 ```
 
 #### Database Adaptor List
- [x] mssql
- [x] mysql
- [x] postgresql
- [ ] ~mongodb~

mtuchi:
@josephjclark should database URL be formatted as `URI` ? Eg `mongodb://xxx`

josephjclark:
The strictness of this probably depends on the adaptor, the client, and the underlying implementation. I don't think I can give you one answer for all databases.

What I would say is: if schema validation is too string and its getting in the way, just make it a string type . The `hostUrl` as you've shown it in the PR description looks way too complicated and for no real benefit.

aleksa-krolls:
@mtuchi @josephjclark should we prioritize this in Mtuchi's backlog?

josephjclark:
@aleksa-krolls it's probably something to fix reactively, as it becomes a problem. I'm not sure it's a priority.

mtuchi:
I have changed the format from `hostname` or `uri`  to `string` for 
- [x] postgresql
- [x] mysql
- [x] mongodb
- [x] mailgun
- [x] hive

See #516

PR #512, opened 2024-04-16
`mssql` update schema and expose cursor

mtuchi:
## Summary

Improved the `configuration-schema` to support extra properties. All properties have default values and can be change if need when setting up a credential for `mssql`

## Details
- Set the following options in configuration-schema with default values 
```
- encrypt = true
- rowCollectionOnRequestCompletion = true
- trustServerCertificate = true
- port = 1433
```

- I have also remove un-used import of `url`
- I have update the configuration-schema for `server` to use one of `hostname, uri, on ipv4` #511 
- Import `cursor()` from `language-common`

## Review Checklist

### DIFF
--- a/packages/mssql/configuration-schema.json
+++ b/packages/mssql/configuration-schema.json
@@ -1,14 +1,27 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft-07/schema#",
   "properties": {
     "server": {
       "title": "Server URL",
-      "type": "string",
       "description": "The database instance server URL or IP address",
-      "format": "uri",
+      "oneOf": [
+        {
+          "type": "string",
+          "format": "hostname"
+        },
+        {
+          "type": "string",
+          "format": "ipv4"
+        },
+        {
+          "type": "string",
+          "format": "uri"
+        }
+      ],
       "minLength": 1,
       "examples": [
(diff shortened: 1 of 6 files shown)


packages/mssql/configuration-schema.json
```diff
@@ -1,14 +1,27 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft-07/schema#",
   "properties": {
     "server": {
       "title": "Server URL",
-      "type": "string",
       "description": "The database instance server URL or IP address",
-      "format": "uri",
+      "oneOf": [
```
mtuchi, line 7: @elias-ba i am curious if this change will break lightning credential form ?

josephjclark:
@mtuchi Would you mind giving this PR a more specific name? I recommend `msql: update schema and expose cursor`

When we have very generic  PR names it gets very hard to go back and audit and understand when and why things changed.

Also, if you find the PR title has too many things in it, maybe that's a sign it should be multiple PRs!

mtuchi:
Noted @josephjclark , i have updated the PR title

packages/mssql/configuration-schema.json
josephjclark: We're going to simplify this so it doesn't matter now :)

josephjclark (approved):

PR #516, opened 2024-04-16
Update configuration-schema for `postgresql`, `mysql`, `hive`

mtuchi:
## Summary

## Details

## Issues

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/hive/configuration-schema.json
+++ b/packages/hive/configuration-schema.json
@@ -5,7 +5,7 @@
       "title": "Host URL",
       "type": "string",
       "description": "The database instance host URL",
-      "format": "uri",
+      "format": "string",
       "minLength": 1,
       "examples": [
         "some-host-url.compute-1.amazonaws.com"
@@ -54,7 +54,7 @@
   "additionalProperties": true,
   "required": [
     "host",
-    "port",
+    "database",
     "username",
     "password"
   ]
--- a/packages/mysql/configuration-schema.json
+++ b/packages/mysql/configuration-schema.json
@@ -1,59 +1,61 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema#",
-    "properties": {
-      "host": {
-        "title": "Host URL",
-        "type": "string",
-        "description": "The database instance host URL",
-        "format": "uri",
-        "minLength": 1,
-        "examples": [
-          "some-host-url.compute-1.amazonaws.com"
-        ]
-      },
-      "port": {
-        "title": "Port",
-        "type": "integer",
-        "default": 3306,
-        "description": "Database instance port",
-        "minLength": 1,
-        "examples": [
-          3306
-        ]
-      },
-      "database": {
-        "title": "Database",
-        "type": "string",
-        "description": "The database name",
-        "minLength": 1,
-        "examples": [
-          "demo-db"
-        ]
-      },
-      "user": {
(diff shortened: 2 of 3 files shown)


.changeset/late-zebras-draw.md
```diff
@@ -0,0 +1,7 @@
+---
+'@openfn/language-hive': patch
+'@openfn/language-mysql': patch
+'@openfn/language-postgresql': patch
+---
+
+update `configuration-schema`
```
josephjclark, line 7: Why are these release notes different? I prefer the more detailed notes in `tame-files-judge.md`

packages/hive/configuration-schema.json
```diff
@@ -54,7 +54,7 @@
   "additionalProperties": true,
   "required": [
     "host",
-    "port",
+    "database",
```
josephjclark, line 57: This looks out of scope - is it intended?

packages/postgresql/configuration-schema.json
```diff
@@ -6,16 +6,8 @@
...
-        {
-          "format": "uri"
-        },
-        {
-          "format": "ipv4"
-        }
-      ],
       "examples": [
-        "https://some-host.compute-1.amazonaws.com",
+        "some-host.compute-1.amazonaws.com",
```
josephjclark, line 10: Is removing https intentional? https would still work right?

We're trying to relax the validation, not change the behaviour or documentation

packages/mysql/configuration-schema.json
```diff
@@ -1,59 +1,61 @@
...
+        "@super(!)Secretpass"
+      ]
+    }
+  },
+  "type": "object",
+  "additionalProperties": true,
+  "required": [
+    "host",
+    "database",
+    "user",
```
josephjclark, line 58: Are we sure user and pass are required? Is this a behaviour change?

packages/hive/configuration-schema.json
mtuchi: It's kind related, There is an issue on docs for displaying only required properties in sample configuration. And hive was missing `database` in required properties 

[See the issue here](https://github.com/OpenFn/docs/issues/471)

packages/mysql/configuration-schema.json
mtuchi: Yes, it's a database connection, You must have a username and password

packages/postgresql/configuration-schema.json
mtuchi: For database host url you need to use `hostname` or `ipAddress` only
There no `http` or `https` protocol

.changeset/late-zebras-draw.md
mtuchi: Thanks for spotting this @josephjclark , the initial requirement changed a bit. 
I have added a changeset per adaptor change which covers the changes in more detail

packages/hive/configuration-schema.json
josephjclark: That's fine - so long as it's deliberate and expected. It should really be flagged in the main PR description, but you've clarified it now :)

josephjclark (approved):

josephjclark:
This is great @mtuchi, thanks. Please bump versions, push tags and merge and your leisure

Issue #518, opened 2024-04-17
surveycto: test `fetchSubmissions()` & make updates if needed


aleksa-krolls:
## Request
Once we confirm which test form & data we can query (see the email to Armand I copied you on), I'd like us to test out this [fetchSubmissions() function](https://docs.openfn.org/adaptors/packages/surveycto-docs#fetchsubmissions) to see if it still works. 

If not, can you please update it by Monday EOD? 

## How to test
Watch the email chain with Armand for confirmation on whether we can deploy & submit test data for this `test form`: https://impactwater.surveycto.com/main.html#Design

For test creds, see LP `Impact Water Survery CTO test account`. Let @daissatou2 if you have any issues logging in/accessing the forms. Please email the client directly if you need different API access enabled for our user. 

See API docs: https://docs.surveycto.com/05-exporting-and-publishing-data/05-api-access/01.api-access.html
https://www.surveycto.com/data-management/6-tips-for-automating-survey-data-management-with-surveycto-api/

aleksa-krolls:
hey @mtuchi fyi I deployed the form with `"form_id": "test"` and submitted 3 form submissions, so should be ready to test whenever you're ready.

aleksa-krolls:
@AishaHassen see my comment above from 5hrs ago (when I deployed the `test` surveycto form). If you've tested the adaptor since then, then pls leave a comment and tag @mtuchi if you have any adaptor testing feedback (and pls share links and screenshots, where you can). Thanks!

mtuchi:
Adding reference to the REST API documentation 👇🏽 , You can use the support link if you have an account with surveyCTO, [REST API documentation](https://support.surveycto.com/hc/en-us/articles/360033156894-REST-API-documentation) 

<details><summary>REST API documentation</summary>

_Note that this documentation is comprehensive - all capabilities are listed. Also please bear in mind that this documentation is distributed with the assumption that interested parties have the expertise and/or time to work out what they need to do based on this documentation._

The REST API can be used by SurveyCTO users on trials, paid and community subscriptions to access server data for both encrypted and not-encrypted forms.

## Downloading data in JSON format

The server also supports downloading wide-format data in JSON format, using a URL like this:

**`https://servername.surveycto.com/api/v2/forms/data/wide/json/formid?date=[D]`**

The response is a JSON array. If there is no data, the response is an empty JSON array (i.e.,  
"\[\]"). Here is a sample JSON response for a form that contains only two submissions:

```json
[
    {
[...]

PR #519, opened 2024-04-19
`surveycto` re-implement `fetchSubmissions()` function

mtuchi:
## Summary

Newly improved `fetchSubmissions()` function for surveyCTO adaptor

## Notes
Using the `common/http` util function. i have replace `request` package that was used in `fetchSubmissions` and improved the function take `formId, options, callback`.  

I have also updated the `configuration-schema` to use `servername` instead of `instanceName` and added `apiVersion` in the schema

Since `surveyCTO` doesn't have enough documentation for the entire application. I have created a `request()` function that can be used to perform any `CRUD` operation in `surveryCTO`

See [API documentation](https://github.com/OpenFn/adaptors/issues/518#issuecomment-2066343459) to learn more about surveyCTO

Ref #518 

## Review Checklist

### DIFF
--- a/packages/surveycto/package.json
+++ b/packages/surveycto/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openfn/language-surveycto",
-  "version": "1.2.1",
+  "version": "2.0.0",
   "description": "A SurveyCTO Language Pack for OpenFn",
   "main": "dist/index.cjs",
   "scripts": {
--- a/packages/surveycto/package.json
+++ b/packages/surveycto/package.json
@@ -20,8 +20,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "^1.12.0",
-    "request": "^2.88.2"
+    "@openfn/language-common": "workspace:^1.12.0"
   },
   "devDependencies": {
     "@openfn/simple-ast": "0.4.1",
@@ -30,7 +29,6 @@
     "deep-eql": "^4.1.2",
     "esno": "^0.16.3",
     "mocha": "^10.1.0",
-    "nock": "^13.2.9",
     "rimraf": "^3.0.2"
   },
   "repository": {
--- a/packages/surveycto/src/Adaptor.js
+++ b/packages/surveycto/src/Adaptor.js
@@ -76,15 +76,14 @@ export function fetchSubmissions(formId, options, callback = s => s) {
 
     const path =
       format === 'csv'
-        ? `/forms/data/csv/${resolvedFormId}`
-        : `/forms/data/wide/${format}/${resolvedFormId}`;
+        ? `forms/data/csv/${resolvedFormId}`
+        : `forms/data/wide/${format}/${resolvedFormId}`;
 
     const contentType =
       format === 'csv' ? 'text/plain;charset=UTF-8' : 'application/json';
 
     console.log(`Fetching '${resolvedFormId}' submissions for: ${date}`);
 
-    console.log(makeSurveyCTODate(date), 'From ada');
     return requestHelper(
       state,
       path,
--- a/packages/surveycto/src/Adaptor.js
+++ b/packages/surveycto/src/Adaptor.js
@@ -1,6 +1,6 @@
 import { execute as commonExecute } from '@openfn/language-common';
 import { expandReferences } from '@openfn/language-common/util';
-import { requestHelper, formDate } from './Utils';
+import { requestHelper, makeSurveyCTODate } from './Utils';
 
 /**
  * Execute a sequence of operations.
@@ -94,7 +94,7 @@ export function fetchSubmissions(
           'content-type': contentType,
         },
         query: {
-          date: formDate(date),
+          date: makeSurveyCTODate(date),
           r: status,
         },
       },
--- a/packages/surveycto/src/Adaptor.js
+++ b/packages/surveycto/src/Adaptor.js
@@ -30,7 +30,6 @@ export function execute(...operations) {
  * @typedef {Object} FormSubmissionOptions
  * @property {string} [date=0] - Form completion or submission date. Default to `0` which will return all submission data
  * @property {string} [format='json'] - Form response type, It can be in `csv` or `json`. Default to `json`
- * @property {string} [timestamp='milliseconds'] - Form date format. Default to `milliseconds`
  * @property {string} status - (Opt)Review status. Can be either, `approved`, `rejected`, `pending` or combine eg `approved|rejected`.
  */
 
@@ -43,7 +42,7 @@ export function execute(...operations) {
  * @example
  * fetchSubmissions('test', { date: 'Apr 18, 2024 6:26:21 AM' });
  * @example
- * fetchSubmissions('test', { date: '2024-04-20', timestamp: 'seconds' });
+ * fetchSubmissions('test', { date: '2024-04-20', format: 'csv' });
  * @example
  * fetchSubmissions('test', {
  *   date: '2024-04-18',
@@ -78,10 +77,9 @@ export function fetchSubmissions(formId, options, callback = s => s) {
     const defaultOptions = {
       date: 0,
       format: 'json',
-      timestamp: 'milliseconds',
     };
 
-    const { date, format, timestamp, status } = {
+    const { date, format, status } = {
       ...defaultOptions,
       ...resolvedOptions,
     };
@@ -93,7 +91,7 @@ export function fetchSubmissions(formId, options, callback = s => s) {
       `/forms/data/wide/${format}/${resovledFormId}`,
       {
         query: {
-          date: formDate(date, timestamp),
+          date: formDate(date),
           r: status,
         },
       },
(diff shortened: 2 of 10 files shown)


mtuchi:
@josephjclark this is ready for review. Just a heads up, This adaptor does not have unit tests 😬, I just want to unblock @AishaHassen testing for now

packages/surveycto/package.json
```diff
@@ -20,8 +20,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "^1.12.0",
-    "request": "^2.88.2"
```
josephjclark, line 24: Nice :clap:

packages/surveycto/src/Utils.js
```diff
@@ -0,0 +1,59 @@
+import { composeNextState } from '@openfn/language-common';
+import {
+  request as commonRequest,
+  logResponse,
+  makeBasicAuthHeader,
+} from '@openfn/language-common/util';
+
+const addBasicAuth = (configuration = {}, headers) => {
```
josephjclark, line 8: I am seeing this get copied into a lot of adaptors. Maybe we need to just push it into common?

This isn't a job for now, just an observation

packages/surveycto/src/Utils.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+};
+
+const buildUrl = (configuration = {}, path) => {
+  const { servername, apiVersion = 'v1' } = configuration;
+  if (!servername) throw 'Please specify servername in your credentials';
+  return `https://${servername}.surveycto.com/api/${apiVersion}${path}`;
+};
+
+export const formDate = date => Math.floor(new Date(date).getTime() / 1000);
```
josephjclark, line 21: Can we give this a better name, like `makeSurveyCTODate`? Or something more descripive?

packages/surveycto/src/Utils.js
```diff
@@ -0,0 +1,59 @@
...
+  const { username, password } = configuration;
+  if (username && password) {
+    Object.assign(headers, makeBasicAuthHeader(username, password));
+  }
+};
+
+const buildUrl = (configuration = {}, path) => {
+  const { servername, apiVersion = 'v1' } = configuration;
+  if (!servername) throw 'Please specify servername in your credentials';
+  return `https://${servername}.surveycto.com/api/${apiVersion}${path}`;
```
josephjclark, line 18: This will break if path doesn't start with `/`. We should add some kind of tolerence.

I would probably put the `/` in the URL and, if the path starts with `/`, I'd substring it

packages/surveycto/src/Utils.js
```diff
@@ -0,0 +1,59 @@
...
+    response: responseWithoutBody,
+  };
+
+  return callback(nextState);
+};
+
+export const requestHelper = (state, path, params, callback = s => s) => {
+  let {
+    body = {},
+    headers = { 'content-type': 'application/json' },
```
josephjclark, line 36: Another note of feedback for `common.http` - we have to do this a lot and it also feels annoying. 

Oh, also, it's not really important by when GETing, we shouldn't include a content-type header (because there is no content)

packages/surveycto/src/Adaptor.js
```diff
@@ -24,92 +21,133 @@ export function execute(...operations) {
...
- * @param {object} params - data to make the fetch
+ * @param {string} formId - Form id
+ * @param {FormSubmissionOptions} options - Form submission date, timestamp, format, status parameters
+ * @param {function} callback - (Optional) Callback function
  * @returns {Operation}
  */
-export function fetchSubmissions(formId, afterDate, postUrl) {
+export function fetchSubmissions(
+  formId,
+  options = { date: 0, format: 'json' },
```
josephjclark, line 69: Is date optional, or is it always required?

packages/surveycto/src/Adaptor.js
```diff
@@ -24,92 +21,133 @@ export function execute(...operations) {
...
+ * @typedef {Object} FormSubmissionOptions
+ * @property {string} [date=0] - Form completion or submission date. Default to `0` which will return all submission data
+ * @property {string} [format='json'] - Form response type, It can be in `csv` or `json`. Default to `json`
+ * @property {string} status - (Opt)Review status. Can be either, `approved`, `rejected`, `pending` or combine eg `approved|rejected`.
+ */
+
+/**
+ * Fetch form submissions
+ * @example <caption>Using state lazy load</caption>
+ * fetchSubmissions($.formId || 'test', { date: '2024-04-18' });
```
josephjclark, line 39: I am not sure I'm ready to add lazy state into our official docs yet. Tricky one.

I think for your should do `fetchSubmissions(state.formId || 'test', { date: '2024-04-18' });`. It's close enough.

packages/surveycto/src/Adaptor.js
```diff
@@ -24,92 +21,133 @@ export function execute(...operations) {
...
+ *   },
+ *   state => {
+ *     console.log('Hello from the callback!');
+ *     return state;
+ *   }
+ * );
  * @function
- * @param {object} params - data to make the fetch
+ * @param {string} formId - Form id
+ * @param {FormSubmissionOptions} options - Form submission date, timestamp, format, status parameters
```
josephjclark, line 63: Date needs better documentation than "form submission date". For one thing, the key is `options.date`, which is only referenced in examples.

I think it's really important to make it clear that you can pass any anything parseable by ` new Date(dateString)` and it will be converted into second resolution for the surveyCTO surver. Because that's the contract we agreed early and that's what the internal API expects.

I guess it's an implementation detail, but anyone who expects to pass a surveyCTO date in seconds is going to get a nasty shock.

Btw can you pass an actual date object? Should we support that?

packages/surveycto/src/Adaptor.js
```diff
@@ -24,92 +21,133 @@ export function execute(...operations) {
...
  */
-export function fetchSubmissions(formId, afterDate, postUrl) {
+export function fetchSubmissions(
+  formId,
+  options = { date: 0, format: 'json' },
+  callback = s => s
+) {
   return state => {
-    const { baseUrl, username, password, instanceName } = state.configuration;
+    const [resovledFormId, { date, format, status }] = expandReferences(
```
josephjclark, line 73: Typo on `resolvedFormId`

packages/surveycto/src/Utils.js
josephjclark: Oh, but actually when we use `requestHelper` at the moment, we ALWAYS pass a `content-type` header (as JSON or CSV).

So let's cut out the complication for now and just not set the default here

packages/surveycto/src/Adaptor.js
mtuchi: Very good question joe 🤔 
Here is what i found out, They have two apiVersions. `v1` and `v2`. 

**In `v1`. date is required. If it's not specified in the query we get.**
```json
{
  "error": {
    "code": 400,
    "message": "Please provide a date using the 'date' parameter.",
    "responseObject": null
  }
}
```
>PS: 
The reason it's default to `0`, is because that's how you get all form submission data in `v1` 

**But in `v2` it's not required. If you don't specify the date, we will get all data by default**

aleksa-krolls:
@mtuchi @josephjclark would it be possible to merge this today? Hoping to walk a client through this ideally later today in a training bc the current surveycto version does not work

packages/surveycto/src/Adaptor.js
aleksa-krolls: @mtuchi confirming `0` is the default value only and users will still have the option to add their own date filter?

packages/surveycto/src/Adaptor.js
aleksa-krolls: @mtuchi just link to the REST API docs (link below) and say the following: 
```
Date (request) parameters should be specified in [UTC time](https://en.wikipedia.org/wiki/Coordinated_Universal_Time). Bypass and request all data by specifying `date=0`.
See API docs: https://support.surveycto.com/hc/en-us/articles/360033156894-REST-API-documentation
```

And you can add the below from their docs if you want more examples: 
```
For example, in order to fetch data after midnight Oct. 12 2015, you can use a URL like one of the following:

https://servername.surveycto.com/api/v2/forms/data/wide/json/formid?date=1444694400

https://servername.surveycto.com/api/v2/forms/data/wide/json/formid?date=1444694400000
```

packages/surveycto/src/Adaptor.js
mtuchi: @aleksa-krolls we have two default values `date: 0`, which will return all form submission and `format: json` which will return the submissions as JSON object. 

The other format is `csv` which will return the result as CSV string

packages/surveycto/src/Adaptor.js
mtuchi: I have removed the date conversion, Now you can pass a either `unix timestamp` in seconds/milliseconds or a date time with this format `MMM dd, yyyy h:mm:ss a`
cc @aleksa-krolls @josephjclark

josephjclark (approved):

mtuchi:
@josephjclark i have bump version and push tags ✅

PR #524, opened 2024-04-23
surveycto: add new cursor

### LINKED ISSUE #520: surveycto: export custom cursor function
Survey CTO uses timestamps in second-precision, so it accepts timestamps in two formats: a unix timestamp or a particular `mm dd` format (I think it's  `MMM dd, yyy h:mm:ss a`, but this needs checking)

This is great if you're a surveyCTO developer and want to write a date in the correct format by hand.

But for other developers, or for dynamic timestamps (ie if you want to do `cursor('today')` or `new Date()`), it's quite hard to get the date into the correct format.

The answer is for our `surveycto` adaptor to override the cursor function in common. The overridden cursor should accept timestamps in basically any input format - ISO strings, epoch timestamps, ddmm strings, whatever - and convert them all into the expected `mmm dd...` format (we use this rather than unix because it's human readable)

The cursor function should basically absorb all the complexity of date conversions, ensuring that all `state.cursor` values are in the correct format.

To facilitate this, it would be useful for the common cursor to have a `format` function as an argument. This function would be called just  before cursor writes to state.cursor, giving you the opportunity to standardise the cursor value

So to call the common cursor you would do:
```
cursor('today', { format: (date) => dateFns.format(date, 'MMM dd, yyy h:mm:ss a')
```
I suppose it would also be helpful to export surveycto's format function for use in job code, in case any job writer wants to take a date and convert it in an ad-hoc way. This should be exported as a util, not an operation.

josephjclark:
## Summary

This PR adds:

* A `format` function in the common cursor
* Simplified timestamp handling in surveyCTO
* A custom `cursor` exported by `surveycto`

## Issues

Closes #520 

## Simplified Date handling

The `date` option in `fetchSubmissions` can now accept a surveyCTO date, an epoch or unix timestamp, or basically anything you can pass into new Date(). This has been documented lightly in a way that encourages the CTO `MMM dd, yyy h:mm:ss a` format, but doesn't prescribe it.

The "magic" is handled by a little function which takes a date value and _guarantees_ to return a string formatted in the surveyCTO way.

Why settle on this format? Because it's quite human readable, so when queries get printed in the log, you can see the actual parameters without having to do the maths on the timestamp.

I also think it's important for openfn users who aren't surveyCTO experts to be able to use dates in the way they expect. Just pass a date into the date property - we'll figure out the formatting. Life's too short to worry about this stuff.

Documentation has been updated and simplified to this effect.

The actual support for this is implemented in `fetchSubmissions` and the new `cursor` function.

## common.cursor.format

The new format function in common allows you to intercept the cursor value just before it gets written to state, so that you can specifier your own formats (or even values). It's more than a formatter tbh but its usage feels quite intuitive as a `format` (as opposed to `parse` or something).

See the docs PR for usage: https://github.com/OpenFn/docs/pull/479

## surveycto.cursor

surveyCTO takes advantage of the new cursor functionality.

It exports its own `cursor` function with a formatter baked in. So when you pass a date into `cursor`, what will actually be written to state is a surveyCTO style string.

You can pass in `today` or `Apr 18, 2024 6:26:21 AM` or `2024-04-23T15:49:31.764Z` and it'll all just work. The result is compatible with `filterSubmissions`.


packages/surveycto/src/Adaptor.js
```diff
@@ -26,43 +29,45 @@ export function execute(...operations) {
...
+ * If set to 0, all submissions will be retrieved.
+ * @property {string} [format=json] - Format the submission data type as  `csv` or `json`.
+ * @property {string} [status] - Review status. Can be either, `approved`, `rejected`, `pending` or combine eg `approved|rejected`.
  */
 
 /**
  * Fetch form submissions
  * @example <caption>Fetch all form submissions</caption>
  * fetchSubmissions('test');
- * @example <caption> With `MMM dd, yyyy h:mm:ss a` date format</caption>
```
josephjclark, line 40: These examples have been trimmed down so that each one only shows one thing. Ie, the CSV example doesn't include a date.

packages/surveycto/src/Utils.js
```diff
@@ -50,3 +51,43 @@ export const requestHelper = (state, path, params, callback = s => s) => {
...
+  }
+
+  // And return in the correct formatting (utc time)
+  const formatted = formatInTimeZone(
+    dateObj.toISOString(),
+    'UTC',
+    'MMM dd, yyy h:mm:ss a'
+  );
+
+  console.log(`Converted timestamp "${date}" to "${formatted}" (UTC)`);
```
josephjclark, line 90: I added this as an afterthought because it seems quite important :thinking: 

On the other hand it's noisy and maybe a bit defensive. Opinions welcome!

josephjclark:
What is even up with this failture while building the ast:

image

When I build locally on main I get basically the same error

josephjclark:
Hi @mtuchi I think this will be interesting to you - do you have time to take a look today?

mtuchi:
I ran `pnpm i & pnpm build` i am seeing new changes in ast.json file 
```
modified:   packages/asana/ast.json
modified:   packages/common/ast.json
modified:   packages/googlesheets/ast.json
modified:   packages/http/ast.json
modified:   packages/kobotoolbox/ast.json
modified:   packages/msgraph/ast.json
modified:   packages/mssql/ast.json
modified:   packages/surveycto/ast.json
```

josephjclark:
Thank you @mtuchi , that's fine, we still need to do the release engineering stuff. I'm more interested in whether this behaviour is as useful as I hope it is!

mtuchi:
The cursor is working but i am seeing these double logs with the cursor 
```
Compiled all expressions in workflow
Converted timestamp "Thu Apr 25 2024 00:00:00 GMT+0300 (East Africa Time)" to "Apr 24, 2024 9:00:00 PM" (UTC)
Setting cursor "today" to: Apr 24, 2024 9:00:00 PM
Converted timestamp "Apr 24, 2024 9:00:00 PM" to "Apr 24, 2024 6:00:00 PM" (UTC)
Fetching 'test' submissions for: Apr 24, 2024 6:00:00 PM
GET https://impactwater.surveycto.com/api/v1/forms/data/wide/json/test - 200 in 828ms
```
When i run the following code 
```
cursor('today');
fetchSubmissions('test', { date: $.cursor });
```

My gut tells me the cursor should also tell me the date in `unix timestamp`. I think it's useful when i need to debug. But i don't think it's very important since the cursor date format is in human readable and i can easily relate to that

mtuchi:
What's you thought on supporting something like this ?
```
cursor('today', {format: 'unix'});
fetchSubmissions('test', { date: $.cursor });
```

josephjclark:
Hmm, yeah the logging is tricky because it's hard to set context. I'll think about it because the logs are definitely too verbose right now.

What would be your expected behavour for this?
```
cursor('today', {format: 'unix'});
```
To write state.cursor as `1714057191`?

What would the utility of that be?

mtuchi:
I guess the use case is setting the `state.cursor` to unix timestamp because i think we still want to support unix timestamp right ?
Eg: `date:'1714057191'` Should still work right ?
```
fetchSubmissions('test', { date: '1714057191' });

//Logs
Converted timestamp "1714057191" to "Apr 25, 2024 2:59:51 PM" (UTC)
Fetching 'test' submissions for: Apr 25, 2024 2:59:51 PM
```

mtuchi:
I am not sure if a user specify a unix timestamp, we should convert it to UTC?
Right now any date input gates converted to UTC. Eg 
```
fetchSubmissions('test', { date: '1714057191' });

//Logs
Converted timestamp "1714057191" to "Apr 25, 2024 2:59:51 PM" (UTC)
Fetching 'test' submissions for: Apr 25, 2024 2:59:51 PM
```
I guess it doesn't hurt making sure the timestamp is in UTC 🤷🏽 .

Also if i am in UTC timezone i don't think we need to convert the time to UTC right ?

josephjclark:
Users can still pass unix timestamps in. We'll still convert them for strings, which gives clearer logs.

We could choose to leave unix timestamps alone but the idea of all of this for the `cursor` to take any reasonable date format and "normalise" it to a UTC string. That's the whole idea.

So this works fine: `fetchSubmissions('test', { date: '1714057191' });` (wait, is the timestamp a number or a string?)

Don't get hunt up on "converting to UTC", the point is more that we convert to String. Unix and Epoch timestamps are always in UTC, so we're basically just stringifying them for readability.

mtuchi:
Okay got you 👍🏽 

Regarding 
>So this works fine: fetchSubmissions('test', { date: '1714057191' }); (wait, is the timestamp a number or a string?)

I have tried both number and string and it worked 

```
fetchSubmissions('test', { date: 1714057191 });
fetchSubmissions('test', { date: '1714057191' });
// Both log 👇🏽 
Converted timestamp "1714057191" to "Apr 25, 2024 2:59:51 PM" (UTC)
Fetching 'test' submissions for: Apr 25, 2024 2:59:51 PM
```

josephjclark:
@mtuchi I think this is ready for release and we're both happy with it?

Would you mind giving it an approval before I merge?

Sadly because I've touched common all the adaptors have been bumped, even thouhh they're unaffected (only adaptors exporting cursor are affected). I've raised an issue to look into this but tbh I think it's a neccessary evil

mtuchi (approved):

mtuchi:
@josephjclark you can go ahead build & push tags then squash merge this pr

PR #540, opened 2024-05-07
http: don't override the auth header

### LINKED ISSUE #539: Http forces basic Auth if username and password is set
If user and password is set on config, http will force a basic Auth header to be added.

This is no good if the user is passing their own authorization header.

From community:

```
post('/myEndpoint', {
  body: state.data,
  headers: {
    'Content-Type': 'text/plain; charset=utf-8',
    'Accept': 'application/json',
    'Authorization': `Bearer: ${state.configuration.password}`
  }
})
```

josephjclark:
We may have stumbled on an issue where if state.configuration has a username and password,  AND the user passes their own Authorization header, we overwrite their header with own own basic auth.

This feels wrong: if the user has defined their own authorization header, we should use it regardless of what exists on state.

Fixes #539

### DIFF
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Fixed an issue where the Authorization header could be overwritten if username
   and password are set on config
+- Configuration schema: demoted `username` and `password` to optional
+- Configuration schema: Added `token`. This is not automated and must be
+  manually add to the HTTP headers.
 
 ## 6.2.0
 
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-http
 
+## 6.2.1
+
+### Patch Changes
+
+- Fixed an issue where the Authorization header could be overwritten if username
+  and password are set on config
+
 ## 6.2.0
 
 ### Minor Changes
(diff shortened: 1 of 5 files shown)


josephjclark:
Hey @mtuchi would you mind giving this a super quick review?

mtuchi:
@josephjclark i couldn't test this properly locally because if you use a `baseUrl` like this `https://api.github.com` You get an error saying `[R/T] ✘ TypeError: TypeError: Invalid URL`. 

But basically what i wanted to test is 
`state.json`
```
{
    "configuration": {
        "baseUrl": "https://api.openconceptlab.org"
        "username": "",
        "password": "",
        "ghToken": ""
    }
}
```

Then my job code 

```
// This will use basic Auth
get('/mappings');

// This will use accessToken 
get('https://api.github.com/emojis', {
  headers: {
    Authorization: `Bearer ${$.configuration.ghToken}`,
    'X-GitHub-Api-Version': '2022-11-28',
  },
});
```

I think the changes are okay but i wanted to make sure that i would be able to connect and pull data from two different systems in one job.

josephjclark:
@mtuchi that code example should work fine. If `Authorization` isn't explicitly passed then basic auth will be applied by default.

What's up with the baseurl thing? Do we have an issue there?

mtuchi:
@josephjclark i figured out the problem. It was with my `get('/mappings');` operation, it's suppose to be `get('/mappings/');`

Issue #528, opened 2024-04-25
`commcare` add function


aleksa-krolls:
## Request
Currently our adaptor does not support extracting `case`, `form`, `user`, `group` and `lookup table` data out of CommCare.

See CommCare API docs: 
1. [Get cases](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957300/List+Cases+version+3)
2. [Get a single case](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957360)
3. [Get users](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957296/List+Mobile+Workers)
4. [Get lookup table data](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957453/Fixture+Data+APIs+or+Lookup+Tables)
5. [Get forms](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957334/List+Forms)
6. [Get a single form](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957429/Form+Data+APIs)

_Plus_ we should also ensure that all `http` functions are available in this adaptor, in case users want to send http requests to other api endpoints not listed above. 

## Examples
- See get `cases` data example: https://github.com/OpenFn/lwala/blob/master/bulk/bulkGetCases-Person.js#L18-L19
- See get `form` data example: https://github.com/OpenFn/grassroot-soccer/blob/master/jobs/1.getForms.js

## To test
See LP `CommCare Demo Project` for a credential to our `test` project.

josephjclark:
Note from mtuchi: we may want to rename "application name" to "domain" to better reflect commcare's own language.

Also app id may not be needed in each request

josephjclark:
Hi @aleksa-krolls,

How important is this ticket? I am thinking about asking Hunter to do it now as she's already working on the commcare adaptor. Make sense to me to pack a few features in while shes in the headspace for it.

After a quick chat with @mtuchi , I would suggest we expose general `get` and `post` helper operations, so that users can get at any resources they need.

```
get('cases', { query: { name: 'Buck Rogers' }})
```
or
```
```
get('/cases?name="Buck Rogers")  // (note to me, we need to ensure the URL is safely escaped, does undici do that for us?)
```

Then later we can think about adding special `getCase`, `getCases` etc operations for the most commonly used ones.

Sound cool?

aleksa-krolls:
[...]

PR #545, opened 2024-05-07
Commcare: replace superagent with common http helpers

### LINKED ISSUE #414: Commcare: don't use language-http

hunterachieng:
## Summary

Switch `commcare` adaptor implementation to use `fetch` for all http calls

## Implementation Details

Currently the `commcare` module utilizes superagent and `language-http` to handle its requests.  Our aim was to switch the implementation with normal fetch.

## How to Test

- [x] Uses the standard format for http requests defined in https://github.com/OpenFn/adaptors/blob/main/packages/template/src/Utils.js
- [x] Have no reference to  superagent or `language-http` . 

## Issues

Fixes #414

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -42,7 +42,7 @@ async function clientPost({ state, path, body }) {
     method: 'POST',
     path: path,
     data: body,
-    contentType: 'text/html',
+    contentType: 'text/xml',
     parseAs: 'text',
   });
 
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -71,14 +71,12 @@ async function clientPost({ state, path, body }) {
  */
 export function submitXls(formData, params) {
   return async state => {
-    const { applicationName, username, apiKey, hostUrl } = state.configuration;
+    const { applicationName, username, apiKey } = state.configuration;
 
     const [json] = expandReferences(state, formData);
     const { case_type, search_field, create_new_cases } = params;
 
     const path = `/a/${applicationName}/importer/excel/bulk_upload_api/`;
-
-    const url = (hostUrl || 'https://www.commcarehq.org').concat(path);
 
     const workbook = xlsx.utils.book_new();
     const worksheet = xlsx.utils.json_to_sheet(json);
@@ -96,8 +94,6 @@ export function submitXls(formData, params) {
     data.append('case_type', case_type);
     data.append('search_field', search_field);
     data.append('create_new_cases', create_new_cases);
-
-    console.log('Posting to url: '.concat(url));
 
     const response = await request({
       state,
@@ -144,14 +140,10 @@ export function submit(formData) {
       // it is what lives after www.commcarehq.org/a/...
       applicationName,
       appId,
-      hostUrl,
     } = state.configuration;
 
     const path = `/a/${applicationName}/receiver/${appId}/`;
 
-    const url = (hostUrl || 'https://www.commcarehq.org').concat(path);
-
-    console.log('Posting to url: '.concat(url));
     console.log('Raw JSON body: '.concat(JSON.stringify(jsonBody)));
     console.log('X-form submission: '.concat(body));
 
@@ -175,13 +167,7 @@ export function fetchReportData(reportId, params, postUrl) {
   return async state => {
     const path = `/a/${state.configuration.applicationName}/api/v0.5/configurablereportdata/${reportId}/`;
 
-    console.log(
-      'getting from url: '.concat(
-        state.configuration.hostUrl,
-        `/a/${state.configuration.applicationName}/api/v0.5/configurablereportdata/${reportId}/`
-      )
-    );
-    console.log('with params: '.concat(params));
+    console.log('with params: '.concat(JSON.stringify(params)));
 
     return await request({
       state,
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -164,7 +164,7 @@ export function submit(formData) {
  * @returns {Operation}
  */
 export function fetchReportData(reportId, params, postUrl) {
-  return async state => {
+  return state => {
     const path = `/a/${state.configuration.applicationName}/api/v0.5/configurablereportdata/${reportId}/`;
 
     console.log('with params: '.concat(JSON.stringify(params)));
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -1,11 +1,5 @@
-import {
-  execute as commonExecute,
-  http,
-  expandReferences,
-} from '@openfn/language-common';
-import pkg from '@openfn/language-http';
-const { get, post } = pkg;
-// import request from 'superagent';
+import { execute as commonExecute } from '@openfn/language-common';
+import { expandReferences } from '@openfn/language-common/util';
 import FormData from 'form-data';
 import js2xmlparser from 'js2xmlparser';
 import xlsx from 'xlsx';
@@ -30,10 +24,6 @@ export function execute(...operations) {
   };
 
   return state => {
-    // state.configuration.authType = 'basic';
-    // state.configuration.baseUrl = 'https://www.commcarehq.org/a/'.concat(
-    //   state.configuration.applicationName
-    // );
     return commonExecute(...operations)({ ...initialState, ...state });
   };
 }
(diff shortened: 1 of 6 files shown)


packages/commcare/src/Utils.js
```diff
@@ -0,0 +1,43 @@
...
+  const { body, ...responseWithoutBody } = response;
+  const nextState = {
+    ...composeNextState(state, response.body),
+    response: responseWithoutBody,
+  };
+
+  return callback(nextState);
+};
+
+export function request(state, method, path, data, params, header, authType) {
```
josephjclark, line 18: This signature is getting very long! Instead of declaring all these parameters, I would just declare one: an object.

Basically just wrap the parameters up into an object, like this:
```
export function request({ state, method, path, data, params, header, authType })
```

packages/commcare/src/Adaptor.js
```diff
@@ -203,29 +172,41 @@ export function submit(formData) {
...
+
+    console.log(
+      'getting from url: '.concat(
+        state.configuration.hostUrl,
+        `/a/${state.configuration.applicationName}/api/v0.5/configurablereportdata/${reportId}/`
+      )
+    );
+    console.log('with params: '.concat(params));
+
+    return await request({
```
josephjclark, line 186: You should not mix `await` and `then`. Either use a promise chain (`then().then()`), or async/await.

Both styles are acceptable, just not at the same time :sweat_smile:

In this case you can just drop the `await` keyword and return the promise chain. That works nicely.

packages/commcare/src/Adaptor.js
```diff
@@ -44,23 +33,21 @@ export function execute(...operations) {
...
+    state,
+    method: 'POST',
+    path: path,
+    data: body,
+    authType: 'basic',
+    contentType: 'text/html',
+    parseAs: 'text',
   });
+
+  return prepareNextState(state, response, s => s);
```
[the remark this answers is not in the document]
josephjclark: Ah, to make this work more cleanly, you need to default the parameter value _inside_ `prepareNextState`

```
export const prepareNextState = (state, response, callback = s => s) =>  { ...}
```

packages/commcare/src/Utils.js
```diff
@@ -0,0 +1,53 @@
...
+  const options = {
+    body: data,
+    headers: headers,
+    query: params,
+    parseAs: parseAs,
+  };
+
+  const url = `${hostUrl}${path}`;
+
+  return commonRequest(method, url, options).then(response => {
```
josephjclark, line 50: I think you can just do:

```
return commonRequest(method, url, options).then(logResponse);
```

josephjclark (commented):
A few comments to look into tomorrow Hunter!

It's generally a nice, clean, minimal PR. I think you can make it a bit thinner, and you've hit upon a couple of quirks of the old adaptor which we might need to address. I'll sleep on it and let you know tomorrow.

packages/commcare/src/Adaptor.js
```diff
@@ -113,21 +98,20 @@ export function submitXls(formData, params) {
...
-      });
+
+    const response = await request({
+      state,
+      method: 'POST',
+      path: path,
+      data,
+      header: {
+        ...data.getHeaders(),
+        Authorization: `ApiKey ${username}:${apiKey}`,
```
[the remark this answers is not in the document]
hunterachieng: Yes. I had a conversation with @mtuchi  but if I change it to use basic auth it might create disruptions in jobs using the adaptor so I think creating another issue to handle this would be ideal

packages/commcare/src/Adaptor.js
```diff
@@ -113,21 +98,20 @@ export function submitXls(formData, params) {
...
+      },
+      authType: 'APIKEY',
```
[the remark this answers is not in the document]
hunterachieng: I wrote that to distinguish the auth type since we have two different ways of doing auth. Let me refactor it

packages/commcare/src/Adaptor.js
```diff
@@ -44,23 +33,21 @@ export function execute(...operations) {
...
-      .auth(username, password)
-      .set('Content-Type', 'application/xml')
-      .send(body)
-      .end((error, res) => {
-        if (!!error || !res.ok) {
-          reject(error);
-        }
-        resolve(res);
-      });
+async function clientPost({ state, path, body }) {
```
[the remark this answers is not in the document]
josephjclark: I just wanna run this past @mtuchi, but the decision is:

- Let's break the API and move this function into utils. Call it `post`, which is a thin wrapper around `request`. It should not take state. Oh and do mark it `@private`
- Let's release this and #528 at the same time in the same major version

packages/commcare/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Great - raised as #554

packages/commcare/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Ok, nice idea, but I think right now it adds more confusion than clarity.

I've raised an issue to refactor auth - let's  make sure we have a nice solution there. In the meantime I recommend removing this line.

packages/commcare/src/Adaptor.js
```diff
@@ -151,42 +131,23 @@ export function submitXls(formData, params) {
  * @returns {Operation}
  */
 export function submit(formData) {
-  return state => {
-    const jsonBody = expandReferences(formData)(state);
+  return async state => {
```
josephjclark, line 134: I'm being picky but this doesn't need to be async anymore :)

packages/commcare/src/Adaptor.js
```diff
@@ -203,29 +164,35 @@ export function submit(formData) {
...
-      console.log('with params: '.concat(params));
-      return params;
-    },
-    callback: function (state) {
-      var reportData = state.response.body;
-      return post(postUrl, {
-        body: reportData,
-      })(state).then(function (state) {
-        delete state.response;
+  return async state => {
```
josephjclark, line 167: This one also shouldn't be async. We only use async if there's an `await` statement.

packages/commcare/src/Utils.js
```diff
@@ -0,0 +1,51 @@
...
+  method,
+  path,
+  data,
+  params = {},
+  header = {},
+  authType,
+  contentType = 'application/json',
+  parseAs = 'json',
+}) {
+  const { hostUrl, username, password } = state.configuration;
```
josephjclark, line 29: Ah this is why we need to pass state. hmm.

Ok, maybe we'll address this in the auth issue #554. Ignore my comments on state for now

josephjclark: Basically I think we should pass an `auth` object, which is basically state.config, and we decide what auth to used based on the keys.

packages/commcare/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Hey @hunterachieng , sorry I should have posted in Friday - we're going to go ahead and do this! Please move and rename this function

packages/commcare/src/Adaptor.js
```diff
@@ -44,23 +33,20 @@ export function execute(...operations) {
...
-        }
-        resolve(res);
-      });
+async function clientPost({ state, path, body }) {
+  const response = await request({
+    state,
+    method: 'POST',
+    path: path,
+    data: body,
+    contentType: 'text/html',
```
josephjclark, line 45: Oh hey, this changes from `text/xml` to `text/html`. Is that correct?

packages/commcare/src/Adaptor.js
```diff
@@ -203,29 +162,35 @@ export function submit(formData) {
...
-      return post(postUrl, {
-        body: reportData,
-      })(state).then(function (state) {
-        delete state.response;
+  return state => {
+    const path = `/a/${state.configuration.applicationName}/api/v0.5/configurablereportdata/${reportId}/`;
+
+    console.log('with params: '.concat(JSON.stringify(params)));
+
+    return request({
```
josephjclark, line 170: I've just had an idea!

The nested promises here have always seemed weird to me. Nothing wrong with it, just wierd and a little hard to follow the logic.

Does it get any clearer if we re-write this whole chain with async/await?

josephjclark (commented):
Hey @hunterachieng we're nearly there :)

Two extra things on top of the comments:

1) do you have a test job you're using against commcare? Could you check it in under `tests/integration/whatever.js`? This isn't a unit test, it's something that can be manually run with the cli. 

2) we need to generate a change set! There's notes in the readme but run `pnpm changeset` from root and add a major change with a note about what you've done, then check the file in.

packages/commcare/src/Adaptor.js
josephjclark: This is untested bit I was thinking about this:
```
const { body: reportData } = await request({
  state,
  method: 'GET',
  path,
  authType: 'basic',
})
const result = await request({
  state,
  method: 'POST',
  path: postUrl,
  params,
  data: reportData,
  authType: 'basic',
});

delete result.response;
console.log('fetchReportData succeeded.');
console.log('Posted to: '.concat(postUrl));
return prepareNextState(state, {});
```

packages/commcare/src/Utils.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * @private
+ *  * @function
+ * @param {string} path - Path to resource
+ * @param {object} data - Object which defines data that will be used to create a given instance of resource
+ * @param {string} contentType- Optional field for the header content-type
+ * @param {string} parseAs- Optional for data response parse i.e json
+ */
+export const post = async ({
```
josephjclark, line 16: Ok, this isn't quite what I was expecting!

The `post` function will never work, because it doesn't have access to auth credentials.

I note that you've refactored Adaptor.js to not use `post` at all, but instead use request. This is fine but not what I expected.

Let's just remove the `post` function entirely, since it's broken and we're not using it :)

hunterachieng: I tried using the post but the issue was that it has its own state function inside it and the function inside the adaptor. When I run it, I get undefined

Issue #554, opened 2024-05-10
commcare: standardise auth


josephjclark:
commcare at the moment uses two different auth schemes: basic and api key.

This is fine, but a big problem is that different functions supports different schemes.

I presume that any commcare API can use either auth, but this needs validating.

We should ensure that the new `request` helper is smart enough to load the two different credential types from config, and set up the header appropriately.

## Design

Here's what I suggest

When calling request, we should pass state.config as the first argument:

```
request(state.configuration, { path: 'case' })
```

That way, you don't need to pass the whole state object.

The request signature should like be:
```
request(auth: Object, options: Object): Promise<>
```

We should not need to pass `authType` into the request at all.

Inside `request`, we set set the Authorization differently depending on the shape of `auth`:
* If there's an apikey,we do `Authorization: `ApiKey ${username}:${apiKey}`,`
* Otherwise if there's a password, we do: `Authorisation: ${Buffer.from(`${username}:${password}`).toString('base64')}`
* Otherwise we should probably throw an error like "no authorisation credentials provided"
* Note that we don't do any of this if an Authorization header has already been set

We probably want a helper function like:
```
function configureAuth(auth, headers = {}) {
   // ...
  return headers
}
```

I'd love to see some unit tests against that function!

hunterachieng:
@aleksa-krolls  This has been merged into this [branch](https://github.com/OpenFn/adaptors/tree/release/commcare)

PR #560, opened 2024-05-15
Commcare: standardize auth

hunterachieng:
## Summary

Creating a single auth systems that will check the `config` and authorize depending on the availability of an api key or password

## Details

The current implementation of auth is to allow for both basic and API key at the same time and we needed to make it better by only allowing one type of auth depending on the `config`. 

## Issues

#554 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/commcare/src/Utils.js
+++ b/packages/commcare/src/Utils.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-param-reassign */
 import { composeNextState } from '@openfn/language-common';
 import {
   request as commonRequest,
--- a/packages/commcare/src/Utils.js
+++ b/packages/commcare/src/Utils.js
@@ -7,18 +7,16 @@ import {
 } from '@openfn/language-common/util';
 
 export const configureAuth = (auth, headers = {}) => {
-  const configKeys = Object.keys(auth);
-
-  if (!configKeys.includes('apiKey') && !configKeys.includes('password')) {
-    throw new Error('No authorization credentials provided');
-  } else if (configKeys.includes('apiKey')) {
-    headers = {
+  if ('apiKey' in auth) {
+    Object.assign(headers, {
       Authorization: `ApiKey ${auth.username}:${auth.apiKey}`,
-    };
+    });
+  } else if ('password' in auth) {
+    Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password));
   } else {
-    headers = {
-      ...makeBasicAuthHeader(auth.username, auth.password),
-    };
+    throw new Error(
+      'Invalid authorization credentials. Include an apiKey or password in state.configuration'
+    );
   }
 
   return headers;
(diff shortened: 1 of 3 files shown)


packages/commcare/src/Utils.js
```diff
@@ -1,10 +1,29 @@
...
 import {
   request as commonRequest,
   makeBasicAuthHeader,
   logResponse,
 } from '@openfn/language-common/util';
 
+export const configureAuth = (auth, headers = {}) => {
+  const configKeys = Object.keys(auth);
+
+  if (!configKeys.includes('apiKey') && !configKeys.includes('password')) {
```
josephjclark, line 12: I would restructure this if-else branch to be more positive. I tend to avoid `if(!x) {} else { }` branches, I much prefer to do `if(x) {} else { }`.

There are exceptions of course but I don't think this is one :)

So I'd write this as:

```
if ('apiKey' in auth) {
  headers = {
    Authorization: `ApiKey ${auth.username}:${auth.apiKey}`,
  };
} else if ('password' in auth) {
  headers = {
    ...makeBasicAuthHeader(auth.username, auth.password),
  };
} else {
  throw new Error('Invalid authorization credentials. Include an apiKey or password in state.configuration');
}
```

The error state is also a bit more comprehensive I think

josephjclark: I'be not tested this, but I'd also be temped to use Object.assign  - I think it's a bit nicer than the spread?
```
if ('apiKey' in auth) {
  Object.assign(headers, {
    Authorization: `ApiKey ${auth.username}:${auth.apiKey}`,
  });
} else if ('password' in auth) {
  Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password))
  };
} else {
  throw new Error('Invalid authorization credentials. Include an apiKey or password in state.configuration');
}
```

josephjclark (changes requested):
Looks good @hunterachieng, thank you!

I've made a couple of suggestions in the name of readability, but your approach is correct.

I'm a little unhappy about the signature of `request` . I'm not sure that this the place to deal with it - I'm just going to have a little think about it

josephjclark:
Yeah OK, I would like to change the `request` signature here. It's a bit feature creep but it's also related.

I don't like that we have to pass `state` into request just so that we can pull the auth keys out of it. Feels wrong to me.

I also think that auth keys and path are required for every request, and everything else is optional.

So I'd like to change the signature to this:
```
export function request(configuration, path, opts) {
   // And then you can expand the opts object out and set your defaults
  const opts = {
    method,
    data,
    params = {},
    headers = {},
    contentType = 'application/json',
    parseAs = 'json',
  } 
}
```

Now Adaptor.js needs to call `request` like:
```
  const result = await request(state.configuration, postUrl, {
      method: 'POST',
      params,
      data: reportData,
    });
```

I think this is way cleaner, and will make it easier to add the new `get()` and `post()` operations in the next issue

packages/commcare/src/Utils.js
```diff
@@ -1,10 +1,27 @@
+/* eslint-disable no-param-reassign */
```
josephjclark, line 1: Do we still need this?

PR #563, opened 2024-05-16
Commcare: Get function handler

hunterachieng:
## Summary

Creating a `get` function to be used to fetch data from `commcare`

## Details

Commcare does not have a get handler that can be used to fetch various data. The `get` function will be used to fetch data such as `case, forms`.

## Issues

#528 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -54,16 +54,21 @@ export function get(path, params = {}) {
       params
     );
 
-    const response = await request(
-      state.configuration,
-      `/a/${applicationName}/api/v0.5/${resolvedPath}`,
-      {
-        method: 'GET',
-        params: resolvedParams,
-      }
-    );
+    try {
+      const response = await request(
+        state.configuration,
+        `/a/${applicationName}/api/v0.5/${resolvedPath}`,
+        {
+          method: 'GET',
+          params: resolvedParams,
+          contentType: 'application/json',
+        }
+      );
 
-    return prepareNextState(state, response);
+      return prepareNextState(state, response);
+    } catch (e) {
+      throw e.body ?? e;
+    }
   };
 }
 
@@ -192,6 +197,7 @@ export function fetchReportData(reportId, params, postUrl) {
 
     const { body: reportData } = await request(state.configuration, path, {
       method: 'GET',
+      contentType: 'application/json',
     });
 
     const result = await request(state.configuration, postUrl, {
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -74,6 +74,7 @@ export function get(path, params = {}, callback = s => s) {
           contentType: 'application/json',
         }
       );
+      console.log({response});
 
       return prepareNextState(state, response, callback);
     } catch (e) {
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -74,7 +74,6 @@ export function get(path, params = {}, callback = s => s) {
           contentType: 'application/json',
         }
       );
-      console.log({response});
 
       return prepareNextState(state, response, callback);
     } catch (e) {
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -40,6 +40,9 @@ export function execute(...operations) {
  *      offset:0,
  *    }
  * )
+ * // The returned objects will be written into state.data,
+ * // and the meta object will be written to state.response
+ * // along with the status code and returned headers.
  * @function
  * @param {string} path - Path to resource
  * @param {Object} params - Optional request params such as limit and offset.
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -29,6 +29,44 @@ export function execute(...operations) {
 }
 
 /**
+ * Make a get request to any commcare endpoint
+ * @public
+ * @example
+ * get(
+ *    "case"
+ *    {
+ *      limit: 1,
+ *      offset:0,
+ *    }
+ * )
+ * @function
+ * @param {string} path - Path to resource
+ * @param {Object} params - Optional request params such as limit and offset.
+ * @returns {Operation}
+ */
+export function get(path, params = {}) {
+  return async state => {
+    const { applicationName } = state.configuration;
+    const [resolvedPath, resolvedParams] = expandReferences(
+      state,
+      path,
+      params
+    );
+
+    const response = await request(
+      state.configuration,
+      `/a/${applicationName}/api/v0.5/${resolvedPath}`,
+      {
+        method: 'GET',
+        params: resolvedParams,
+      }
+    );
+
+    return prepareNextState(state, response);
+  };
+}
+
+/**
  * Convert form data to xls then submit.
  * @public
  * @example
@@ -158,14 +196,12 @@ export function fetchReportData(reportId, params, postUrl) {
 
     const { body: reportData } = await request(state.configuration, path, {
       method: 'GET',
-      authType: 'basic',
     });
 
     const result = await request(state.configuration, postUrl, {
       method: 'POST',
       params,
       data: reportData,
-      authType: 'basic',
     });
 
     delete result.response;
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -52,9 +52,10 @@ export function execute(...operations) {
  * @function
  * @param {string} path - Path to resource
  * @param {Object} params - Optional request params such as limit and offset.
+ * @param {function} [callback] - Optional callback to handle the response
  * @returns {Operation}
  */
-export function get(path, params = {}) {
+export function get(path, params = {}, callback = s => s) {
   return async state => {
     const { applicationName } = state.configuration;
     const [resolvedPath, resolvedParams] = expandReferences(
@@ -74,7 +75,7 @@ export function get(path, params = {}) {
         }
       );
 
-      return prepareNextState(state, response);
+      return prepareNextState(state, response, callback);
     } catch (e) {
       throw e.body ?? e;
     }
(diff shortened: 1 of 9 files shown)


packages/commcare/src/Utils.js
```diff
@@ -53,8 +53,9 @@ export function request(configuration, path, opts) {
     headers,
     query: params,
     parseAs,
+    maxRedirections: 100,
```
josephjclark, line 56: Ah yes,  can we set this right down to  `1`?  We shouldn't get getting redirected a lot and if we are, maybe it's only good  that we start seeing errors

packages/commcare/test/index.js
```diff
@@ -97,3 +97,50 @@ describe('SubmitXls', () => {
     expect(formdata.get('create_new_cases')).to.equal('on');
   });
 });
+
+describe('getCases', () => {
+  it('should fetch cases', async () => {
+    const domain = 'my-domain';
```
josephjclark, line 103: Let's move these domain and app variables up to the top of the file, so all tests can share them

packages/commcare/test/index.js
```diff
@@ -97,3 +97,50 @@ describe('SubmitXls', () => {
...
+        path: `/a/${domain}/api/v0.5/case`,
+        method: 'GET',
+      })
+      .reply(200, req => {
+        // save out the form data that was upladed
+        method = req.method;
+        path=req.path;
+
+        // simulate a return from commcare
+        return { code: 200, message: 'success' };
```
josephjclark, line 120: Can we maybe include some sample case data here?

Like if I call `get('/case')`, where does my actual case data end up?

josephjclark:
This wasn't in the spec, but looking at the PR, I think we should make the returned data a bit easier to find.

I'll have to rely on your judgement here because it depends on what the return looks like.

In the http adaptor, when you `get()` something, the response body gets written to `state.data`, and the rest of the response stuff gets written to `state.response`. This makes it easy to get the data you wanted - which is 95% of the time - but also lets you access the original response on the occasion when you need it.

I think we should try to do the same here? Take the response payload and write it to `state.data`, and stick the other response stuff to `state.response`.

But this only works if the API  is predictable. If `/cases` returns like `{ status ,message, cases }` and `/users` returns like `{ status, message, users }` then this doesn't work

packages/commcare/test/index.js
josephjclark: Is this even what the cases data looks like? If I check [the docs ](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957300/List+Cases+version+3#JSON-Format) the json response looks very different

josephjclark:
Hi @aleksa-krolls (and maybe @mtuchi and @taylordowns2000 !) I am wondering what to do about pagination on the new `get` operation.

We currently have a nice clean API like this:

```
get('case')
```

This will call out to the `api/v0.5/case` endpoint and, by default right now, fetch 20 records. They'll be written to state.data and the `meta` object, which has your pagination data (total_count, offset, limit, next), will be written to state.response. That's all quite neat.

This means that right now users have to do their own pagination, like `get('case', { offset: 1000, limit: 1000 })`. But this is hard in job code because we have no loop construct. I don't even know how we'd write it.

I think we should automatically fetch ALL data, using the limit provided (we should default to 1000 or something I think), by continually querying until we've consumed all the data. We could disable this behaviour if the user passes in an offset (they can use offset:0 to start their own pagination).

The question should then be: do we invoke the callback for each chunk, or just once with all the data? Probably just once with all data is easiest? Then maybe later we can add an `onPage` callback or something which is passed each page.

Thoughts?

packages/commcare/src/Utils.js
```diff
@@ -24,14 +24,14 @@ export const configureAuth = (auth, headers = {}) => {
 export const prepareNextState = (state, response, callback = s => s) => {
   const { body, ...responseWithoutBody } = response;
   const nextState = {
-    ...composeNextState(state, response.body),
-    response: responseWithoutBody,
+    ...composeNextState(state, body.objects ? body.objects : response.body),
```
josephjclark, line 27: I would rewrite this as:
```
...composeNextState(state, body.objects ?? body)
```

packages/commcare/src/Adaptor.js
```diff
@@ -29,6 +29,52 @@ export function execute(...operations) {
...
+ * )
+ * // The returned objects will be written into state.data,
+ * // and the meta object will be written to state.response
+ * // along with the status code and returned headers.
+ * @function
+ * @param {string} path - Path to resource
+ * @param {Object} params - Optional request params such as limit and offset.
+ * @returns {Operation}
+ */
+export function get(path, params = {}) {
```
josephjclark, line 51: Ah, I made a mistake here -  I forgot about the callback!

The third parameter should be an optional `callback` function, just like in `openmrs`. The callback should be passed into `prepareNextState`  (which might need updating to support it)

mtuchi:
@josephjclark My [thought on pagination](https://github.com/OpenFn/adaptors/pull/563#issuecomment-2117072842).

Currently the way we do it, or the way have done it before is creating some sort of recursive workflow 
It look something like this 
`1. getCases.js`
```
get(
  "case",
  {
    query: {
      limit: 1000, //max limit: 1000
      offset:
        state.meta && state.meta.next
          ? state.meta.limit + state.meta.offset
          : 0,
    },
  },
  (state) => {
    state.meta = state.data.meta;
    return state;
  }
);
```
Edge condition to check If we should trigger the workflow again. `state.meta.next`

2a. Trigger Workflow Again | if `state.meta.next: true`
```
post('webhookUrl', { body: $.meta })
```
2b. Do something with the data 
```
//Using the data from 1st step, map and load to destination
```

That approach above is good enough, It avoids anti-pattern and it's easy to re-test the workflow with small batches
But we can do better. One solution we can try is introducing the `autoFetch` or `fetchNext` boolean option. Which will fetch the next batch until we have all records. 
Eg
```
get('case' , {'autoFetch: true}
```
The caveats on that approach if it fails after fetching couple of records already, when you try it start again from beginning

**Worth Mentioning**
There is another common pattern i use with database adaptors. If i know the total number of records that i need to fetch 
i can build an array of `offset` values. This usually allow me to use the `each()` helper function to fetch all data

Eg: Say you have build an array of `offset` values. `state.offsets = [1000, 2000, 3000]`
```
fn((state) => {
  state.offsets = [1000, 2000, 3000];
  state.allCases = [];
  return state;
});

each(
  "$.offsets[*]",
  get("case", { query: { limit: 1000, offset: $.data } }, (state) => {
    state.allCases.push(state.data);
    return state;
  })
);
```

josephjclark:
hey Professor @mtuchi that is an incredibly useful post!

We need to copy and paste this into the job writing guide.

The edge condition pattern is strong enough that I'm happy to not add an automated solution today. Maybe I'll raise an issue to add auto-fetch functionality. This is something I generally quite like but I really worry about what happens if there are a million records (then again, if there are a million records, what would you do different manually?). Also, the "onPage" callback would help process items before they're lost (and sounds like good practice). 

As an FYI to me, a failed job that still returns state is valuable if you're runching over 100 pages of 1000 records. The last failed state can still inform the next run.

packages/commcare/test/index.js
```diff
@@ -97,3 +96,74 @@ describe('SubmitXls', () => {
...
+  it('should fetch cases', async () => {
+    testServer
+      .intercept({
+        path: `/a/${domain}/api/v0.5/case`,
+        method: 'GET',
+      })
+      .reply(200, () => {
+        // simulate a return from commcare
+        return {
+          data: [
```
josephjclark, line 110: I would add a `meta` key as well and check that the returned meta gets written to the right part of state. 

The value of these tests is to ensure that, given a URL from the user and some response data from commcare, the `get` function produces a state object with the right shape.

And actually we probably should have an assertion on `state.response` - just check the statusCode or something

packages/commcare/test/index.js
```diff
@@ -97,3 +96,74 @@ describe('SubmitXls', () => {
...
+        applicationName: domain,
+        appId: app,
+        username: 'user',
+        password: 'password',
+      },
+    };
+
+    const { data } = await execute(get('case'))(state);
+
+    // The response  should be on state.data
```
josephjclark, line 131: I would remove this comment as it isn't very helpful

packages/commcare/test/index.js
josephjclark: Oh hang on this still isn't right - we need to return `{ meta, objects }` here

packages/commcare/test/index.js
```diff
@@ -97,3 +96,74 @@ describe('SubmitXls', () => {
...
+
+    expect(data.data[0]).to.haveOwnProperty('case_id');
```
josephjclark, line 133: Since we know the value of `case_id`, we may as well assert the actual value:
```
expect(data.data[0].case_id).to.equal('12345')
```

And actually before doing that I would assert on the length of state.data.
````
expect(data.length).to.equal(1)
```

packages/commcare/test/index.js
```diff
@@ -97,3 +96,74 @@ describe('SubmitXls', () => {
...
+  it('should fetch a single case', async () => {
+    testServer
+      .intercept({
+        path: `/a/${domain}/api/v0.5/case/12345`,
+        method: 'GET',
+      })
+      .reply(200, () => {
+        // simulate a return from commcare
+        return {
+          data: {
```
josephjclark, line 145: For a single case, the data we get back from commcare is like this:
```
{
  case_id: 'xxx',
  closed: false,
  closed_by: null,
  date_closed: null,
  date_modified: '2024-04-25T15:47:28.193000Z',
  domain: 'test-20190880',
  id: 'xxx,
  indexed_on: '2024-04-25T15:47:31.009517Z',
  indices: {},
  opened_by: 'xxx',
  properties: { ... },
  resource_uri: '',
  server_date_modified: '2024-04-25T15:47:28.247927Z',
  server_date_opened: '2024-04-25T15:47:28.247927Z',
  user_id: 'xxx',
  xform_ids: [ 'xxx' ]
}
```

To be clear, this `reply` function needs to return exactly the same structure as what commcare returns when you GET on `api/v0.5/case/12345`.

This function is mocking out the actual request to commcare - but our adaptor still thinks its making a GET request and getting real data back.

Happy to talk about this a bit more if you're not familiar with mocking in unit tests!

josephjclark (approved):

PR #564, opened 2024-05-17
`fhir` Update configuratio-schema

mtuchi:
## Summary

Update required properties in configuration schema and improved the `create()` example

## Details

The required properties for `fhir` adaptor were wrong, `resource` property doesn't exist and the other properties were not required. This PR update the required properties list to contain only the required items

## Review Checklist


josephjclark (approved):
Nice. Please bump versions and merge at your lesiure

Issue #565, opened 2024-05-17
`facebook` Add example token in facebook adaptor 


mtuchi:
### Description

Currently the facebook adaptor does not have an example `access_token` in it's configuration-schema. When building the sample configuration for facebook it shows the value of `access_token: false`  which is not accurate. 

### The Task
Update the `configuration-schema` to have at-least one example of `access_token` for facebook adaptor 

See screenshot below 👇🏽 
![Uploading Screenshot 2024-05-17 at 11.26.21 AM.png…]()

mtuchi:
Closing this was done in #566

PR #566, opened 2024-05-17
 Update configuration-schemas for `facebook`, `openimis` and `surveycto`

mtuchi:
## Summary

- Add an example `access_token` for facebook adaptor, 
- Fix a typo in `surveycto` adaptor 
- Fix `openimis` configuration-schema. Other properties were nested inside `baseUrl` property, I have also added `baseUrl` to required properties 

Ref #565 

## Review Checklist


packages/http/configuration-schema.json
```diff
@@ -46,5 +46,8 @@
     }
   },
   "type": "object",
-  "additionalProperties": true
-}
+  "additionalProperties": true,
+  "required": [
+    "baseUrl"
```
josephjclark, line 51: baseUrl isn't required though? :thinking:

mtuchi: I think all our adaptor should at-least have on required property because. We want to display credential based on required properties.

josephjclark: We shouldn't force required properties if they're not needed, even for the sake of consistency. It's annoying and misleading.

What do you mean "displaying the credential"? Where are we displaying it? You mean in Lightning?

mtuchi: @josephjclark i have removed the required for `http`. Please review this PR again

.changeset/rotten-rockets-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-surveycto': patch
+---
+
+Fix servername typo
```
josephjclark, line 5: Can you just qualify that this is in the configuration schema? And for facebook

packages/facebook/configuration-schema.json
```diff
@@ -7,10 +7,15 @@
       "type": "string",
       "description": "Your Facebook API access token",
       "writeOnly": true,
-      "minLength": 1
+      "minLength": 1,
+      "examples": [
+        "EAAGm0PX4ZCpsBAEHuZBwZBZCDEZDZD7ZC8ZBZBt7MbZA7sVZA...."
```
josephjclark, line 12: Two things:

1) the `...` is a bit weird, do we need it? Can we just leave a partial token? 

2) For the record, `...` is an ellipsis (from greek to leave out"). It always has _3_ docs, Not 2 dots, not 4 dots, not 27 dots. Always three.

packages/http/configuration-schema.json
mtuchi: I agree this should be fixed on `docs` side not adaptors

.changeset/rotten-rockets-cheer.md
mtuchi: Done ✅

packages/facebook/configuration-schema.json
mtuchi: The actual token was very long, let me remove the dots, Thanks for sharing about ellipsis, I was totally miss using it 🤣

mtuchi: I have removed the dot's from the token example

josephjclark: Great thanks. I think it's fine to shorten it and we don't really need to advertise that fact. It's not like anyone is going to copy our key (and if they try, tough!)

.changeset/rotten-rockets-cheer.md
mtuchi: Done ✅

josephjclark (approved):
Awesome. Please bump versions, push tags, and merge!

Issue #562, opened 2024-05-16
commcare: update configuration schema


josephjclark:
We need to make a few changes to the commcare config schema

What we call `applicationName` in our credential is what commcare calls a `domain`. We should update our configuration schema and code to match this.

This is a breaking change. We should not support the legacy applicationName unless someone comes up with a really good reason.

Also,  `appId` and `name`  should be made optional properties.

Finally, we've recently formally introduced a new property `apiKey`, which can be used in place of a password. This should be added to the schema.

A clear changeset needs to document all these changes.

aleksa-krolls:
@hunterachieng moved to Client Backlog

aleksa-krolls:
@hunterachieng moved to Client Backlog

hunterachieng:
@aleksa-krolls  I am still adding more test cases on #528 issue and might get to this later in the evening

aleksa-krolls:
@hunterachieng thanks for update, but I am going to include in next week's sprint since it sounds like there is some work to wrap up :)

hunterachieng:
@aleksa-krolls  I have moved this to code review. A PR is open and added @josephjclark  as a reviewer

hunterachieng:
@aleksa-krolls  this has been merged

PR #571, opened 2024-05-20
Feature/562 update commcare configuration

hunterachieng:
## Summary

Add apiKey and change applicationname to domain in commcare configuration schema

## Details

The adaptor currently uses `applicationName` as the `domain` name. to avoid conflicts, we are renaming `applicationName` to match commcare's definition: `domain`.
`apiKey` is also being added to the configuration schema as an optional field together with `username` and `appId` which have been made optional.

## Issues

#562 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -57,7 +57,7 @@ export function execute(...operations) {
  */
 export function get(path, params = {}, callback = s => s) {
   return async state => {
-    const { applicationName } = state.configuration;
+    const { domain } = state.configuration;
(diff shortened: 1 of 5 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -167,11 +167,11 @@ export function submit(formData) {
     const {
       // this should be called project URL.
```
josephjclark, line 168: Can we remove these two comment lines? This PR is addressing them (corretly) and leaving the comment is confusing. Thanks!

josephjclark (approved):

Issue #570, opened 2024-05-18
commcare: generic `post` function


aleksa-krolls:
## Request
Add generic `post` function. 

See docs on CommCare API endpoints we could use this with: 
https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957350/User+Creation+Mobile+Worker
https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143950677/Submission+API#Submission-as-the-body-of-a-post
https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143946023/Bulk+upload+Lookup+Tables

## To test
See LP `CommCare Demo Project` for a credential to our `test` project.

josephjclark:
Looking over those docs it seems there are a few supported formats for the post. But we also already have `submit` and `submitXls` operations.

I suggest we keep it simple and just post JSON. Users will be able to set their own headers if they wish to upload a differemt format,

So we probably want an API like:

```
post(path, data, params, callback)
```

Where data is a JSON array or object.

It should be a very simple function, like `get`, which just wraps the existing request handler.  The only automation we offer is a) a nice clean API and b) setting up auth according to `state.configuration`

aleksa-krolls:
Sounds good to me!

hunterachieng:
@aleksa-krolls  This is under code review

PR #573, opened 2024-05-21
Commcare: Implement generic `post` function

hunterachieng:
## Summary

Creating a generic `post` function for commcare

## Details

We currently have `submit` and `submitxls` which cater to submitting form-data and xls data. However, we do not have a generic post that can allow to send JSON data. 

## Issues

#570 

## Review Checklist

      migration guide followed?

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -86,8 +86,10 @@ export function get(path, params = {}, callback = s => s) {
  * Make a post request to commcare
  * @example
  * post(
- *   "idgen/identifiersource/8549f706-7e85-4c1d-9424-217d50a2988b/identifier",
- *   {}
+ *   "user",
+ *  {"username":"test",
+ *  "password":"somepassword"},
+ *   (state) =< state.data
  * );
  * @function
  * @param {string} path - Path to resource
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -83,6 +83,49 @@ export function get(path, params = {}, callback = s => s) {
 }
 
 /**
+ * Make a post request to commcare
+ * @example
+ * post(
+ *   "idgen/identifiersource/8549f706-7e85-4c1d-9424-217d50a2988b/identifier",
+ *   {}
+ * );
+ * @function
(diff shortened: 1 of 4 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -82,6 +82,49 @@ export function get(path, params = {}, callback = s => s) {
   };
 }
 
+/**
+ * Make a post request to commcare
+ * @example
+ * post(
+ *   "idgen/identifiersource/8549f706-7e85-4c1d-9424-217d50a2988b/identifier",
```
josephjclark, line 89: Is this a real url from commcare?

I'm not it's a great example, the empty object in particular isn't intuitive - is that data or options? Why is it empty either way?

User creation would seem like a simpler story:
```
post('user', (state) => state.data)
```

packages/commcare/test/index.js
```diff
@@ -320,3 +320,44 @@ describe('getCases', () => {
...
+    testServer
+      .intercept({
+        path: `/a/${domain}/api/v0.5/user`,
+        method: 'POST',
+      })
+      .reply(201, () => {
+        // simulate a return from commcare
+        return {
+       
+        id:'123456'
```
josephjclark, line 335: A minor detail but formatting is a bit quirky here. Do you have `prettier` set up?

josephjclark (changes requested):
Looks great Hunter! Nice and clean.

Just one small comment on the example really :)

packages/commcare/test/index.js
hunterachieng: Yes I have. I have fixed it

packages/commcare/src/Adaptor.js
```diff
@@ -82,6 +82,51 @@ export function get(path, params = {}, callback = s => s) {
...
 }
 
+/**
+ * Make a post request to commcare
+ * @example
+ * post(
+ *   "user",
+ *  {"username":"test",
+ *  "password":"somepassword"},
+ *   (state) =< state.data
```
josephjclark, line 92: Little typo on the arrow function

I would actually remove this though, because it complicates the example and for no particular benefit

josephjclark (approved):

Issue #577, opened 2024-05-22
`salesforce`: report better error when oath not provided for bulkQuery


josephjclark:
If using the `bulkQuery` function in salesforce, the user must include an API Version in the oauth credential.

Right now we throw a misleading error:
```
AdaptorError: The requested resource does not exist
{"message":"The requested resource does not exist","name":"AdaptorError"}
```

We should

1) Report a nice clean error if there's now apiVersion
2) Maybe default the apiVersion with a warning? Is the version tied to oauth?
2) Add some documentation around this in `bulkQuery` docs, and maybe some help text in the schema as well (api version is required for bulkQuery)

PR #579, opened 2024-05-22
`salesforce` Update default `apiVersion`

mtuchi:
## Summary

`bulkQuery` uses [Bulk API 2.0 Query](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_create_job.htm) which available in API version 47.0 and later. I have updated the default `apiVersion` to `47.0`, If apiVersion is not set.

## Implementation
- Set the default apiVersion to `47.0` 
- Throw an error in `bulkQuery` if apiVersion is less than `47.0`
- Update `bulkQuery` docs with the link to `Bulk API 2.0 Query`

Ref #577 

## Review Checklist

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -317,7 +317,9 @@ export function bulkQuery(qs, options, callback) {
       qs,
       options
     );
-    const apiVersion = connection.version;
+
+    if (connection.version < 47.0)
+      throw new Error('bulkQuery requires API version 47.0 and later');
 
     const { pollTimeout, pollInterval } = {
       ...defaultOptions,
@@ -328,7 +330,7 @@ export function bulkQuery(qs, options, callback) {
 
     const queryJob = await connection.request({
       method: 'POST',
-      url: `/services/data/v${apiVersion}/jobs/query`,
+      url: `/services/data/v${connection.version}/jobs/query`,
       body: JSON.stringify({
         operation: 'query',
         query: resolvedQs,
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -318,7 +318,7 @@ export function bulkQuery(qs, options, callback) {
       options
     );
 
-    if (connection.version < '47.0')
+    if (parseFloat(connection.version) < 47.0)
       throw new Error('bulkQuery requires API version 47.0 and later');
 
     const { pollTimeout, pollInterval } = {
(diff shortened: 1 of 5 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -317,7 +317,9 @@ export function bulkQuery(qs, options, callback) {
       qs,
       options
     );
-    const apiVersion = connection.version;
+
+    if (connection.version < 47.0)
```
josephjclark, line 321: `connection.version` returns a string, so this test likely won't work if you pass in `apiVersion: "46"`

packages/salesforce/src/Adaptor.js
```diff
@@ -709,11 +711,11 @@ function getConnection(state, options) {
...
 
   if (apiVersion && apiVersionRegex.test(apiVersion)) {
-    console.log('Using Salesforce API version', apiVersion);
+    console.log('Using Salesforce API version:', apiVersion);
     options.version = apiVersion;
   } else {
-    console.log('apiVersion is not defined');
-    console.log('We recommend using Salesforce API version 52.0 or latest');
+    options.version = '47.0';
+    console.log('Using Salesforce API version:', options.version);
```
josephjclark, line 718: A small nitpick but since the messages are the same, can we restructure like this?:
```
if (apiVersion && apiVersionRegex.test(apiVersion)) {
   options.version = apiVersion;
} else {
   options.version = '47.0';
}
console.log('Using Salesforce API version:', options.version);
```

packages/salesforce/src/Adaptor.js
mtuchi: Thanks @josephjclark, i have updated this to `connection.version < '47.0'`.

In other news i have learnt more about Javascript implicit type conversation 😀 ,
Screenshot 2024-05-23 at 1 31 24 PM

packages/salesforce/src/Adaptor.js
mtuchi: Done ✅

packages/salesforce/src/Adaptor.js
```diff
@@ -317,7 +317,9 @@ export function bulkQuery(qs, options, callback) {
       qs,
       options
     );
-    const apiVersion = connection.version;
+
+    if (connection.version < '47.0')
```
josephjclark, line 321: This won't work - you'll need to parse the version into a number

This should do it
```
if (parseFloat(connection.version) < 47)
```

mtuchi: Done ✅

josephjclark (approved):

PR #582, opened 2024-05-24
http: automatically add bearer tokens to a request

### LINKED ISSUE #581: Add OAuth2 support to http adaptor
OpenFn v2 users can now build their own OAuth2 clients in a few clicks, making it possible for other users on the platform to create OAuth2 credentials for any app super easily.

Soon, we'll have way _more_ OAuth2 credential types than we have adaptors!

I think that we should consider adding OAuth2 support to http so that folks can use a specific OAuth2 credential and a basic HTTP helper function to make interacting with new REST APIs super easy.

We might be able to do this by simply looking for `state.configuration.access_token` and using _that_ if it's present. Thoughts?

josephjclark:
## Summary

This PR reads the `token` from the config object and adds it to an `Authorization: Bearer ${token}` header.

## Details

I think this is all that's needed? It should work with oath and I think most api key-style APIs. I don't know how much variance there is in the pattern, maybe we can add more flexibility later if needed.

Note that users can provide their own Authorization header in the request, so even if this doesn't work in all cases, there's an easy work around.

## Issues

Closes #581

### DIFF
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Patch Changes
 
-- Automatically apply api/oauth tokens"
+- Automatically apply api/oauth tokens from config
 
 ## 6.2.1
 
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @openfn/language-http
 
+## 6.2.2
+
+### Patch Changes
+
+- Automatically apply api/oauth tokens"
+
 ## 6.2.1
 
 ### Patch Changes
--- a/packages/http/configuration-schema.json
+++ b/packages/http/configuration-schema.json
@@ -17,7 +17,7 @@
     "access_token": {
       "title": "Access Token",
       "type": "string",
-      "description": "Bearer token or API key",
+      "description": "OAuth Access token, API key or other Bearer token",
       "writeOnly": true,
       "examples": ["00QCjAl4MlV-WPX"]
     },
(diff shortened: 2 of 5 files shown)


taylordowns2000 (commented):
I think it expects it inside `access_token` (looking here: https://github.com/OpenFn/adaptors/blob/main/packages/salesforce/src/Adaptor.js#L744) but will wait for @mtuchi to verify.

@mtuchi , if a change needs to be made here please make it directly rather than requesting changes

josephjclark:
@taylordowns2000  `access_token` is just what the salesforce client uses to add its own oath header. I could rename it to make them consistent, but I'm not really sure it matters. Also it'll start a snake_case argument

taylordowns2000 (changes requested):
oh drat. i didn't realize this woudl include a configuration-schema change, but of course it does 😢 

sooooo..... `access_token` is different from `token` - and it looks like we wanted people to enter `token` before? why?

should we hide `access_token` from the config schema? (it should NOT show up on the credential form for HTTP). this would be my vote. make it magically discover it in the code, but don't mess with the configuration-schema (which is how we build a form UI for lightning.)

taylordowns2000:
> 
> Just to be clear - the idea is that users get an oauth JSON object from whatever tool they want (including lightning), and they add that to their state (which obviously Lightning will do), and because the config aligns with the oath standard it'll Just Work ™️

Yes! Exactly.
- We're supporting a standard OAuth response to be used as `state.configuration` per [this spec](https://www.rfc-editor.org/rfc/rfc6749#section-5.1).
- And then we're popping `access_token` in `Bearer` per this [spec](https://www.rfc-editor.org/rfc/rfc6749#section-7.1).

@josephjclark , @elias-ba , @mtuchi , it won't _always_ work, but looks like it will work for 90% of credentials created using our new standard OpenFn/Lightning OAuth2 client/credential creator.

For those edge cases, we'll need to build adaptors. (But hey, more adaptors?? That's not so bad either!)

Issue #529, opened 2024-04-25
`satusehat`: new adaptor to handle auth and expose http functions


aleksa-krolls:
## Background
[SATUSEHAT](https://satusehat.kemkes.go.id/platform/docs/id/playbook/) is the Indonesian government's FHIR health information system. A customer wants to send FHIR-format data from CommCare to SATUSEHAT, so we need to create a new adaptor to support [this project](https://github.com/OpenFn/asri).

## Request
Create a new `satusehat` adaptor that: 
1. handles authentication
2. exposes all `http` functions so that we can send requests to the dozens of API endpoints the adaptor exposes (incl. `/patient`, `/encounter`, etc.)

## To test
- Use this postman collection: https://www.postman.com/satusehat/workspace/satusehat-public/request/30206597-5a1e7e8e-8381-4854-953f-761a622cca86
- For the *staging* API credentials, see LP `HIH ASRI SATUSEHAT creds` for the `organization id`, `client key`, etc. 
- Once auth is built in the adaptor, you can trying using the adaptor to send a GET request to [this Organization endpoint](https://www.postman.com/satusehat/workspace/satusehat-public/request/30206597-5d218374-181b-465d-a2df-3ce557f165a8) (and use the `organization id` from the LP credential as the parameter `/api/Organization/:id`)

Staging login url: `https://api-satusehat-stg.dto.kemkes.go.id/oauth2/v1/accesstoken?grant_type=client_credentials`

hunterachieng:
@aleksa-krolls  I have talked to @josephjclark  and we are good to start on this task

PR #587, opened 2024-05-27
Satusehat: create satusehat adaptor

hunterachieng:
## Summary

Create a satusehat adaptor

## Details

We are implementing a new adaptor: `satusehat` with a generic `post` and `get` methods

## Issues

#529 

## Review Checklist

      migration guide followed?


packages/satusehat/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @openfn/language-template
```
josephjclark, line 1: Did you generate this with the tempate CLI? If so, we need to fix the changelog!

hunterachieng: I have changed it

josephjclark: Can we fix the template too? It's fine to do it in this PR

josephjclark (changes requested):
Hey Hunter - looks good, nice and clean!

I've got a few comments - mostly picky little things.

Neither github nor my system can read the logos - can  you check them?

packages/satusehat/CHANGELOG.md
hunterachieng: @josephjclark  I have fixed the template as well

packages/satusehat/src/Adaptor.js
```diff
@@ -0,0 +1,127 @@
...
+
+    try {
+      const response = await request(
+        state.configuration,
+        `/fhir-r4/v1/${resolvedPath}`,
+        {
+          method: 'POST',
+          data: resolvedData,
+          params: resolvedParams,
+          contentType: 'application/json',
```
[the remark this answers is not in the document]
hunterachieng: @josephjclark  I can switch this to not take in headers. I have seen majority is json

packages/satusehat/configuration-schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "baseUrl": {
```
josephjclark, line 4: We will need to add `access_token` too. Here's what googlesheets looks like:

```
    "access_token": {
      "title": "Access Token",
      "type": "string",
      "description": "Your Google Sheets access token",
      "writeOnly": true,
      "minLength": 1,
      "examples": [
        "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIng1dCI6IjlGWERwYmZNRlQyU3ZRdVhoODQ2WVR3RUlCdyIsI"
      ]
    }
```

packages/satusehat/configuration-schema.json
```diff
@@ -0,0 +1,50 @@
...
+      "title": "Base URL",
```
josephjclark, line 5: Will `baseUrl` ever be different in satusehat? Can users self-deploy or does everyone use the same satusehat domain?

At the  very least, baseUrl I think should be optional (it is currently required), defaulting to satusehat.com or whatever. But I suspect we want to remove it entirely.

packages/satusehat/configuration-schema.json
```diff
@@ -0,0 +1,50 @@
...
+      "writeOnly": true,
+      "minLength": 1,
+      "examples": [
+        "the-long-access-token-from-your-auth"
+      ]
+    }
+  },
+  "type": "object",
+  "additionalProperties": true,
+  "required": [
```
josephjclark, line 45: Ok I'm now worried that if the oauth token is set, clientId and clientSecret are optional.

I'm not actually sure how we want to handle this so I've asked in slack. This is probably fine for v1 but I want to leave this comment here for now.

packages/satusehat/src/Utils.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export async function request(configuration, path, opts) {
+  const { baseUrl, access_token } = configuration;
+
+  const { method, data, params = {}, parseAs = 'json' } = opts;
+
+  const headers = {
+    Authorization: `Bearer ${access_token}`,
+    'User-Agent': `nodejs/${process.version} @openfn/language-satusehat`,
```
josephjclark, line 74: There's a second user-agent string here.

Maybe we can add a single `generateUserAgent` function to handle this for us?

tools/generate/src/generate-adaptor.ts
```diff
@@ -49,10 +49,21 @@ const copyAndRename = async (
       await fs.mkdir(targetPath, { recursive: true });
       await copyAndRename(sourcePath, targetPath, adaptorName);
     } else if (item.isFile()) {
-      let content = await fs.readFile(sourcePath, 'utf8');
-      content = content.replace(/{{TEMPLATE}}/g, adaptorName);
-
-      await fs.writeFile(targetPath, content);
+      const ext = path.extname(item.name).toLowerCase();
+      if (
+        ext === '.png' ||
```
josephjclark, line 54: Oh good catch! That explains why the images were messed up. Nice one.

I would use a regex for this:

`if(/(png|jpe?g|gif|svg|webp)$/.test(ext)) { ... }`

The dollar means "end of string", so this is a bit like string.endsWith()`. I also snuck webp in there just for future reference.

tools/generate/src/generate-adaptor.ts
```diff
@@ -49,10 +49,21 @@ const copyAndRename = async (
...
+        ext === '.jpeg' ||
+        ext === '.gif' ||
+        ext === '.svg'
+      ) {
+        const content = await fs.readFile(sourcePath);
+        await fs.writeFile(targetPath, content);
+      } else {
+        let content = await fs.readFile(sourcePath, 'utf8');
+        content = content.replace(/{{TEMPLATE}}/g, adaptorName);
+        await fs.writeFile(targetPath, content);
```
josephjclark, line 65: Because the `writeFile` line is the same, I would declare content as a `let` outside the if/else block, read and maybe transform the content INSIDE the blocks, and then write the file AFTER the if/else. So that you're not repeating any code :sunglasses:

josephjclark (commented):
A few more little things. The logos need to be Satusehat logos if we can get them, not openfn ones

packages/satusehat/src/Utils.js
```diff
@@ -3,7 +3,7 @@ import {
   request as commonRequest,
   logResponse,
 } from '@openfn/language-common/util';
-import pkg from '../package.json'  assert { type: 'json' };
+import pkg from '../package.json';
```
josephjclark, line 6: I'm confused by this commit! I expect this to break the adaptor, and  certainly unit tests are failing right now

packages/satusehat/configuration-schema.json
josephjclark: Ok, there are at least two satusehat APIs - the main one and the sandbox. So we do need baseURL.

Let's leave it as required for the time being, to ensure that users intentioanlly set the right server

josephjclark:
Looking great Hunter!

One last thing. Looking at the `satusehat` docs, they also use PUT and for some reason PATCH requests. I don't think they use DELETE.

So we need to add `put()` and `patch()`.

To reduce the amount of repetition, can we first add a `request()` operation, an then make `get`, `post` ,`put`  and `patch` very thin wrappers around it:

This untested code, but this is what I mean:
```
// Change the utils import so that we don't clash with `request`
import util from './Utils';

// GENERIC REQUEST OPERATION (can be used in job code)
export function request(method, data = null, path, params = {}, callback = s => s) {
  return async state => {
    const [resolvedPath, resolvedData, resolvedParams] = expandReferences(
      state,
      path,
      params
    );
    try {
      const response = await util.request( // note that I am using util. here
        state.configuration,
        `/${resolvedPath}`,
        {
          method
          params: resolvedParams,
          data: resolvedData // NOTE - you might have to only add the data key if data is truthy. Try this first.
        }
      );

      return util.prepareNextState(state, response, callback); // note that I am using util. here
    } catch (e) {
      throw e.body ?? e;
    }
  };
}

// WRAPPED GET OPERATION
    return request('get', path, null, params, callback)
}
```

packages/satusehat/test/Adaptor.test.js
```diff
@@ -0,0 +1,523 @@
...
+      },
+      references: [],
+      data: null,
+    });
+  });
+});
+
+describe('getOrganizations', () => {
+  beforeEach(() => {
+    testServer
```
josephjclark, line 75: You shouldn't need to `.persist()` AND `beforeEach()`. 

You should just be able to set up a persistent interceptor at the top of the file (perhaps in a `before()` block) and use it in all the tests

packages/satusehat/test/Adaptor.test.js
```diff
@@ -0,0 +1,523 @@
...
+      .intercept({
+        path: `/fhir-r4/v1/Organization`,
+        method: 'GET',
+        query: {
+          name: 'somename',
+        },
+      })
+      .reply(200, () => {
+        // simulate a return from satusehat
```
josephjclark, line 99: We don't need this comment on every interceptor - please remove them all

josephjclark (approved):
This is great, thanks @hunterachieng !

PR #588, opened 2024-05-27
salesforce: Ensure anyscii is loaded before executing

josephjclark:
It's theoretically possible that `any-ascii` has not loaded before `toUTF8` is called.

This seems vanishingly unlikely since we have to call out to the salesforce server before we start executing. Can we really handshake with Salesforce quicker than we can load a Javascript Module? 

But Mutchi is seeing an runtime error  where `anyAscii is not a function`. If I locally disable the salesforce connection code and just call `toUTF`, then it fails the same way. So it feels like somehow the module loader is losing the race :shrug: 

Anyway this is a more robust solution and I wish I'd thought of this earlier.

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -24,11 +24,11 @@ import { expandReferences as newExpandReferences } from '@openfn/language-common
 import jsforce from 'jsforce';
 import flatten from 'lodash/flatten';
 
-// use a dynamic import because any-ascii is pure ESM and doesn't play well with CJS
-// Note that technically we should await this, but in practice the module will be loaded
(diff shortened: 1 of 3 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -792,6 +792,7 @@ export function execute(...operations) {
     // Note: we no longer need `steps` anymore since `commonExecute`
     // takes each operation as an argument.
     return commonExecute(
+      loadAnyAscii,
```
mtuchi, line 795: @josephjclark where are we using this loadAnyAscii ?

mtuchi (approved):

Issue #552, opened 2024-05-08
`salesforce` rename `create` to `insert`


mtuchi:
### Description

Salesforce `create()` function should be renamed as `insert()` because that is what the DB operation is called in salesforce. 

Is there a way to make this change but also ensure backwards compatibility with jobs that use `create()`?

### Context 
When we use the word “create” the user thinks we’re creating metadata, instead of “inserting” data into an existing `table/object`

mtuchi:
According to [jsforce](https://jsforce.github.io/document/#create), create is synonym of `insert`.  The recommended approach is we add an alias for `insert` which is similar to `create`.

<img width="763" alt="Screenshot 2024-05-30 at 5 01 44 PM" src="https://github.com/OpenFn/adaptors/assets/6592749/c581f813-6034-41cd-88bf-b090565eeb40">

Another recommendation is we improve the documentation to clearly state `create()` function. Only creates new sObject record. Does not create sObject metadata

josephjclark:
Sorry @mtuchi I was supposed to leave a comment on this - got distracted

That's a useful bit of documentation. If `create` and `insert`  are synonyms then I'm quite happy to alias it in the adaptor. Doing anything more dramatic that needs some serious consideration.

josephjclark:
Done in #632

Issue #537, opened 2024-04-29
Update salesforce examples 


mtuchi:
#### Description 
Update salesforce jsdoc examples to use latest best practice

PR #590, opened 2024-05-30
Add `insert` function and improve jsdocs

mtuchi:
## Summary

Add a `insert()` function which is an alias for `create()`. I have also made improvements on functions examples.

## Notes
- Add a `insert()` function which is an alias for `create()`.  #552 
- Improve jsdocs for `query`, `bulk`, `describe`. #537 
- Include the Salesforce query response in the result, even if no records are found.

## Review Checklist

### DIFF
--- a/.changeset/many-maps-lay.md
+++ b/.changeset/many-maps-lay.md
@@ -4,3 +4,4 @@
 
 - Add `insert()` function which is an alias for `create()`.
 - Improve jsdocs for query, bulk, describe, create, upsert
+- Deprecated `upsertIf()` and `createIf()`
--- a/.changeset/many-maps-lay.md
+++ b/.changeset/many-maps-lay.md
@@ -2,6 +2,15 @@
 '@openfn/language-salesforce': minor
 ---
 
-- Add `insert()` function which is an alias for `create()`.
-- Improve jsdocs for query, bulk, describe, create, upsert
-- Deprecated `upsertIf()` and `createIf()`
+### Added
+
+- `insert()` function as an alias for `create()`.
+
+### Improved
+
+- JSDocs for `query`, `bulk`, `describe`, `create`, and `upsert`.
+
+### Deprecated
+
+- `upsertIf()` and `createIf()` functions are now deprecated. Use
+  `fnIf(condition, upsert())` instead.
--- a/.changeset/many-maps-lay.md
+++ b/.changeset/many-maps-lay.md
@@ -4,5 +4,3 @@
 
 - Add `insert()` function which is an alias for `create()`.
 - Improve jsdocs for query, bulk, describe, create, upsert
-- Include the Salesforce query response in the result, even if no records are
-  found.
--- a/.changeset/many-maps-lay.md
+++ b/.changeset/many-maps-lay.md
@@ -1,0 +1,8 @@
+---
+'@openfn/language-salesforce': minor
+---
+
+- Add `insert()` function which is an alias for `create()`.
+- Improve jsdocs for query, bulk, describe, create, upsert
+- Include the Salesforce query response in the result, even if no records are
+  found.
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -912,6 +912,7 @@ export {
   field,
   fields,
   fn,
+  fnIf,
   http,
   humanProper,
   index,
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -585,7 +585,8 @@ export function insert(sObject, attrs) {
 
 /**
  * Create a new sObject if conditions are met.
- * @deprecated
+ *
+ * > The `createIf()` function has been deprecated. Use `fnIf(condition,create())` instead.
  * @public
  * @example
  * createIf(true, 'obj_name', {
@@ -673,7 +674,8 @@ export function upsert(sObject, externalId, attrs) {
 
 /**
  * Conditionally create a new sObject record, or updates it if it already exists
- * @deprecated
+ *
+ * > The `upsertIf()` function has been deprecated. Use `fnIf(condition,upsert())` instead.
  * @public
  * @example
  * upsertIf(true, 'obj_name', 'ext_id', {
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -82,7 +82,7 @@ export function describeAll() {
 }
 
 /**
- * Outputs basic information about an sObject to `STDOUT`.
+ * Returns basic information about an sObject.
  * @public
  * @example
  * describe('obj_name')
@@ -160,7 +160,7 @@ export function retrieve(sObject, id, callback) {
  * @function
  * @param {String} qs - A query string. Must be less than `4000` characters in WHERE clause
  * @param {Object} options - Options passed to the bulk api.
- * @param {boolean} [options.autoFetch=false] - Fetch next records if available.
+ * @param {Boolean} [options.autoFetch=false] - Fetch next records if available.
  * @param {Function} callback - A callback to execute once the record is retrieved
  * @returns {Operation}
  */
@@ -307,8 +307,8 @@ const defaultOptions = {
  * @function
  * @param {String} qs - A query string.
  * @param {Object} options - Options passed to the bulk api.
- * @param {integer} [options.pollTimeout=90000] - Polling timeout in milliseconds.
- * @param {integer} [options.pollInterval=3000] - Polling interval in milliseconds.
+ * @param {Integer} [options.pollTimeout=90000] - Polling timeout in milliseconds.
+ * @param {Integer} [options.pollInterval=3000] - Polling interval in milliseconds.
  * @param {Function} callback - A callback to execute once the record is retrieved
  * @returns {Operation}
  */
@@ -364,16 +364,29 @@ export function bulkQuery(qs, options, callback) {
  * Create and execute a bulk job.
  * @public
  * @example
- * bulk('Patient__c', 'insert', { failOnError: true, pollInterval: 3000, pollTimeout: 240000 }, state => {
- *   return state.data.someArray.map(x => {
- *     return { 'Age__c': x.age, 'Name': x.name }
- *   })
- * });
+ * bulk(
+ *   "Patient__c",
+ *   "insert",
+ *   { failOnError: true, pollInterval: 3000, pollTimeout: 240000 },
+ *   (state) => state.someArray.map((x) => ({ Age__c: x.age, Name: x.name }))
+ * );
+ * @example <caption>Bulk upsert</caption>
+ * bulk(
+ *   "vera__Beneficiary__c",
+ *   "upsert",
+ *   { extIdField: "vera__External_ID__c", failOnError: true },
+ *   (state) => state.data
+ * );
  * @function
  * @param {String} sObject - API name of the sObject.
- * @param {String} operation - The bulk operation to be performed
+ * @param {String} operation - The bulk operation to be performed.Eg "insert" | "update" | "upsert"
  * @param {Object} options - Options passed to the bulk api.
- * @param {Function} records - an array of records, or a function which returns an array.
+ * @param {Integer} [options.pollTimeout=240000] - Polling timeout in milliseconds.
+ * @param {Integer} [options.pollInterval=6000] - Polling interval in milliseconds.
+ * @param {String} [options.extIdField] - External id field.
+ * @param {Boolean} [options.failOnError] - Fail the operation on error.
+ * @param {Boolean} [options.allowNoOp] - For skipping operation.
+ * @param {Array} records - an array of records, or a function which returns an array.
  * @returns {Operation}
  */
 export function bulk(sObject, operation, options, records) {
@@ -511,7 +524,7 @@ export function destroy(sObject, attrs, options) {
 }
 
 /**
- * Create a new object.
+ * Create a new sObject record(s).
  * @public
  * @example
  * create('obj_name', {
@@ -520,7 +533,7 @@ export function destroy(sObject, attrs, options) {
  * })
  * @function
  * @param {String} sObject - API name of the sObject.
- * @param {Object} attrs - Field attributes for the new object.
+ * @param {Object} attrs - Field attributes for the new record.
  * @returns {Operation}
  */
 export function create(sObject, attrs) {
@@ -540,7 +553,24 @@ export function create(sObject, attrs) {
 }
 
 /**
- * Create a new object if conditions are met.
+ * Synonym for "create(sObject, attrs)".
+ * @public
+ * @example
+ * insert('obj_name', {
+ *   attr1: "foo",
+ *   attr2: "bar"
+ * })
+ * @function
+ * @param {String} sObject - API name of the sObject.
+ * @param {Object} attrs - Field attributes for the new record.
+ * @returns {Operation}
+ */
+export function insert(sObject, attrs) {
+  return create(sObject, attrs);
+}
+
+/**
+ * Create a new sObject if conditions are met.
  * @public
  * @example
  * createIf(true, 'obj_name', {
@@ -548,7 +578,7 @@ export function create(sObject, attrs) {
  *   attr2: "bar"
  * })
  * @function
- * @param {boolean} logical - a logical statement that will be evaluated.
+ * @param {Boolean} logical - a logical statement that will be evaluated.
  * @param {String} sObject - API name of the sObject.
  * @param {Object} attrs - Field attributes for the new object.
  * @returns {Operation}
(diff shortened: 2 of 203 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -82,7 +82,7 @@ export function describeAll() {
 }
 
 /**
- * Outputs basic information about an sObject to `STDOUT`.
+ * Returns basic information about an sObject.
```
josephjclark, line 85: `describe` and `describeAll` need to be consistent - they both have very similar signatures and descriptions.

I agree that "output to stdout" is not very useful. But `returns` isn't very accurate either.

This is one of those strange apis that doesn't write to state.data, but dumps its data in state.references.

If we're gonna make a change, I would like this (and `describeAll()`) to say: `Prints a description of an sObject and pushes the result to state.references`

packages/salesforce/src/Adaptor.js
```diff
@@ -306,8 +307,8 @@ const defaultOptions = {
  * @function
  * @param {String} qs - A query string.
  * @param {Object} options - Options passed to the bulk api.
- * @param {integer} [options.pollTimeout=90000] - Polling timeout in milliseconds.
- * @param {integer} [options.pollInterval=3000] - Polling interval in milliseconds.
+ * @param {Integer} [options.pollTimeout=90000] - Polling timeout in milliseconds.
```
josephjclark, line 310: Why the case change here? I would expect a primitive `integer` to be lower case.  In the case of `Boolean`, `Boolean` is a class name in Javascript - so do we mean `new Boolean(true)` or `true`?

I would set them all to lowercase unless there's a compelling reason for the change

packages/salesforce/src/Adaptor.js
```diff
@@ -363,16 +364,29 @@ export function bulkQuery(qs, options, callback) {
...
+ *   "insert",
+ *   { failOnError: true, pollInterval: 3000, pollTimeout: 240000 },
+ *   (state) => state.someArray.map((x) => ({ Age__c: x.age, Name: x.name }))
+ * );
+ * @example <caption>Bulk upsert</caption>
+ * bulk(
+ *   "vera__Beneficiary__c",
+ *   "upsert",
+ *   { extIdField: "vera__External_ID__c", failOnError: true },
+ *   (state) => state.data
```
josephjclark, line 378: There's nothing in this callback, what's the purpose of including it in this example?

I think this is a consistent problem in our examples - the inclusion of irrelevant or even misleading information.

We don't have to address this now, but for future reference a good set of examples would be `simple bulk insert`, `upsert with custom options`, `bulk operation with callback`, etc, where each has the minimum possible amount of syntax.

packages/salesforce/src/Adaptor.js
```diff
@@ -510,7 +524,7 @@ export function destroy(sObject, attrs, options) {
 }
 
 /**
- * Create a new object.
+ * Create a new sObject record(s).
```
josephjclark, line 527: Wait, this can do multiple records?

Is this confirmed behaviour? I see the `jsforce` supports  an array. I guess we just pass whatever through?

Basically if you're gonna drop a bombshell like `(s)` it needs documenting properly somewhere.

packages/salesforce/src/Adaptor.js
```diff
@@ -579,7 +610,8 @@ export function createIf(logical, sObject, attrs) {
 }
 
 /**
- * Upsert an object.
+ * Upsert an sObject.
```
josephjclark, line 613: If you delete the phrase `Upsert an sObject.`, these docs become 200% better

I would perhaps phrase it as

`Create a new sObject record, or updates it if it already exists`

josephjclark (changes requested):

packages/salesforce/src/Adaptor.js
mtuchi: I wanted consistency, i think it should be consistency for all adaptor. I will change to lower case

mtuchi: I have rename all params type to be lower case ✅ , with exception of `RequestOptions`

packages/salesforce/src/Adaptor.js
mtuchi: I am glad you mentioned that, it turns out you can create a single record or multiple records using `create, update or upsert` function. So i have update the documentation accordingly. Please have another look

packages/salesforce/src/Adaptor.js
mtuchi: Thanks this done ✅

packages/salesforce/src/Adaptor.js
mtuchi: Your right it's clear what is `state.data`, Is it an object, string or an array. 
I have updated the example to use an array of objects

packages/salesforce/src/Adaptor.js
josephjclark: Perfect!

packages/salesforce/src/Adaptor.js
josephjclark: Good. Any class-like type should be ProperCased, but a primitive should be lower case. I think that's what we usually do, but I'm sure there are inconsistencies. Not a big deal!

josephjclark: Ah, wait, Operation should be capital O

mtuchi: @josephjclark so for `Object` and `Array` they should be ProperCased right ?

packages/salesforce/src/Adaptor.js
```diff
@@ -548,9 +584,9 @@ export function create(sObject, attrs) {
  * })
  * @function
  * @param {boolean} logical - a logical statement that will be evaluated.
- * @param {String} sObject - API name of the sObject.
- * @param {Object} attrs - Field attributes for the new object.
- * @returns {Operation}
+ * @param {string} sObject - API name of the sObject.
+ * @param {(object|object[])} attrs -  The data to process, either a single object or an array of objects.
```
josephjclark, line 588: I don't think the brackets `()` are needed? Do they have meaning?

I don't know how we handle multiple types in other adaptors. I'm not at my desk today, can you check?

This isn't a formal declaration so we could say `object or array of objects`. I'd even be happy with `object or array`

mtuchi: @josephjclark the `()` is used to group union types. See jsdoc here https://jsdoc.app/tags-type

mtuchi: I have updated the description to say `object or array of objects` because it's an array we are expecting it to be an array of objects

packages/salesforce/src/Adaptor.js
```diff
@@ -58,12 +58,12 @@ export function relationship(relationshipName, externalId, dataSource) {
 }
 
 /**
- * Outputs basic information about available sObjects.
+ * Prints a total number of all available sObjects and pushes the result to state.references
```
josephjclark, line 61: Should be `prints THE total number...`.  Also there's no trailing full stop

packages/salesforce/src/Adaptor.js
```diff
@@ -58,12 +58,12 @@ export function relationship(relationshipName, externalId, dataSource) {
...
  * @public
  * @example
  * describeAll()
  * @function
- * @returns {Operation}
+ * @returns {operation}
```
josephjclark, line 66: Operation needs to be ProperCased

packages/salesforce/src/Adaptor.js
```diff
@@ -1,13 +1,13 @@
...
  * @property {object} data JSON Data.
  * @property {Array<Reference>} references History of all previous operations.
  * @ignore
  */
 
 /**
- * @typedef {Function} Operation
- * @param {State} state
+ * @typedef {function} Operation
+ * @param {state} state
```
josephjclark, line 10: State needs to be ProperCased

packages/salesforce/src/Adaptor.js
josephjclark: All I said is that `Operation` should be have a captital O, because there's a specific type interface to an Operation. It has a formal definition.

For the record, object should probably be lowercase because it's an informal definition (we provide no type interface for the shape of the object, it's just an object).

Array can go both ways because you can use the term informally. For consistency's sake we should probably capitalise it.

I'm starting to  formalise some of this stuff into a style guide. Maybe it'll help encourage consistency (eventually)

packages/salesforce/src/Adaptor.js
```diff
@@ -363,17 +364,37 @@ export function bulkQuery(qs, options, callback) {
...
+ *   (state) => state.someArray.map((x) => ({ Age__c: x.age, Name: x.name }))
+ * );
+ * @example <caption>Bulk upsert</caption>
+ * bulk(
+ *   "vera__Beneficiary__c",
+ *   "upsert",
+ *   { extIdField: "vera__External_ID__c", failOnError: true },
+ *   [
+ *     {
+ *       vera__Reporting_Period__c: 2023,
```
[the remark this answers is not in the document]
mtuchi: Yes i have updated the example to use the correct `extIdField: vera__Result_UID__c`

packages/salesforce/src/Adaptor.js
```diff
@@ -363,17 +364,37 @@ export function bulkQuery(qs, options, callback) {
  * Create and execute a bulk job.
  * @public
  * @example
- * bulk('Patient__c', 'insert', { failOnError: true, pollInterval: 3000, pollTimeout: 240000 }, state => {
- *   return state.data.someArray.map(x => {
- *     return { 'Age__c': x.age, 'Name': x.name }
- *   })
- * });
+ * bulk(
```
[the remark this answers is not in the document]
mtuchi: I have added the captions for both insert and upsert operations

packages/salesforce/src/Adaptor.js
```diff
@@ -363,17 +364,37 @@ export function bulkQuery(qs, options, callback) {
...
+ *   "Patient__c",
+ *   "insert",
+ *   { failOnError: true, pollInterval: 3000, pollTimeout: 240000 },
```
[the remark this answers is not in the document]
mtuchi: Probably only the `failOnError: true` option,

packages/salesforce/src/Adaptor.js
```diff
@@ -363,17 +364,37 @@ export function bulkQuery(qs, options, callback) {
...
- * @param {Function} records - an array of records, or a function which returns an array.
- * @returns {Operation}
+ * @param {string} sObject - API name of the sObject.
+ * @param {string} operation - The bulk operation to be performed.Eg "insert" | "update" | "upsert"
+ * @param {object} options - Options passed to the bulk api.
+ * @param {integer} [options.pollTimeout=240000] - Polling timeout in milliseconds.
+ * @param {integer} [options.pollInterval=6000] - Polling interval in milliseconds.
+ * @param {string} [options.extIdField] - External id field.
+ * @param {boolean} [options.failOnError] - Fail the operation on error.
+ * @param {boolean} [options.allowNoOp] - For skipping operation.
```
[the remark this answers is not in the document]
mtuchi: From what i have understood `allowNoOp` is basically for skipping the operation if we don't have any records for the operation. This smells like a problem that can be solved with `fnIf()`

packages/salesforce/src/Adaptor.js
```diff
@@ -363,17 +364,37 @@ export function bulkQuery(qs, options, callback) {
...
- * @param {Object} options - Options passed to the bulk api.
- * @param {Function} records - an array of records, or a function which returns an array.
- * @returns {Operation}
+ * @param {string} sObject - API name of the sObject.
+ * @param {string} operation - The bulk operation to be performed.Eg "insert" | "update" | "upsert"
+ * @param {object} options - Options passed to the bulk api.
+ * @param {integer} [options.pollTimeout=240000] - Polling timeout in milliseconds.
+ * @param {integer} [options.pollInterval=6000] - Polling interval in milliseconds.
+ * @param {string} [options.extIdField] - External id field.
+ * @param {boolean} [options.failOnError] - Fail the operation on error.
```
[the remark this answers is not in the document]
mtuchi: for `allowNoOp` is false and for `failOnError` is false as well. I have updated the docs

packages/salesforce/src/Adaptor.js
```diff
@@ -549,8 +594,8 @@ export function create(sObject, attrs) {
  * })
  * @function
  * @param {boolean} logical - a logical statement that will be evaluated.
- * @param {String} sObject - API name of the sObject.
- * @param {Object} attrs - Field attributes for the new object.
+ * @param {string} sObject - API name of the sObject.
+ * @param {(object|object[])} attrs - Field attributes for the new object.
  * @returns {Operation}
  */
 export function createIf(logical, sObject, attrs) {
```
josephjclark, line 601: Hey, now that `fnIf` is out, want to be brave and deprecate this?

We won't remove it and its not a major change. But in the docs we should recommend people use `fnIf` instead, and add a release note to the changelog

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Right, so we send that flag to salesforce and salesforce won't throw an error if we send an empty array?

I agree that fnif is a better solution.

Let's remove this documentation - we never documented this option before and, if you know the option from jsforce, it will continue to work. But let's not call attention to it.

josephjclark (changes requested):
Just one more thing that I can see - let's softly deprecate `createIf` and `upsertIf`

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: It's not a jsforce thingy or salesforce option. It's specific for this adaptor function. 
If i am to translate it using `fnIf` it will be something like this 
```
fnIf(
  $.records.length > 0,
  bulk(
    "vera__Beneficiary__c",
    "upsert",
    { extIdField: "vera__Result_UID__c" },
    $.records
  )
);
```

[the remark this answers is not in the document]
mtuchi: @josephjclark removed `allowNoOp` docs

.changeset/many-maps-lay.md
```diff
@@ -0,0 +1,7 @@
+---
+'@openfn/language-salesforce': minor
+---
+
+- Add `insert()` function which is an alias for `create()`.
+- Improve jsdocs for query, bulk, describe, create, upsert
+- Deprecated `upsertIf()` and `createIf()`
```
josephjclark, line 7: this note should show an example of usage with fnIf

packages/salesforce/src/Adaptor.js
```diff
@@ -629,15 +680,19 @@ export function upsert(sObject, externalId, attrs) {
...
+ * @param {string} sObject - API name of the sObject.
+ * @param {string} externalId - ID.
+ * @param {(object|object[])} attrs - Field attributes for the new object.
  * @returns {Operation}
  */
 export function upsertIf(logical, sObject, externalId, attrs) {
   return state => {
     const resolvedLogical = expandReferences(logical)(state);
 
+    console.warn(
```
josephjclark, line 692: hmm, I was thinking of something softer than this. Like adding an `@deprecated` tag. I wonder how the docsite will deal with that? I'll check.

We certainly need to add a clear line to the public documentation, even if we can't use @deprecated and even if we keep the logged warning.

josephjclark: Hey actually you could add @deprecated, run `pnpm docs:build`, and load the docsite from the `generate-adaptors-devx` branch. That should let you see the changes! That's what I'll be trying this afternoon, but it would be great to test it from your machine

mtuchi: @josephjclark This is how the docs look like 
Screenshot 2024-06-14 at 3 14 38 PM

.changeset/many-maps-lay.md
mtuchi: Updated the changeset

packages/salesforce/src/Adaptor.js
josephjclark: Hey how are you in dark mode?! I never tried that 😂 Looks neat

Anyway the deprecated thing looks terrible. Hmm. Let's just add a simple string to the docs then

mtuchi: It's vscode markdown view 😆 , here is how it looks in docs site 
Screenshot 2024-06-14 at 3 47 15 PM

mtuchi: Is this better ?
Screenshot 2024-06-14 at 3 51 58 PM

josephjclark: Better - instead of quoting it, can we make it all bold?

josephjclark: And docs.openfn.org org does come with dark mode - I just never switched it on for some reason 
image

josephjclark: VSC's table stying is way nicer...

mtuchi: I have change from quote to bold @josephjclark

josephjclark (approved):
Thanks @mtuchi, looks good

Issue #583, opened 2024-05-24
Update adaptor template


josephjclark:
It's time to review and update the adaptor template, again!

We currently have a very strong, repetitive pattern for new adaptors:

* Wrap `common.request` with auth rules
* Set some default/base url and maybe provide a url builder
* Expose get/put/post/delet/request operations
* Generate a test job

We need to update the template to reflect this pattern. No custom high-level helper  functions - just a basic HTTP layer.

After the command has run, we need to print a list of things to do, like:

```
<adaptor-name> adaptor generated!

Next steps:
- update the base URL in src/Utils.request()
- Check that the authorisation function in src/Utils is doing the right thing
- Set sensible defaults in utils.request for content types and response handling
- check the configuration schema defines the right values for auth
- Carefully review all generated files. Remove or comment out any code you don't need
- Update the images in assets with new  256x256 and 215x190 logos
- Consider updating the changelog to describe the first release to users
- Consider adding new unit tests. Use the templates provided and add realistic URLs and response data from your data source
- Create a test job
 - Update the readme with basic docs
- Test your new adaptor againt the test job with `pnpm build` and `openfn -ma <Adaptor-name>
```

Plus anything else. It's quite a long list init? Perhaps we can also add a command like `generate --reminder` which prints the same todo list (just generaically)

Template docs and comments need to direct the user to the bits they need to edit.

OAuth and basic auth patterns need to be supported as standard.

The new template should mean that when plugging into a new service, the user has to do really very little work. Maybe set the base url.

We should consider adding an oauth helper function like we're doing for `satusehat` - maybe a commented function or something that the user can opt into if needed

josephjclark:
I'm sure we can also reduce the list of devdependencies per adaptor (using the monorepo root instead)

josephjclark:
Done in https://github.com/OpenFn/adaptors/pull/601/files

PR #594, opened 2024-06-03
Better template

josephjclark:
This PR updates the template to optimise around new-school http connectivity

Changes:

* Add `request`, `get` and `post` operations to Adaptor.js
* Add `request` to `utils`
* Remove execute by default (its just confusing)
* Update basic tests to match

Closes https://github.com/OpenFn/adaptors/issues/583


josephjclark:
I've removed execute from the template but @hunterachieng points out that any oauth enabled adaptor will need to use execute.

Maybe the template should be for an oauth adaptor and can trap all that boilerplate.

Issue #534, opened 2024-04-28
Implement OpenFn adaptor for ODK


christad92:
Using the ODK API reference, we want to provide an ODK adaptor on OpenFn with the following capability: 

- [ ] Authenticate using email and password to generate a session token which will be used to authorize other actions
- [ ] Include HTTP for generic `get` and `post` methods
- [ ] Helper function `getForms` specifically for the [list forms endpoint](https://docs.getodk.org/central-api-form-management/#list-all-forms): `GET /v1/projects/{projectId}/forms`
- [ ] Helper function `getSubmissions` specifically for the [list forms endpoint](https://docs.getodk.org/central-api-submission-management/#listing-all-submissions-on-a-form): `GET /v1/projects/{projectId}/forms/{xmlFormId}/submissions`

FYI Taylor really wants to be able to demo `getSubmissions` to our ODK contacts. 

API  Docs: https://docs.getodk.org/central-api-authentication/
Access: LastPass "ODK Sandbox" (ask @taylor if you need help getting in!)
About ODK: https://getodk.org/

Note that there is already a form with a submission in there.

Potentially useful sample code:
```
const options = {
  url: `${bundle.authData.url}/v1/projects/${bundle.inputData.projectId}/forms/${bundle.inputData.xmlFormId}.svc/${bundle.inputData.table}`,
  method: 'GET',
  headers: {
    'Content-Type': 'application/json',
    'Authorization': `Bearer ${bundle.authData.token}`
  },
  params: {
      '$top': bundle.inputData.top
  },
}
```

## Client considerations
Taylor wants this for an upcoming demo, but no set timeline or major urgency here. This adaptor will probably look very similar to our CommCare and Kobo adaptors, because this is a very similar mobile data collection app. 

## Toggl
`adaptors`

aleksa-krolls:
@christad92 Confirming we can add this ODK adaptor for Hunter's sprint next week

hunterachieng:
@taylordowns2000  I am unable to see `ODK Sandbox` creds on LP. I am only seeing `Agros ODK Central` cc @aleksa-krolls

aleksa-krolls:
@hunterachieng I just shared these creds with you on LP. Try again, and ping me on Slack if you're still having issues/ 
cc @taylordowns2000

hunterachieng:
@josephjclark  ODK adaptor is ready for review. Getting the correct images from the internet is a bit tricky but still trying for the larger sized one.
[...]

Issue #578, opened 2024-05-22
Build URLs a bit more safely


josephjclark:
This issue affects many adaptors (especially some of our new ones!)

When building URLs, we often use patterns like `www.blah.com/${appId}/${path}`, where `path` is a user defined value.

But should path start with a `/`? Or end with one?

In practice it shouldn't matter, and I think usually it doesn't matter. But we should take care to build well-formed URLs which are robust to leading and trailing slashes from users.

I thought we could use the URL global object for this but on checking the API I'm not so sure. We might be better off using `node:path.resolve(a,b,c)` to build the path

See `commcare` for an example of this.

josephjclark:
This has bitten TD a couple of times in the last couple of days - we really shouldn't be so brittle

josephjclark:
Using undici's baseUrl is a pattern we should use more - undici will deal with at least the base slash. That means adaptors need to use baseUrl rather than assembling their own url. The template needs to do this too

josephjclark:
`path.join` is the answer here:
```
import nodepath from 'node:path'
...
const safePath = nodepath.join(state.configuration.version, resource, path)
```

This will sort out all the leading and trailing slashes for us

josephjclark:
The node path pattern is working really well.

I'm starting to phase this into adaptors and a lot of fixes have been made in #683

This issue was raised as a reminder and tracker. I'm satisfied we have a solution and it's being rolled out as we go. So I'm closing it down.

PR #601, opened 2024-06-05
ODK: Create odk adaptor

hunterachieng:
## Summary

Creating an ODK adaptor

## Details

We have implemented a new adaptor `odk` that contains the `get`, `post`, `getForms`, and `getSubmissions` functions.

To test, ensure you have updated your configuration with the following:

`baseUrl`

`username`

`password`

## Issues

#534 

## Review Checklist

      migration guide followed?


packages/odk/src/Adaptor.js
```diff
@@ -0,0 +1,148 @@
...
+ * @param {string} xmlFormId - Id of the form you want to get its submissions
+ * @param {Object} params - Optional request params
+ * @param {Object} headers - Optional request headers
+ * @param {function} [callback] - Optional callback to handle the response
+ * @returns {Operation}
+ */
+export function getSubmissions(
+  projectId,
+  xmlFormId,
+  params,
```
josephjclark, line 77: I don't think we should support params and headers here. The whole idea if this function is to abstract out the http layer. The function should add whatever params and headers are needed

josephjclark: This is also true on `getForms`

packages/odk/src/Adaptor.js
```diff
@@ -0,0 +1,148 @@
...
+ */
+export function post(path, data, params, callback) {
+  return request('POST', path, data, params, callback);
+}
+
+/**
+ * Make a general HTTP request
+ * @example
+ * request("POST", 'v1/projects', { name: 'Project Name' });
+ * @function
```
josephjclark, line 104: Ah we need to add `@public` here. That's my fault - can you also update the the template so that the default operations have `@public`?

packages/odk/src/Utils.js
```diff
@@ -0,0 +1,95 @@
+/**
```
josephjclark, line 1: We should remove this comment from the actual adaptor (it's correct to leave it in the template!)

packages/odk/src/Utils.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const nextState = {
+    ...composeNextState(state, response.body),
+    response: responseWithoutBody,
+  };
+
+  return callback(nextState);
+};
+
+function removeLeadingSlash(url) {
```
josephjclark, line 58: I've love to find a better solution to this. It's tracked in https://github.com/OpenFn/adaptors/issues/578

This is fine as it for now - I just want to link the issue to this PR.

packages/odk/src/Utils.js
```diff
@@ -0,0 +1,95 @@
...
+export const request = (
+  configuration = {},
+  method,
+  path,
+  data,
+  headers = {}
+) => {
+  const { baseUrl, access_token } = configuration;
+
+  const errors = {
```
josephjclark, line 74: I know it's in the template but I would actually remove the errors object entirely. Maybe we should do the same in the template too, I'm not sure at this state :thinking: 

Alternatively, I don't know anything about ODK - what do its errors look like, and should we add more mappings?

josephjclark (changes requested):
This is great @hunterachieng. I've just got a couple of small comments but not big deal!

packages/odk/configuration-schema.json
```diff
@@ -0,0 +1,49 @@
...
+      "minLength": 1,
+      "examples": [
+        "the-long-access-token-from-your-auth"
+      ]
+    }
+  },
+  "type": "object",
+  "additionalProperties": true,
+  "required": [
+    "password",
```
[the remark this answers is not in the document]
hunterachieng: Should I remove the items within? I think we need them for the auth

packages/odk/src/Adaptor.js
hunterachieng: Should I also add `@public` to `get` and `post`?

packages/odk/configuration-schema.json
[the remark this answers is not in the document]
josephjclark: Let's leave them required for now. I've raised an issue to try and work this out

packages/odk/src/Adaptor.js
josephjclark: Ooh yes please, on all public operations! Both here and in the template

packages/odk/src/Utils.js
josephjclark: If we get a clean error in the CLI then we don't need to add this error mapping object here.

josephjclark (approved):

PR #610, opened 2024-06-06
Refactor docs build into two scripts

josephjclark:
To make local docs development easier, this PR splits the docs.json build into two steps: one to build the docs, and one to commit and publish.

Now you can run `pnpm docs:build` to generate your docs.json file, and easily import that into your local docs repo (see https://github.com/OpenFn/docs/pull/498)

I've also moved the scripts into `./scripts`, where they are more discoverable, and hooked them into package.json

This should make absolutely no change to the actual build.

### DIFF
--- a/scripts/build-docs.sh
+++ b/scripts/build-docs.sh
@@ -1,5 +1,7 @@
 #!/bin/sh -l
 
+echo "Building docs.json"
+
 # clean tmp directory
 mkdir -p tmp
 rm -rf tmp/*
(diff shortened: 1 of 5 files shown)


scripts/build-docs.sh
```diff
@@ -15,8 +17,9 @@ find tmp/*/* -type d -empty -delete
 
 # flatten json files
 # jq -s 'flatten' packages/*/docs/*.json > docs/docs.json
-rm tmp/*.json
-# add first opening bracked
+# rm tmp/*.json
```
josephjclark, line 20: This line is generating an error for me locally. Can we remove it? I feel like we shouldn't need it...

CC @mtuchi

PR #612, opened 2024-06-07
Show state in docs

### LINKED ISSUE #602: New pattern for documenting state mutation clearly
### Description

josephjclark:
This PR enables a table of state mutations to be shown in the JSDoc for an adaptor.

This makes it easier to understand how an operation manipulates state.

image

## Issues

Closes #602

## How it works

In an adaptor's JSDoc, just add `@state` annotations to describe every property of state that is modified by that operation:

```
/**
 * Make a GET request
 * @param {string} path - Path to resource
 * @param {RequestOptions} params - Query, Headers and Authentication parameters
 * @param {function} callback - (Optional) Callback function
 * @state data - the response body will be written to state.data
 * @state response - the response from the HTTP server, including headers, statusCode, body, etc
 * @state references - the last response, without the body, will be pushed to references
 * @returns {Operation}
 */
  return sendRequest('GET', path, params, callback);
}
```

You can also reference an object as a TypeDef, enabling re-usable definitions:
```
/**
 * Make a GET request
 * @param {string} path - Path to resource
 * @param {RequestOptions} params - Query, Headers and Authentication parameters
 * @param {function} callback - (Optional) Callback function
 * @state HttpState
 * @returns {Operation}
 */
  return sendRequest('GET', path, params, callback);
}
```

Each `@state` key acts a bit like a mixin. Properties will override each other and the last to be applied will win. So you could go crazy and do something like this:

```
/**
 * Make a GET request
 * @param {string} path - Path to resource
 * @param {RequestOptions} params - Query, Headers and Authentication parameters
 * @param {function} callback - (Optional) Callback function
 * @state BaseState
 * @state HttpState
 * @state data - the GET response body will be written to state.data
 * @returns {Operation}
 */
  return sendRequest('GET', path, params, callback);
}
```

Neat.

## Defaults

As a final note, you can also do `@state data` or `@state refereces` and just set the name - in which case a default string will be loaded from `tools/build/jsdoc/state-defaults.cjs`.

This is a convenience which is probably made redundant by the typedefs thing. But I'll leave it in.

## QA

I have updated the HTTP adaptor with full state docs as part of this PR.

I have tested state overrides but didn't leave any in because the docs don't actually need them in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "setup": "pnpm --filter \"./tools/**\" install",
     "docs:publish": "pnpm exec scripts/publish-docs.sh",
     "docs:build": "pnpm exec scripts/build-docs.sh",
-    "docs:watch": "chokidar packages/*/docs/*.md -c 'pnpm docs:build' --throttle 2000",
+    "docs:watch": "chokidar packages/*/docs/*.md -c 'pnpm docs:build'",
     "build:tools": "pnpm --filter \"./tools/**\" build",
     "build:adaptors": "pnpm --filter \"./packages/**\" build",
     "build": "pnpm build:tools && pnpm build:adaptors",
--- a/packages/http/src/Adaptor.js
+++ b/packages/http/src/Adaptor.js
@@ -54,6 +54,7 @@ export function execute(...operations) {
  * @param {string} path - Path to resource
  * @param {RequestOptions} params - Query, Headers and Authentication parameters
  * @param {function} callback - (Optional) Callback function
+
  * @returns {Operation}
  */
 export function request(method, path, params, callback) {
@@ -72,6 +73,10 @@ export function request(method, path, params, callback) {
  * @param {string} path - Path to resource
  * @param {RequestOptions} params - Query, Headers and Authentication parameters
  * @param {function} callback - (Optional) Callback function
+ * @magic path - it magic you know
+ * @state data - the response body will be written to state.data
+ * @state response - the resonse from the HTTP server, including headers, statusCode, body, etc
+ * @state references - the last response, without the body, will be pushed to references
  * @returns {Operation}
  */
 export function get(path, params, callback) {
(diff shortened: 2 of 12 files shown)


packages/http/src/Adaptor.js
```diff
@@ -15,7 +15,16 @@ import { request as sendRequest, xmlParser } from './Utils';
  */
 
 /**
- * Execute a sequence of operations.
+ * State object
+ * @typedef {Object} HttpState
+ * @private
```
josephjclark, line 20: Marking it private means that this typedef won't be rendered in the docs as a standalone thing

packages/http/src/Adaptor.js
```diff
@@ -172,6 +187,8 @@ export function del(path, params, callback) {
  * @param {String} body - data string to be parsed
  * @param {function} script - script for extracting data
  * @param {function} callback - (Optional) Callback function
+ * @state data - the parsed XML as a JSON object
+ * @state references - an array of all previous data objects used in the Job
```
josephjclark, line 191: I've had to copy and paste references here. One thing I can't do is reference a type and then REMOVE a key.

But it's fine like this.

Another way to approach the problem would be to have a `BaseState` of `data`, `references` and `HTTPState` which adds `response`, and use BOTH as mixins. But it's not worth the overhead.

package.json
```diff
@@ -10,7 +10,7 @@
     "setup": "pnpm --filter \"./tools/**\" install",
     "docs:publish": "pnpm exec scripts/publish-docs.sh",
     "docs:build": "pnpm exec scripts/build-docs.sh",
-    "docs:watch": "chokidar packages/*/docs/*.md -c 'pnpm docs:build' --throttle 2000",
```
josephjclark, line 13: I took this off because with throttle and debounce it seemed to rebuild like twice and then stop. Anyway it seems to be fine

Issue #522, opened 2024-04-23
Implement new util pattern


josephjclark:
I'm keen to start implementing this pattern:

* All exported functions from an adaptor are OPERATIONS (take state, return state)
* Any non-operation functions are exported onto the `util` (or `utils`? I think it should be singular) object.

Typically util functions will be in a `util.js` file but this is less important. What's important is that the main adaptor object includes a utils object.

When using non-operation functions in job code, you always do `util.whatever`, like `util.fetch` or `util.jsonPath(state, path)`.

Doing this helps make it clear to job writers what is an operation (and needs state passing) and what is just a function. Because it's pretty confusing right now!

## [SOLVED] Difficulty #1

The hardest thing about this pattern is the docs!

How do document an exported namespace? 

We don't handle this well at the moment. There are no docs on the common/http helpers.

There is a commented out bit in the docs template for interfaces, I think. Maybe we just restore this, but it may not be very readable.

Basically for each adaptor we need to list Functions (actually, Operations); and Utils, with the note that utils are available on the utils object.

## Difficulty #2

What do we do about common, which right now has `http` and `util` namespaces?

Also common.http is actually old stuff.

I think the answer, as in #507 , is just to add the http helpers to the util object (even though they're in the http namespace). It's fine :)

## Difficulty #3 

When every adaptor adopts this pattern we are living in utopia, nirvana, shangra-la.

But while migrating, we're basically just introducing even more inconsistency to the design :( It reduces predictability.

This just means we should migrate the top 10 adaptors quickly, but maybe even do a push on the next 20.

josephjclark:
I will spike something here

josephjclark:
As of right now, the doc site supports namespacing. So if we were to add a `utils.` namespace, the docsite can easily handle it. Hurrah!

PR #613, opened 2024-06-07
Docs: split API into Operations and Functions list

### LINKED ISSUE #439: Make it clear whether a function is an Operation or just a Function
Adaptors export a bunch of functions to be used in job code.

Most of these functions are Opertions - outside of `common`, probably 99% of them are operations.

But some are just functions,

Operations and functions should be treated slightly differently by end users (see #428), so some kind of `operation` label/pill on the adaptor docs, or maybe a `show operations only` checkbox on the docs page, would be really helpful.

It would also be nice for users to see the docs and say "hey, what's this operation stuff all about?" and see a link to a really clear explanation.

josephjclark:
## Summary

This PR changes the  docs template so that Operations and Functions feature in two different lists.

This helps make it clear what is an Operation and what is a Plain Old Function

## Details

As a bonus fix, indenting on the right hand contents menu has been fixed.

These changes do not affect the adaptor JSdoc at all. It just works out of the box.

Screenshots from a first pass:

image
image
image

## Issues

Closes https://github.com/OpenFn/docs/issues/439

Related to #522

## Further work

* [ ] Add an `@operation` jsdoc tag, which marks the doclet public and is used to recognise operations
* [ ] The Operations list doesn't show signatures
* [ ] Put the operation/function summary contents into a table with two columns, so you can see both. Maybe just on desktop windows?
* [ ] Don't show the return type in an Operation signature
* [ ] I think the Operation keyword should feature somewhere in each Operation, like a badge. Clicking it would take you to our best explanation of what an Operation is
* [ ] Improve the summary docs at the top of the Operations and Functions sections (include links)

### DIFF
--- a/tools/build/src/commands/docs.ts
+++ b/tools/build/src/commands/docs.ts
@@ -48,6 +48,19 @@ const build = async (lang: string) => {
     'no-cache': true,
   });
 
+  // Identify Operations
+  // TODO we can later add a supporting @operation tag, but this heuristic will go a long way
+  templateData.forEach(doclet => {
+    if (
+      doclet.returns
+        ?.at(0)
+        ?.type?.names.at(0)
+        .match(/operation/i)
+    ) {
+      doclet.kind = 'operation';
+    }
+  });
+
   // sort template data
   // see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort
   templateData.sort(function (a, b) {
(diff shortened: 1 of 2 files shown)


josephjclark:
I wonder how the common API fits into this.

I don't want 4 lists - adaptor operations, adaptor functions, common functions, common operations. That's chaos.

We can't "merge" operations and functions from common, because it's confusing.

So maybe we list everything in one list, and perhaps add a* if it comes from common?

josephjclark:
One of the things with the common API is that, because it's the same for all adaptors, once you know the main common functions, they just sort of get in the way.

So yes, you want to see that the common functions are there. And its critical for new users. But I do think they're second class citizens, or there should be the option to hide them, or something.

josephjclark:
I wonder if just adding a tag for Operations is a better way to go here. That still leaves namespaces to work out though...

PR #614, opened 2024-06-11
common: Fix request params

josephjclark:
In common http, request parameters in the URL were not getting sent on to the server.

I've tested against the `http` adaptor with this code:
```
get('/tracker/trackedEntities/qv0j4JBXQX0.json?fields=*', {
  // query: { fields: '*' },
});

fn(state => {
  console.log(state.data);
});
```
And this credential (public sandbox)
```
{
  "configuration": {
    "username": "admin",
    "password": "district",
    "baseUrl": "https://play.im.dhis2.org/stable-2-40-3-2/api/40/"
  }
}
```

This code calls out to dhis2 for a tracked entity and asks for all fields to be returned (via query parameter). When the query does not get sent, dhis only returns part of the entity, not the full thing.

In production, the logged response does not include `enrollments` or `programOwners`. Against the built monorepo, it does.

Versions are bumped, tests are added, this is a critical issue ready for release

### DIFF
--- a/.changeset/tame-parrots-complain.md
+++ b/.changeset/tame-parrots-complain.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-commcare': minor
----
-
-Implement better error handling and make post a public function
--- a/packages/bigquery/CHANGELOG.md
+++ b/packages/bigquery/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-bigquery
 
+## 2.0.10
+
+### Patch Changes
+
+- Updated dependencies
+  - @openfn/language-common@1.13.5
+
 ## 2.0.9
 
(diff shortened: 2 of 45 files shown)


mtuchi (approved):

mtuchi:
I have tested this code with http adaptor
```
get('/tracker/trackedEntities/qv0j4JBXQX0.json?fields=*', {
  // query: { fields: '*' },
});

fn(state => {
  console.log(state.data);
  return state;
});
```
Also tested this 👇🏽 
```
get('/tracker/trackedEntities/qv0j4JBXQX0.json', {
  query: { fields: '*' },
});

fn(state => {
  console.log(state.data);
  return state;
});
```

Both works fine

PR #618, opened 2024-06-12
Refactor docs build into two scripts

josephjclark:
To make local docs development easier, this PR splits the docs.json build into two steps: one to build the docs, and one to commit and publish.

Now you can run `pnpm docs:build` to generate your docs.json file, and easily import that into your local docs repo (see https://github.com/OpenFn/docs/pull/498)

I've also moved the scripts into `./scripts`, where they are more discoverable, and hooked them into package.json

This should make absolutely no change to the actual build.


josephjclark:
This worked great!

Cartodb updated with the change

image

Issue #604, opened 2024-06-05
`postgresql` expandRefs in queryHandler


mtuchi:
### Description 

Expand reference in queryHandler function

PR #619, opened 2024-06-12
Update `postgresql` functions to use util `expandReference`

mtuchi:
## Summary

Update all functions to use util `exapandReferences` and add `composeNextState` in `findValue` function

- Ref #604 

## Review Checklist

### DIFF
--- a/.changeset/green-monkeys-compete.md
+++ b/.changeset/green-monkeys-compete.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-postgresql': minor
+---
+
+- Update all operations to use util `expandReferences`
+- Add `findValue` result to state
--- a/.changeset/green-monkeys-compete.md
+++ b/.changeset/green-monkeys-compete.md
@@ -1,5 +1,5 @@
 ---
-'@openfn/language-postgresql': minor
+'@openfn/language-postgresql': major
 ---
 
 - Update all operations to use util `expandReferences`
--- a/packages/postgresql/src/Adaptor.js
+++ b/packages/postgresql/src/Adaptor.js
@@ -157,25 +157,25 @@ function queryHandler(state, query, options, callback) {
  * @example
  * sql(state => `select(*) from ${state.data.tableName};`, { writeSql: true })
  * @function
- * @param {string} sqlString - A sequence of characters that forms a valid SQL command.
+ * @param {string} sqlQuery - The SQL query as a string.
  * @param {object} [options] - Optional options argument
  * @param {boolean} [options.writeSql] - A boolean value that specifies whether to log the generated SQL statement. Defaults to false.
  * @param {boolean} [options.execute] - A boolean value that specifies whether to execute the generated SQL statement. Defaults to false.
  * @param {function} callback - (Optional) callback function
  * @returns {Operation}
  */
-export function sql(sqlString, options, callback) {
+export function sql(sqlQuery, options, callback) {
   return state => {
     const { client } = state;
 
-    const [resolvedSqlString, resolvedOptions] = expandReferences(
+    const [resolvedSqlQuery, resolvedOptions] = expandReferences(
       state,
-      sqlString,
+      sqlQuery,
       options
     );
     try {
       console.log('Preparing to execute sql statement');
-      return queryHandler(state, resolvedSqlString, resolvedOptions, callback);
+      return queryHandler(state, resolvedSqlQuery, resolvedOptions, callback);
     } catch (e) {
       client.end();
       throw e;
--- a/packages/postgresql/src/Adaptor.js
+++ b/packages/postgresql/src/Adaptor.js
@@ -1,8 +1,8 @@
 import {
   execute as commonExecute,
   composeNextState,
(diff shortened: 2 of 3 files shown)


packages/postgresql/src/Adaptor.js
```diff
@@ -157,23 +157,25 @@ function queryHandler(state, query, options, callback) {
...
- * string of SQL.
+ * @param {string} sqlString - A sequence of characters that forms a valid SQL command.
  * @param {object} [options] - Optional options argument
  * @param {boolean} [options.writeSql] - A boolean value that specifies whether to log the generated SQL statement. Defaults to false.
  * @param {boolean} [options.execute] - A boolean value that specifies whether to execute the generated SQL statement. Defaults to false.
  * @param {function} callback - (Optional) callback function
  * @returns {Operation}
  */
-export function sql(sqlQuery, options, callback) {
+export function sql(sqlString, options, callback) {
```
josephjclark, line 167: I think `sqlQuery` is still appropriate? It's a string or a function that returns a string.

I agree that the old way was too dogmatic and that making it an expandable string is correct. I just don't think we should rename it.

Also `sequence of characters that forms a valid SQL command` is strange phrasing. I'd just say "the SQL query as a string"

packages/postgresql/src/Adaptor.js
```diff
@@ -234,12 +240,16 @@ export function findValue(filter) {
...
+              ...composeNextState(state, returnValue),
+              result: returnValue,
+            };
+            resolve(nextState);
           }
         });
       });
     } catch (e) {
-      client.end();
+      console.log(e, 'asd');
```
josephjclark, line 252: This logging needs removing. And are you sure you don't want to end the client here?

josephjclark: Wait does this even fire?

Ach, I don't want to touch code I don't understand if I don't need to..

.changeset/green-monkeys-compete.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-postgresql': minor
+---
+
+- Update all operations to use util `expandReferences`
+- Add `findValue` result to state
```
josephjclark, line 6: This is a major change

packages/postgresql/src/Adaptor.js
```diff
@@ -264,23 +274,31 @@ export function findValue(filter) {
...
+        ),
+        handleOptions(resolvedOptions)
       );
 
-      const safeQuery = `INSERT INTO ${table} (${columnsList}) VALUES [--REDACTED--]];`;
+      const safeQuery = `INSERT INTO ${resolvedTable} (${columnsList}) VALUES [--REDACTED--]];`;
 
-      const queryToLog = options && options.logValues ? query : safeQuery;
+      const queryToLog =
+        resolvedOptions && resolvedOptions.logValues ? query : safeQuery;
```
josephjclark, line 299: Try `resolvedOptions?.logValues ? query : safeQuery;`

packages/postgresql/src/Adaptor.js
```diff
@@ -307,30 +325,37 @@ export function insert(table, record, options, callback) {
...
+          ),
+          handleOptions(resolvedOptions)
         );
 
-        const safeQuery = `INSERT INTO ${table} (${columnsList}) VALUES [--REDACTED--]];`;
+        const safeQuery = `INSERT INTO ${resolvedTable} (${columnsList}) VALUES [--REDACTED--]];`;
 
-        const queryToLog = options && options.logValues ? query : safeQuery;
+        const queryToLog =
+          resolvedOptions && resolvedOptions.logValues ? query : safeQuery;
```
josephjclark, line 356: Again, as we're touching it anyway, `resolvedOptions.logValues ? query : safeQuery;`

packages/postgresql/src/Adaptor.js
```diff
@@ -363,14 +388,16 @@ export function insertMany(table, records, options, callback) {
  */
 export function upsert(table, uuid, record, options, callback) {
   return state => {
-    let { client } = state;
+    const { client } = state;
 
+    const [resovledTable, resovledUuid, resolvedRecord, resolvedOptions] =
```
josephjclark, line 393: Typos aplenty here, eg `resovledUuid`

josephjclark (changes requested):
Thanks @mtuchi , just a couple of things

packages/postgresql/src/Adaptor.js
mtuchi: Well you can pretty much write an sql command Eg: `create table tableName (column1 dataType)`,  To me `sqlQuery` means only write sql queries. Eg: `select * from patients`

packages/postgresql/src/Adaptor.js
mtuchi: Removed the `console.log()`

josephjclark: No worries, I do that sort of thing all the time :) This is why we self-review

josephjclark (approved):

Issue #375, opened 2023-08-31
common: add something like `fnIf` or `doIf`


josephjclark:
This would be like salesforce's `upsetIf`

It takes a condition as the first argument, and an operation as the second. The operation will only execute if the condition returns a truthy value.

```
fnIf((state) =? state.data?.lengtth, post(www, () => state.data)
```

This would help us remove some complicated nesting in some job code

josephjclark:
@mtuchi :+1: from #510

josephjclark:
More evidence in https://community.openfn.org/t/upsertif-function-for-mysql/444

Waiting on the response but if they only want a `upsert if data exists` pattern then we can probably build better logic into the adaptor

PR #620, opened 2024-06-12
Add `fnIf` in `common`

mtuchi:
## Summary

Add `fnIf` operation in common, And prettier formatting made tiny change to `cursor` function

Ref #375 

## Review Checklist


mtuchi:
Sample test code for the `fnIf` function 

```
fn(state => {
  state.name = 'ama';
  state.users = [
    {
      name: 'mtuchi',
      age: 'NaN',
    },
    {
      name: 'Africa',
      age: 'NaN',
    },
  ];
  return state;
});

fnIf(
  state => state?.name,
  state => {
    console.log(state.name);
    return state;
  }
);

fnIf(
  state => state.users.length > 1,
  each($.users, state => {
    console.log(state.data);
    return state;
  })
);

// Only if you export common/util 👇🏽 
fnIf($?.name, async state => {
  const { body } = await util.get('https://jsonplaceholder.typicode.com/users');
  state.results = body;

  return state;
});

```

josephjclark:
Quick note: each adaptor needs to explicitly export this from common, and also needs a minor bump

packages/common/src/Adaptor.js
```diff
@@ -67,6 +67,27 @@ export function fn(func) {
   };
 }
 
+/**
+ * A custom step (or operation) that will only execute if the condition returns a truthy value
```
josephjclark, line 71: step is confusing because it means something else in lightning.

I've just seen you've taken this from `fn`, so lets tweak `fn`'s docs too!

Let's say:
```
Execute a block of arbitrary JavaScript code. Take state as an argument and MUST return state at the end
```

packages/common/src/Adaptor.js
```diff
@@ -67,6 +67,27 @@ export function fn(func) {
...
+ * @public
+ * @example
+ * fnIf((state) => state?.data?.name, get("https://example.com"));
+ * @function
+ * @param {Boolean} truthyCondition is the condition that returns true
```
josephjclark, line 76: I would just call it `condition`, and the docs  are something like `execute the function if this condition returns true`

packages/common/src/Adaptor.js
```diff
@@ -67,6 +67,27 @@ export function fn(func) {
...
+ * @returns {Operation}
+ */
+export function fnIf(truthyCondition, operation) {
+  return state => {
+    const [resolvedTrutyCondition] = newExpandReferences(
+      state,
+      truthyCondition
+    );
+
+    return resolvedTrutyCondition ? operation(state) : state;
```
josephjclark, line 87: Good - nice and clean and minimal

josephjclark (approved):
I've left some comments about the docs but actually you should ignore those - I'll do a docs pass on this myself

mtuchi:
I have updated each adaptor export `fnIf` from common and bump their version to a minor

josephjclark:
Oh man @mtuchi I was not ready for this to be released!

I guess I shouldn't have approved it. That was a "thanks, I'll take it from here" sort of approval. I did say I was going to look at the docs.

Because we've got a lot of common changes coming together, I wanted to do one big release (maybe even a major) with them all at once

mtuchi:
My bad @josephjclark 🙌🏽  , I thought this was ready for release. I will make sure to ask before merging next time

Issue #605, opened 2024-06-05
Common: add `group` operation


josephjclark:
EDIT: forget `bin`, this is `group`

The `bin` operation takes an array of data and sorts it into "bins".

```
bin(pathToArray, (state) => binNameString)
```

It works like `each`, taking an array or jsonpath as an argument, and iterating over each item with scoped data.

The return value of the function is the name of the bin to add the item to.

Bins are added as keys to the state object

For example, for this state
```
{
  data: [
    { name: 'anna', role: 'patient' },
    { name: 'betty',  role: 'doctor' },
    { name: 'charlie',  role: 'doctor' },
  ]
}
```
Calling this:
```
bin('$.data', (state) => state.data.role)
```

Results in this state:
```
{
  data: [ /* unchanged */ ]
  patient: [
    { name: 'anna', role: 'patient' },
  ],
  doctor: [
    { name: 'betty',  role: 'doctor' },
    { name: 'charlie',  role: 'doctor' },
  ]
}
```

mtuchi:
I am not familiar with a key word `bin`. But from what i have understood so far, We need to a util helper that will sort an array of objects by a key path. The suggested signature `bin(pathToArray, keyPath)`. 

If this is the correct assessment then my suggestion is we don't use the word bin, We should use something very specific to what this function is doing. Eg sortByKey or sortBy. I personally like the word `groupBy`

I will sync with @josephjclark to get his opinion as well

josephjclark:
Oh man groupBy is a way better name 🤦🤦🤦

Or maybe just `group`?

For me `sort` just means "change the order of an array", whereas this function is "organise this data into groups"

mtuchi:
@josephjclark should this be an operation or a util helper ?

aleksa-krolls:
hey @mtuchi @josephjclark can we close this issue out? I see the PR has been merged

PR #621, opened 2024-06-12
Add  `group()` operation in common

mtuchi:
## Summary

Add a `group()` operation in language-common with unit test

Ref #605

## Review Checklist

### DIFF
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -409,11 +409,7 @@ export function merge(dataSource, fields) {
 
 /**
  * Groups an array of objects by a specified key path.
- *
- * @param {Object[]} arrayOfObjects - The array of objects to be grouped.
- * @param {string} keyPath - The key path to group by.
- * @param {function} callback - (Optional) Callback function
- * @returns {Operation}
+ * @public
  * @example
  * const users = [
  *   { name: 'Alice', age: 25, city: 'New York' },
@@ -423,6 +419,11 @@ export function merge(dataSource, fields) {
  * ];
  * group(users, 'city');
  * // state is { data: { 'New York': [/Alice, Charlie/], 'San Francisco': [ /Bob, David / ] }
+ * @function
+ * @param {Object[]} arrayOfObjects - The array of objects to be grouped.
+ * @param {string} keyPath - The key path to group by.
+ * @param {function} callback - (Optional) Callback function
+ * @returns {Operation}
  */
 export function group(arrayOfObjects, keyPath, callback = s => s) {
   return state => {
@@ -432,7 +433,7 @@ export function group(arrayOfObjects, keyPath, callback = s => s) {
       keyPath
     );
     const results = groupBy(resolvedArray, item => get(item, resolvedKeyPath));
-    return callback(composeNextState(state, results));
+    return callback({ ...state, data: results, results });
   };
 }
 
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -422,6 +422,7 @@ export function merge(dataSource, fields) {
  *   { name: 'David', age: 30, city: 'San Francisco' }
  * ];
  * group(users, 'city');
+ * // state is { data: { 'New York': [/Alice, Charlie/], 'San Francisco': [ /Bob, David / ] }
  */
 export function group(arrayOfObjects, keyPath, callback = s => s) {
   return state => {
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -1,6 +1,6 @@
 import curry from 'lodash/fp/curry.js';
 import fromPairs from 'lodash/fp/fromPairs.js';
-
+import { groupBy, get } from 'lodash';
 import { JSONPath } from 'jsonpath-plus';
 import { parse } from 'csv-parse';
 import { Readable } from 'node:stream';
@@ -406,6 +406,34 @@ export function merge(dataSource, fields) {
 }
 
 /**
+ * Groups an array of objects by a specified key path.
+ *
+ * @param {Object[]} arrayOfObjects - The array of objects to be grouped.
+ * @param {string} keyPath - The key path to group by.
+ * @param {function} callback - (Optional) Callback function
+ * @returns {Operation}
+ * @example
+ * const users = [
+ *   { name: 'Alice', age: 25, city: 'New York' },
+ *   { name: 'Bob', age: 30, city: 'San Francisco' },
+ *   { name: 'Charlie', age: 25, city: 'New York' },
+ *   { name: 'David', age: 30, city: 'San Francisco' }
+ * ];
+ * group(users, 'city');
+ */
+export function group(arrayOfObjects, keyPath, callback = s => s) {
+  return state => {
+    const [resolvedArray, resolvedKeyPath] = newExpandReferences(
+      state,
+      arrayOfObjects,
+      keyPath
+    );
+    const results = groupBy(resolvedArray, item => get(item, resolvedKeyPath));
+    return callback(composeNextState(state, results));
+  };
+}
+
+/**
  * Returns the index of the current array being iterated.
  * To be used with `each` as a data source.
  * @public
@@ -463,6 +491,9 @@ export function toArray(arg) {
  * @returns {State}
  */
 export function composeNextState(state, response) {
+  if (!state.references) {
+    state.references = [];
+  }
   return {
     ...state,
     data: response,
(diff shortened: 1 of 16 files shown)


packages/common/src/Adaptor.js
```diff
@@ -463,6 +493,9 @@ export function toArray(arg) {
  * @returns {State}
  */
 export function composeNextState(state, response) {
+  if (!state.references) {
```
josephjclark, line 496: This needs to be mentioned in the changeset

packages/common/src/Adaptor.js
```diff
@@ -405,6 +407,34 @@ export function merge(dataSource, fields) {
   };
 }
 
+/**
+ * Groups an array of objects by a specified key path.
+ *
+ * @param {Object[]} arrayOfObjects - The array of objects to be grouped.
+ * @param {string} keyPath - The key path to group by.
```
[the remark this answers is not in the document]
josephjclark: Oh, but you're using it with get. So I can pass a path like `address.home.city`.

Ok, that's cool. Let me think about the wording of the docs though

packages/common/test/index.test.js
```diff
@@ -866,95 +867,107 @@ describe('validate', () => {
...
-    expect(result.cursor).to.eql({ page: 3 })
-  })
+    expect(result.cursor).to.eql({ page: 3 });
+  });
+});
 
-});
\ No newline at end of file
+describe('group', () => {
+  it('should group an array of objects by a specified key path. ', async function () {
```
josephjclark, line 963: As this is a common core function I'd like a few more tests please

- use a path rather than a key
- what happens if the key isn't present on any items (should happily return an empty object)
- what happens if a key value is undefined? eg, `{ city: undefined }`
- check that `arrayOfObjects` is expanded (pass a function for the first argument)
- check that `path` is expanded (pass a function for the second argument)

Also, no full stops at the end of test names please!

packages/common/test/index.test.js
```diff
@@ -866,95 +867,107 @@ describe('validate', () => {
...
+    const state = {};
+    let resultingState = await group(testData.store.book, 'category')(state);
+
+    // Assert that state.data has certain keys
```
josephjclark, line 967: assert.deepStrictEqual is kind weird. `expect().eql()` is how we normally do this.

Anyway this isn't a terribly good test. The order of keys doesn't matter and we're not checking the values of the objects either.

I would declare a really simple array at the top:
```
const data = [{ x: 'a' }, { x: 'b'}]
```
And do a full test on the result:
```
expect(result.data).eql({
   a: [{ x:'a' }],
   b: [{ x:'b' }],
}
```

that's so much more declarative, and you can read over the test without any extra context.

packages/common/test/index.test.js
```diff
@@ -866,95 +867,107 @@ describe('validate', () => {
...
+    expect(result.cursor).to.eql({ page: 3 });
+  });
+});
 
-});
\ No newline at end of file
+describe('group', () => {
+  it('should group an array of objects by a specified key path. ', async function () {
+    const state = {};
+    let resultingState = await group(testData.store.book, 'category')(state);
```
josephjclark, line 965: There's no need to `await` this - it's a bit misleading. This is a synchronous function.

And why `let resultingState`?

josephjclark (changes requested):
Looks good - comments mostly on docs and tests.

I will take a look at the docs before merging

josephjclark:
@mtuchi when you've finished the changes, please export this from the most common adaptors (we don't have to do all, just stuff that's used regularly. Unless all is easier for you!) and add a minor changeset

packages/common/test/index.test.js
mtuchi: Let me fix that, it should be a const

mtuchi: Fixed here [f92e3c9](https://github.com/OpenFn/adaptors/pull/621/commits/f92e3c98396d673f5036311894c12665db415461)

packages/common/src/Adaptor.js
```diff
@@ -405,6 +407,34 @@ export function merge(dataSource, fields) {
...
+ * @param {function} callback - (Optional) Callback function
+ * @returns {Operation}
+ * @example
+ * const users = [
+ *   { name: 'Alice', age: 25, city: 'New York' },
+ *   { name: 'Bob', age: 30, city: 'San Francisco' },
+ *   { name: 'Charlie', age: 25, city: 'New York' },
+ *   { name: 'David', age: 30, city: 'San Francisco' }
+ * ];
+ * group(users, 'city');
```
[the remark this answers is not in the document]
mtuchi: Fixed [f92e3c9](https://github.com/OpenFn/adaptors/pull/621/commits/f92e3c98396d673f5036311894c12665db415461)

packages/common/src/Adaptor.js
```diff
@@ -405,6 +407,34 @@ export function merge(dataSource, fields) {
...
+ */
+export function group(arrayOfObjects, keyPath, callback = s => s) {
+  return state => {
+    const [resolvedArray, resolvedKeyPath] = newExpandReferences(
+      state,
+      arrayOfObjects,
+      keyPath
+    );
+    const results = groupBy(resolvedArray, item => get(item, resolvedKeyPath));
+    return callback(composeNextState(state, results));
```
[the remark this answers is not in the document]
josephjclark: Yes, but we don't need to write results twice
```
callback({ ...state, data: results })
```

packages/common/test/index.test.js
josephjclark: I can but I'm not sure I understand the question! The path in your example looks fine?

mtuchi: Ooh okay, i think we are align, i wanted to write a test for `use a path rather than a key` and i was confused on what this mean.

mtuchi: > If the key isn't present on any items (should happily return an empty object)

This means were are removing any object that does not have that key right ?

josephjclark: Well we don't _remove_ anything - we just don't add it. Basically `group` should always write an object to state.data, and that object may or may not have any groupings on it

mtuchi: If we are not removing the stuff instead of empty object, shouldn't that be `undefined` key 
Eg:
```
const data = [
  { x: "a", y: { z: "a" } },
  { x: "b", y: { w: "a" } },
];

group(data, 'y.z');

//Should return 👇🏽 
{
  a: [ { x: 'a', y: { z 'a' }} ],
  undefined: [ { x: 'b', y: { w: 'a' } } ]
}
```

packages/common/test/index.test.js
mtuchi: Using that pattern for new tests

packages/common/test/index.test.js
mtuchi: I have update the tests, I am not sure if i am testing `expand reference` accurately would love your feedback

packages/common/test/index.test.js
```diff
@@ -866,95 +867,172 @@ describe('validate', () => {
...
-    expect(result.cursor).to.eql({ page: 3 })
-  })
+    expect(result.cursor).to.eql({ page: 3 });
+  });
+});
+
+describe('group', () => {
+  it('should group an array of objects by a specified key path', function () {
+    const state = {};
+    const data = [{ x: 'a' }, { x: 'b' }, { x: 'b' }];
```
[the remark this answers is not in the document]
mtuchi: I wanted to show that you will have two items grouped in `b` because otherwise it feels like repeating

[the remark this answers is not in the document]
josephjclark: Oh I see! My brain jumped straight to the pattern.

Ok, that's fine!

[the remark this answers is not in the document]
mtuchi: Got you 😄 , I will export this to our top 5 adaptors, Do you have more feedback ?

packages/common/test/index.test.js
```diff
@@ -866,95 +867,172 @@ describe('validate', () => {
...
+  it('should remove undefined keys', function () {
+    const data = [
+      { x: 'a', y: { z: 'a' } },
+      { x: 'b', y: { z: undefined } },
+    ];
+    const state = {};
+    const result = group(data, 'y.z')(state);
+    expect(result.data).eql({ a: [{ x: 'a', y: { z: 'a' } }] });
+  });
+  it('should expand arrayOfObjects', function () {
```
josephjclark, line 1005: Ooh, ok, this one IS weird 

I expect `result` to be the result of calling `group` - like it is is in every other test.

But here, `result` is used to read a value off input. It's not even a result, it's like you're testing the arguments to the lazy function :thinking: 

I think this test should be just like the others: do `group(s => data, 'x')`, and check that the result is what you expect.

packages/common/test/index.test.js
```diff
@@ -866,95 +867,172 @@ describe('validate', () => {
...
+    let result;
+    group(state => {
+      result = state.data;
+      return result;
+    }, 'y.z')(input);
+
+    expect(result).eql(input.data);
+  });
+
+  it('should expand key path', function () {
```
josephjclark, line 1022: Again, this one is also testing the arguments to the expanded function and that's not really the point. Let's just test the overall result with `group(state.data, () => 'x')`

packages/common/test/index.test.js
mtuchi: I have updated the test to use the same pattern you described above

josephjclark (approved):
Awesome, thanks!

PR #628, opened 2024-06-13
FHIR: fix `create` and `request`

### LINKED ISSUE #622: `fhir`: implement http for generic `get` and `post` methods
## Request
For the Jembi hackathon tomorrow, I really want to showcase a workflow that can send basic GET and POST requests to a demo HAPI FHIR API. (Ideally, this can be ready for testing later _today_!) 

**Please update the existing fhir adaptor to include HTTP for generic `get` and `post` methods.**

(FYI eventually we'll need `put` and `delete` as well, but pls just start with get/post.)

## To test
I recommend testing with the `patient` resource to test. See the docs linked below for sample endpoints and payloads. 
- API docs: https://hapi.fhir.org/baseR4/swagger-ui/?page=Patient#/ 
- API base url: `https://hapi.fhir.org/baseR4/` (no creds needed for this test API)

Screenshot 2024-06-12 at 10 15 10 PM

josephjclark:
See the changelog and issue for details about the changes.

This improves the behaviour of all HTTP methods until we migrate properly onto common.http.

I've spiked the basic changes but I expect unit tests to fail.

We should also add `post` as an alias for `create`

This introduces subtle breaking changes so it's tagged as a major release

Closes #622


josephjclark:
@hunterachieng please take this over in the morning and get it read for release!

@mtuchi Please give this a once over in case I've missed anything or committed some huge faux pax.

This is intended to be a rushed and temporary change to support AK's hackathon

packages/fhir/src/Adaptor.js
```diff
@@ -3,6 +3,15 @@ import { expandReferences } from '@openfn/language-common/util';
 
 import { handleResponse, request } from './Utils';
 
+/**
+ * Options provided to the Create request
```
josephjclark, line 7: All http methood (get create post) all support these options - so we can't say Create here

packages/fhir/src/Adaptor.js
```diff
@@ -33,21 +42,37 @@ export function execute(...operations) {
...
+ * post("Bundle",{
+ * body:{
+ * "resourceType": "Bundle"
+ * },
+ * parseAs:'json',
+ * headers:{content-type:'application/json'}
+ * })
+ * @function
+ * @param {string} path - Path to resource
+ * @param {RequestOptions} params - contains body object to create the new resource, headers, and throwOnError
```
mtuchi, line 54: Just nitty picking, not really important

Should we call it `RequestParams` instead of ~RequestOptions~ ?

josephjclark: There's something going on here with params and options that I'm not that happy about... Trying to work out a good short term solution 

I actually think that create and post have different signatures

packages/fhir/src/Adaptor.js
```diff
@@ -3,6 +3,15 @@ import { expandReferences } from '@openfn/language-common/util';
 
 import { handleResponse, request } from './Utils';
 
+/**
+ * Options provided to the Create request
+ * @typedef {Object} RequestOptions
+ * @property {object|string} body - body data to append to the request. JSON will be converted to a string .
+ * @property {boolean} throwOnError - Optional boolean value to throw if the error is status code 400. Default `true`
```
mtuchi, line 10: Again nitty picking, but will be nice if we can add the default values in jsdoc. 
`Eg: [throwOnError=true]`. 
This will create a `default` column in a `RequestOptions` table.

packages/fhir/ast.json
mtuchi, line 1: Make sure to build again, i am noticing untracked changes locally when i run `pnpm build --watch`

packages/fhir/src/Adaptor.js
mtuchi: I think they are the same well at-least the intention for `create()` / `post()`. I remember when i was working on this we choose to use the word `create` because it's more related to both `developer` and `non-developer`. 

It's the same naming we have in `DHIS2`, the `create()` operation is actually a customized `post()` function which makes it's easy to create different resources

josephjclark: @mtuchi What do you think is the difference between a Param and an Option?

josephjclark:
Alright folks, we've got this wrong :(

`create` and `post` are not the same, and we should not change `create` at all in this work.

`create` is a high level abstraction to create a resource. It should not expose any of HTTP concepts.

`post` is a low level HTTP helper function. It should not expose any FHIR semantics.

I'm just trying to work out what that means for this PR. I really wish we weren't in a rush with this because we're going to create a low-quality release here :(

josephjclark:
Here's the plan @hunterachieng 

1. Revert create to how it is on main
1. Adjust create to work with the new request signature
1. create's return should not call `handleResponse` - it should just return `callback({ ...state, data})` to write the response data to state, and ignore the actual response
1. `post` needs a signature of `post(path, data, params, callback)`, where the body is in the data argument
1. Leave the parseAs option as it is now, it's fine

Get is a real mess. No-one's asked us to change it though, so I don't want to change it's behaviour or API. Let's leave it alone (the changes made on this branch are fine)

packages/fhir/src/Adaptor.js
mtuchi: When i see `params` in my head this is something like `query, body, headers`, where i see options it translate to something like `throwOnError`, `pollTimeout` [Options usually have defaults values]. 

I think both can be used mean the same thing, But usually what goes to my head when i see `options` keyword, I think these are optional params

packages/fhir/src/Adaptor.js
josephjclark: Ok, this documentation is basically all wrong. It's half wrong now and it's half wrong when we make our changes.

We don't have time to fix it.

For this release, I want the documentation to be vague.

Please remove this type def. I will add light documentation to `post`, but we should basically describe the `params` type as an object

packages/fhir/src/Adaptor.js
josephjclark: Interesting

It's an open question. There is no formal official definition, and so far as I know there's no strict openfn usage for these terms.

We often use params for our http helpers, but actually params is not conceptually tied to HTTP at all.

I guess options only works if there are no required keys.

It's the kind of terminology I'd like us to be consistent with across adaptors. I don't know the answers yet - just wanted your thoughts.

josephjclark:
Ha, I can't official Approve this because it's my PR. But I approve!

:heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark:

mtuchi (approved):
Tested with the following snippet 
```
get('Patient');

// fn(state => {
//   console.log(state.data);
//   return state;
// });

post(
  'Bundle',
  $.data,
  {
    headers: { 'content-type': 'application/json' },
    errors: { 400: false },
    parseAs: 'json',
  },
  state => {
    state.data.response = state.response;
    return state;
  }
);
```

It works as expected 👍🏽

PR #645, opened 2024-06-24
mailchimp: handle errors a little better

josephjclark:
.


josephjclark:
Seeing some very funky stuff in the error tests. Digging in a bit.

taylordowns2000 (approved):

PR #646, opened 2024-06-25
Small fixes on ODK

taylordowns2000:
## Summary

- ODK uses `email`, not `username`. This changes the config schema so it matches what ODK wants and what their documentation provides. (https://docs.getodk.org/central-api-authentication/)
- By manually concatenating the `baseUrl` and the `path` in the adaptor, rather than relying on `commonRequest` to do so, we lost the ability to handle `/` addition/omission. We don't want users to have to figure out if they should or shouldn't put `/` in various places.
- This adds a log to explain that authentication failed.
- This removes the try/catch so that users can see the helpful error message returned by `commonRequest`

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/strange-trainers-promise.md
+++ b/.changeset/strange-trainers-promise.md
@@ -1,0 +1,16 @@
+---
+'@openfn/language-odk': patch
+---
+
+Small fixes includes
+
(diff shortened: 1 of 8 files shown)


josephjclark:
How efficient - this closes #684 #647 #649

PR #650, opened 2024-06-25
Enable automated pre-releases

josephjclark:
This PR just switches pre-releases back on

Do no merge until  Lightning is ready


josephjclark:
Oh hey, I need to check that the doc site ignores next tags!

Issue #655, opened 2024-07-01
Satusehat: Fix error handling


hunterachieng:
Satusehat has a different way of logging its errors and this affects how we display them on the logs.

A fix is suggested to ensure proper error logs for the `satusehat` adaptor.

Currently, this is how the logs look like from `satusehat`:
```{
    "resourceType": "OperationOutcome",
    "text": {
        "status": "generated"
    },
    "issue": [
        {
            "severity": "error",
            "code": "duplicate",
            "details": {
                "text": "Found duplicate resource: Encounter"
            }
        }
    ]
}```

hunterachieng:
@aleksa-krolls  I have created this issue for a quick fix on `satusehat`. Currently, we are unable to read the whole error message. Can we prioritise this since its creating a blocker in the HIH workflow?

aleksa-krolls:
@hunterachieng sure! Will add to the sprint now

PR #656, opened 2024-07-01
fix:add error logs

hunterachieng:
## Summary

Adding error logs to `satusehat`

## Details

`satusehat` errors are not correctly displayed therefore a temporary log has been added to display the errors before the CLI error log is fixed.

## Issues

#655 

## Review Checklist

      migration guide followed?

### DIFF
--- a/.changeset/funny-planes-switch.md
+++ b/.changeset/funny-planes-switch.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-satusehat': patch
----
-
-Added extra logging around errors
--- a/.changeset/funny-planes-switch.md
+++ b/.changeset/funny-planes-switch.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-satusehat': minor
+---
+
+Add temporary error logs
--- a/.changeset/funny-planes-switch.md
+++ b/.changeset/funny-planes-switch.md
@@ -1,5 +1,5 @@
 ---
-'@openfn/language-satusehat': minor
+'@openfn/language-satusehat': patch
 ---
 
-Add temporary error logs
+Added extra logging around errors
--- a/packages/satusehat/src/Adaptor.js
+++ b/packages/satusehat/src/Adaptor.js
@@ -57,7 +57,7 @@ export function get(path, params = {}, callback = s => s) {
 
       return prepareNextState(state, response, callback);
     } catch (e) {
-      console.error(e);
+     console.error(e.body)
       throw e;
     }
   };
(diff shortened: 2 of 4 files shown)


.changeset/funny-planes-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-satusehat': minor
+---
+
+Add temporary error logs
```
josephjclark, line 5: I don't really like call these temporary here. I hope they are but still.

Can we just say "Added extra logging around errors" ?

.changeset/funny-planes-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-satusehat': minor
```
josephjclark, line 2: This should be a patch change because it doesn't add anything to the adaptor API. You can just rename `minor` to `patch` in the changeset here

packages/satusehat/src/Adaptor.js
```diff
@@ -57,6 +57,7 @@ export function get(path, params = {}, callback = s => s) {
 
       return prepareNextState(state, response, callback);
     } catch (e) {
+      console.log({ error: e.body });
```
josephjclark, line 60: I think `console.error(e)` would work better here (and in other cases) and be more robust

packages/satusehat/src/Adaptor.js
```diff
@@ -95,7 +96,8 @@ export function post(path, data, params = {}, callback = s => s) {
 
       return prepareNextState(state, response, callback);
     } catch (e) {
-      throw e.body.error ?? e;
+      console.log({ error: e.body });
+      throw e.body.issue ?? e;
```
josephjclark, line 100: The error here is inconsistent with the other errors

josephjclark (commented):
Correct me if I'm wrong - I can't easily test this stuff - but I think we can make this even simpler?

```
try {
 // do the thing
} catch(e) {
  console.error(e)
  throw e;
}
```

The error object `e` returned by the common HTTP helper will have useful stuff like the HTTP error code and the URL that was called.

Inside `e.body` will still be the richer Satusehat error object.

The `console.error` will log the full error object safely, while the thrown error will get written to state (or will be when I fix a bug there anyway)

If `console.log(e)` isn't working, you can do `console.log(e.body)` instead. But let me know because I'll need to raise another issue for it

hunterachieng:
Screenshot 2024-07-01 at 16 34 08

hunterachieng:
Screenshot 2024-07-01 at 16 39 27

josephjclark (approved):

Issue #630, opened 2024-06-19
`mssql`: Credential investigate on lightning 


mtuchi:
### Description 

If you use the `mssql` credential on lightning it enforce you to use `encrypt` and `trusted certificate`. I need to investigate further why this is happening.

aleksa-krolls:
From Mtuchi: I think the bug i am experiencing using mssql credential was suppose to be fixed with this PR https://github.com/OpenFn/adaptors/pull/512,  But something is still not right with mssql credential

mtuchi:
This PR https://github.com/OpenFn/adaptors/pull/512 was for solving the credential issue

mtuchi:
Okay after a throughly investigation i discovered two things 
1. The configuration-schema for mssql adaptor is missing a title for `port` property 
2. The default values in the configuration-schema are not picked up by lightning on credential form. This is why i was getting an error saying `Server requires encryption, set 'encrypt' config option to true.` while i know for a fact that mssql adaptor set the `encrypt` option to `true` by default

PR #667, opened 2024-07-04
`mssql` update port configuration-schema

mtuchi:
## Summary

Add `title` and `description` for port configuration 

Ref #630 

## Review Checklist

### DIFF
--- a/.changeset/nice-pianos-teach.md
+++ b/.changeset/nice-pianos-teach.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-mssql': patch
----
-
-Add title and description for `port` configuration
--- a/.changeset/nice-pianos-teach.md
+++ b/.changeset/nice-pianos-teach.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-mssql': patch
+---
+
+Add title and description for `port` configuration
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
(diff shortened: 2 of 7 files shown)


josephjclark (approved):
This is cool but what broke?

If Lightning is failing because the schema lacks a name and description, that needs raising as a bug in Lightning.

josephjclark:
Also something is wrong with pnpm.

I think it's temporary, and as there's no code changes here I'm happy to merge with CI failing

mtuchi:
The mssql credential form on lightning was missing a title so when creating a new credential the form input title is empty 
Screenshot 2024-07-05 at 9 08 20 AM

mtuchi:
> @mtuchi is there an issue to cover that on the Lightning side? It should be more robust

@josephjclark i opened up an issue yesterday on lightning but it does not mention the port title issue which will be resolved by this PR, See the issue here https://github.com/OpenFn/lightning/issues/2255

josephjclark:
That's a different issue though.

The issue here is that if a field does not have a title, it'll be totally blank in the UI. You can use port as an example but it affects all future schema

mtuchi:
> That's a different issue though.
> The issue here is that if a field does not have a title, it'll be totally blank in the UI. You can use port as an example but it affects all future schema

Okay i will bring it up in a stand up, I am not sure how lightning is suppose to solve this  but i have created an issue here https://github.com/OpenFn/lightning/issues/2259.

josephjclark:
Thank you. It's not your job to solve the issue. You just have to report it!

josephjclark (approved):

josephjclark:
@mtuchi please run `pnpm changeset version` , commit the diffs, run `pnpm changeset tag`, and push the new commits and tags

mtuchi:
Done i have pushed the new tag

Issue #661, opened 2024-07-03
`primero` Fetching referrals improvements


mtuchi:
### Description 

When fetching cases referrals during job execution the logs doesn't mention were fetching referrals, This might seem like the workflow is stuck while in reality we are fetching case referrals 

### Request in few words 
- Add a log to show we are fetching referrals for the retrieved cases
- Improve documentation for `withReferrals` option

josephjclark:
Yeah, the documentation right now is actually wrong!

We need to correct the docs and add an example with referrals. We should also add a note that referrals generates one request per case and may impact performance.

I think we should fix this immediately while we're looking at it. It won't take long.

josephjclark:
It is annoying that primero seems to have an id to bulk-insert referrals for a number of cases; but has no API to bulk-fetch referrals for a number of cases. Oh well.

aleksa-krolls:
hey @mtuchi you can use the demo cred in LP `Primero V2 Cambodia Demo (api_user)` to test this, and let Aissatou know if you need any assistance testing, creating test data, etc.

mtuchi:
Thanks AK, noted 

> hey @mtuchi you can use the demo cred in LP `Primero V2 Cambodia Demo (api_user)` to test this, and let Aissatou know if you need any assistance testing, creating test data, etc.

PR #674, opened 2024-07-10
`primero` Update getCases function

mtuchi:
## Summary

Add a log when fetching cases referrals and improve the jsdoc

## Issues

Ref #661 
## Review Checklist

### DIFF
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -222,6 +222,7 @@ export function getCases(query, options, callback) {
           );
 
           if (expandedOptions?.withReferrals) {
+            console.log(`Fetching ${cases.length} cases referral`);
             for await (const c of cases) {
               const requestParams = {
                 method: 'GET',
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -176,9 +176,19 @@ function cleanupState(state) {
  *   remote: true,
  *   case_id: "6aeaa66a-5a92-4ff5-bf7a-e59cde07eaaz",
  * });
+ * @example <caption>Get case from Primero with referrals</caption>
+ * getCases(
+ *   {
+ *     remote: true,
+ *   },
+ *   {
+ *     withReferrals: true,
+ *   }
+ * );
  * @function
- * @param {object} query - an object with a query param at minimum, option to getReferrals
- * @param {object} options - (Optional) an object with a getReferrals key to fetch referrals
+ * @param {object} query - an object with a query param at minimum, option to getCases
+ * @param {object} [options] - (Optional) an object with a withReferrals key to fetch referrals
+ * @param {boolean} [options.withReferrals] - A boolean value for fetching case referrals
  * @param {function} callback - (Optional) Callback function
  * @returns {Operation}
  */
@@ -222,7 +232,7 @@ export function getCases(query, options, callback) {
           );
 
           if (expandedOptions?.withReferrals) {
-            console.log(`Fetching ${cases.length} cases referral`);
+            console.log(`Fetching ${cases.length} cases referrals`);
             for await (const c of cases) {
               const requestParams = {
                 method: 'GET',


packages/primero/src/Adaptor.js
```diff
@@ -176,9 +176,19 @@ function cleanupState(state) {
  *   remote: true,
  *   case_id: "6aeaa66a-5a92-4ff5-bf7a-e59cde07eaaz",
  * });
+ * @example <caption>Get case from Primero with referrals</caption>
```
josephjclark, line 179: This should say "get cases" (plural), no? Maybe "Get cases and their referrals" is better

packages/primero/src/Adaptor.js
```diff
@@ -176,9 +176,19 @@ function cleanupState(state) {
...
+ *   },
+ *   {
+ *     withReferrals: true,
+ *   }
+ * );
  * @function
- * @param {object} query - an object with a query param at minimum, option to getReferrals
- * @param {object} options - (Optional) an object with a getReferrals key to fetch referrals
+ * @param {object} query - an object with a query param at minimum, option to getCases
+ * @param {object} [options] - (Optional) an object with a withReferrals key to fetch referrals
```
josephjclark, line 190: I'm not really happy about the clarity of these docs. Would a typedef help?

Biggest thing for me though: we should mention that referrals may take a while. It's probably ok to mention that each referral generates a new request (if we word it well)

josephjclark (changes requested):

packages/primero/src/Adaptor.js
mtuchi: Fixed

packages/primero/src/Adaptor.js
mtuchi: It's very tricky defining the params for `query` object, because so far the only reference i have is what is on the examples for `getCases()`. Maybe i should just document those for now ?

.changeset/afraid-points-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-primero': patch
+---
+
+Improve `getCases()` withReferrals docs
```
josephjclark, line 5: We've added fancy logging too!

packages/primero/src/Adaptor.js
josephjclark: Ok, first off, the extra documentation we've added really doesn't shed much light on what the query options are.

For example: `[query.remote] - A remote option to getCases`

I mean, Ok at least we're declaring that there is a `remote` option now. But we're not explaining what this means.

If  we're going to take the time to add documentation, it has to actually be meaningful, you know?

Also this sentence doesn't make a lot of sense:
```
 @param {object} [query] - (Optional) an object with a query param at minimum, option to getCases
```
I think it's saying that `query` is required, but I don't even think that's true.

Interestingly, the main `getCases` docs does shed light on some of this stuff:

>Note that in many implementations, the remote attribute should be set to true to ensure that only cases marked for remote access will be retrieved. You can specify a case_id value to fetch a unique case and a query string to filter result.

So that's a start. Linking to the right part of Primero docs might also be useful.

Then we've got the `withReferrals` problem, which really shouldn't be in its own object. But I don't want to change that now.

I would simply do this:
```
@param {object} [options] - Additional options
@param {boolean} [options.withReferrals] - Set to true to fetch referrals. This will generate an extra request for each case and may take some time to process.
```

packages/primero/src/Adaptor.js
```diff
@@ -222,7 +235,11 @@ export function getCases(query, options, callback) {
           );
 
           if (expandedOptions?.withReferrals) {
+            console.log(
+              `This might take a while, we are fetching ${cases.length} cases referrals`
```
josephjclark, line 239: I'm not a fan of "this might take a while". It's vague, defensive, a bit apologetic; not a very strong tone to be setting.

Can we just say `Fetching ${cases.length}  referrals...`

packages/primero/src/Adaptor.js
```diff
@@ -222,7 +235,11 @@ export function getCases(query, options, callback) {
...
+            );
             for await (const c of cases) {
+              console.log(`Fetching ${c.id.slice(0, -10) + '...'}, referrals`);
```
josephjclark, line 242: Well we can do this, sure, but I didn't ask for it.

it'd be quite happy to literally just `console.log('Fetching referral...')`. 

If you really want to help the user, try something like `fetching referral 1 of 27...`. Then you don't need the top line at all

packages/primero/src/Adaptor.js
mtuchi: The documentation for this is bugging me as well and mainly because i can't find a good Primero API documentation, here is my suggestion let's remove the query params documentation for now and [plan for primero epic](https://github.com/OpenFn/adaptors/issues/680)

packages/primero/src/Adaptor.js
mtuchi: Okay i agree on `might take a while`. 

But something to be aware is a case id might have more than 1 referrals. So saying `Fetching 2 referrals...` is kinda misleading so maybe `Fetching 2 cases referrals...`?

packages/primero/src/Adaptor.js
mtuchi: Okay let's use `Fetching referral...`. 
Although this is not useful when debugging. I wish i have more control of the logs when fetching referrals. Because when an error happen it's a bit painful figuring out what went wrong. 

But again the new CLI has better error message so maybe it will be easier

mtuchi:
Hiya @josephjclark i have created an issue for primero adaptor overhaul https://github.com/OpenFn/adaptors/issues/680, 
Please have a look and add anything that i am missing

packages/primero/src/Adaptor.js
josephjclark: Those docs look pretty clear to me, and I think our current documentation is completely wrong.

Leave this with me - I'll get it sorted today.

josephjclark (approved):

Issue #670, opened 2024-07-08
`openLMIS` Add OAuth support and generic helpers (get, post, put)


mtuchi:
### Description 
To start, this adaptor should handle authentication via OAuth and support the following operation 

- [x] include a generic GET helper so we can get any resource
- [x]  include a generic POST so we can create any resource
- [x]  include a generic PUT so we can update any resource

**Demo Server Credentials:** 
URL: https://demo-v3.openlmis.org/#!/home
username: **administrator**
password: **password**

OpenLMIS API requests are authenticated by an OAuth2 access token passed as a query parameter. See a sample API request below: 

```
GET /api/reports/templates/requisitions/{id}?access_token=
9a556033-ed13-4dde-9561-158469d15134

URI Parameters
id: required (string)

Query Parameters
access_token: required (string)
OAuth2 access token

```

Link to test OAuth2 token here: https://github.com/OpenLMIS/openlmis-auth/blob/master/src/main/resources/db/demo-data/auth.api_keys.csv

mtuchi:
@christad92 i get the access_token and also i can't login to the https://demo-v3.openlmis.org/#!/home with the given credential

mtuchi:
API documentations
- [API Console](https://test.openlmis.org/auth/docs/#!/default/post_api_oauth_token)
- [API Docs 3.0.2](https://docs.openlmis.org/en/rel-3.0.2/api/index.html)
- [Auth Service](https://github.com/OpenLMIS/openlmis-auth/blob/master/DESIGN.md#auth-service-design)

mtuchi:
I have made good progress on this issue,  See #675 But i am missing unit test which i think it's crucial to have
I have skipped the unit test for now but i will implement them before requesting review from Joe

josephjclark:
@mtuchi I can't log into the demo server either - any tips?

mtuchi:
Ooh yeah @josephjclark i was testing using `https://test.openlmis.org` i forgotten to update the issue 🤦🏽 

Since the adaptor is already released can we close this issue now ?

PR #675, opened 2024-07-12
`openlmis` Add generic helpers

mtuchi:
## Summary

Add generic helper for `openlmis` adaptor. `get`, `post` and `put`

## Issues

Ref #670 

## Review Checklist

- [x] If this is a new adaptor, Added the adaptor on marketing website ? https://github.com/OpenFn/marketing/pull/183


.changeset/rich-beers-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-openlmis': minor
+---
+
+Add generic helpers. `request()`, `post()`, `get()` and `put()`
```
josephjclark, line 5: We don't need a changeset here because this is the first release of this adaptor

Just copy this line into `CHANGELOG.md`

packages/openlmis/configuration-schema.json
```diff
@@ -0,0 +1,71 @@
...
+      "type": "string",
+      "description": "Your user based client secrete",
+      "writeOnly": true,
+      "minLength": 1,
+      "default": "changeme"
+    }
+  },
+  "type": "object",
+  "additionalProperties": true,
+  "required": [
```
josephjclark, line 67: Is this config schema accurate? Are username and password required?

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,132 @@
...
+ * Wraps `language-common/execute`, and prepends initial state for odk.
+ * @example <caption> * execute(</caption>
+ *   get("processingPeriods")
+ * )(state)
+ * @private
+ * @param {Operations} operations - Operations to be performed.
+ * @returns {Operation}
+ * @state {OpenLMISState}
+ */
+export function execute(...operations) {
```
josephjclark, line 33: I love how minimal this is!

Do you think we can make it even smaller? I think this will work?

```
export function execute(...operations) {
   return  commonExecute(util.authorize, ...operations)
}
```

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,132 @@
...
+  return state => {
+    return commonExecute(util.authorize, ...operations)(state);
+  };
+}
+/**
+ * Make a GET request in OpenLMIS
+ * @example <caption>Get all supplyLines</caption>
+ * get("supplyLines");
```
josephjclark, line 41: I'd love this example to look like a path - can we at least do `get("/supplyLines");` ?

The same for the other requests and examples

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+} from '@openfn/language-common/util';
+
+export const authorize = state => {
+  const {
+    baseUrl,
+    access_token,
+    username,
+    password,
+    clientId = 'user-client',
+    clientSecrete = 'changeme',
```
josephjclark, line 15: This must be a typo on `clientSecret`

Why use the `changeme` pattern and not jsut do `if (!clientSecret)`{ console.log() }`?

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+  } = state.configuration;
+
+  const headers = makeBasicAuthHeader(clientId, clientSecrete);
```
josephjclark, line 18: We only need to create the headers if there's no access token AND there's a user name and password. So this line needs to move down the function somewhere

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+  if (access_token) return state;
+
+  if (username && password) {
+    if (clientId === 'user-client')
+      console.warn('Using default client id:', clientId);
+    if (clientSecrete === 'changeme')
+      console.warn('Using default client secrete:', clientSecrete);
+
+    const options = {
+      query: { grant_type: 'password', username, password },
```
josephjclark, line 29: Password goes in the query string? For real?

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+          ...state,
+          configuration: {
+            ...state.configuration,
+            access_token: response.body.access_token,
+          },
+        };
+      })
+      .catch(err => {
+        console.error('Error authenticating with OpenLMIS');
+        console.log(err.body);
```
josephjclark, line 50: Correct me if I'm wrong, but on the latest CLI and Worker there should be no reason to intercept the exception here. The error should be nicely logged. In fact, but console.logging it I think you're duplicating the error reporting.

Can you check against the latest version?

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+        throw err;
+      });
+  }
+
+  // warn if no auth found
+  console.warn(
+    `WARNING: no authentication properties found on congfiguration schema!`
```
josephjclark, line 57: Typo on `configuration`

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+  );
+  console.warn(
+    'Make sure to either set an access_token or username & password'
+  );
+};
+
+export const prepareNextState = (state, response, callback = s => s) => {
+  const { body, ...responseWithoutBody } = response;
+
+  if (!state.references) {
```
josephjclark, line 67: Common does this for you now - you don't need to setup state.references yourself

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+  return callback(nextState);
+};
+
+export const request = (configuration = {}, method, path, options) => {
+  const { baseUrl, access_token } = configuration;
+  const { headers = {}, ...otherOptions } = options;
+
+  const opts = {
+    parseAs: 'json',
+    baseUrl: `${baseUrl}/api`,
```
josephjclark, line 85: As with FHIR, `baseUrl` should just be passed through as a baseURL (I'm SURE that undici will throw for this, is this really working? If so, why?).

You should append `/api` to the path. Although if we do this, I think you need to make it clear in the docs. Because otherwise the first thing I'll do is `get('/api/processingPeriods')` and I'll get an error.

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,132 @@
...
+ * @returns {Operation}
+ * @state {OpenLMISState}
+ */
+export function execute(...operations) {
+  return state => {
+    return commonExecute(util.authorize, ...operations)(state);
+  };
+}
+/**
+ * Make a GET request in OpenLMIS
```
josephjclark, line 39: Like with FHIR, I want to make it super clear that this requests against the base URL.

I'm really worried about how quickly this pattern is spreading through the adaptors. Is there anything we can do in template? Either updating the default docs or adding some kind of boilerplate "throw on absolute url"?

I think I need to get on and add a proper fix for this really soon

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,132 @@
...
+  return request('GET', path, null, options, callback);
+}
+
+/**
+ * Make a POST request in OpenLMIS
+ * @example <caption>Creates new program</caption>
+ * post("programs", { name: "Bukayo", code: "abc" });
+ * @function
+ * @public
+ * @param {string} path - Path to resource
```
josephjclark, line 60: A good solution here (and maybe in template too) would be something like `Path to resource (relative to the base URL defined in configuration)

josephjclark (changes requested):

packages/openlmis/configuration-schema.json
mtuchi: Yes for basic auth, you need to specify `username, password and baseUrl` 
I will add `baseUrl` shortly.

packages/openlmis/src/Utils.js
mtuchi: For user-based credential the default values are 
```
client-id: user-client
secret: changeme
```

Here are the API docs 
- https://docs.openlmis.org/en/rel-3.0.2/api/index.html
- https://github.com/OpenLMIS/openlmis-auth/blob/master/DESIGN.md#auth-service-design

mtuchi: I have fixed the typo

packages/openlmis/src/Adaptor.js
mtuchi: I have updated the examples `path`

packages/openlmis/src/Adaptor.js
mtuchi: Yep that works, i have updated the function

.changeset/rich-beers-smoke.md
mtuchi: Done ✅ , Thanks 👍🏽

packages/openlmis/test/Adaptor.test.js
```diff
@@ -0,0 +1,241 @@
...
+          })
+        : mockResponse(200, {
+            access_token: 'fake-token',
+          });
+    })
+    .persist();
+});
+
+describe('HTTP wrappers', () => {
+  it('request() should expand references', async () => {
```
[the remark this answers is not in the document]
mtuchi: Can't take all the credit, this is borrowed from ODK adaptor 🙌🏽

packages/openlmis/test/Adaptor.test.js
```diff
@@ -0,0 +1,241 @@
...
+
+    const state = {
+      configuration,
+    };
+
+    // prettier-ignore
+    const finalState = await execute(get('programs'))(state);
+
+    expect(finalState.data).to.eql(fixtures.programs);
+    expect(finalState.data[0].id).to.eql(
```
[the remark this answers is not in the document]
mtuchi: No good reason, i borrowed the test from ODK, i will remove the id assertion

packages/openlmis/src/Utils.js
josephjclark: But default credentials should be in the credential, not in the adaptor? I'm not aware that we've ever hard coded credentials into an adaptor and I'm not sure we should start now

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,98 @@
...
+    clientSecret = 'changeme',
+  } = state.configuration;
+
+  const headers = makeBasicAuthHeader(clientId, clientSecret);
+
+  if (access_token) return state;
+
+  if (username && password) {
+    if (clientId === 'user-client')
+      console.warn('Using default client id:', clientId);
```
josephjclark, line 24: I'm not really sure I like this pattern either :thinking: 

Some adaptors say things like "Logging in as <user>" during the auth step. I quite like that. Can we do something like `logging with clientid <x>`, and log either the username or the first 8 characters of the api key?

josephjclark:
@mtuchi thanks for the extra commits but there are still some unresolved comments here. I'm particularly concerned about the password going into the URL query

packages/openlmis/src/Utils.js
mtuchi: Okay i will remove the default and add `clientId` and `clientSecret` in required list in `configuration-schema`

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,132 @@
...
+ * @private
+ * @property data - the parsed response body
+ * @property response - the response from the HTTP server, including headers, statusCode, body, etc
+ * @property references - an array of all previous data objects used in the Job
+ **/
+
+/**
+ * Options provided to the HTTP request
+ * @typedef {Object} RequestOptions
+ * @property {object} body - body data to append to the request. JSON will be converted to a string (but a content-type header will not be attached to the request).
```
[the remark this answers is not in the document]
mtuchi: I think the description was miss leading, i have updated the description for body.We are parsing as `json` and the `content-type` is set to `application/json` in `util.request()`

packages/openlmis/src/Adaptor.js
mtuchi: I love that description, let me adapt that in our docs

packages/openlmis/src/Utils.js
mtuchi: Well technically we are requesting an `access_token` then add it to the header. I have removed the logs completely to avoid confusion

mtuchi:
@josephjclark regarding passing password into `URL query`. That is how you create a new `access_token` in openlmis, See their API here https://test.openlmis.org/auth/docs/#!/default/post_api_oauth_token

In Authentication 
username: user-client
password: changeme

Then in query params 
grant_type: password
username: administrator
password: password

packages/openlmis/src/Utils.js
mtuchi: Yep, i was very confused as well 😕

packages/openlmis/src/Utils.js
mtuchi: It's working, if i append the `/api` in `baseUrl`. If i prepend the `/api` in path. Eg `/api/${path}` It fails for operations like `get('/processingPeriods')` because the url is invalid.

packages/openlmis/src/Utils.js
mtuchi: Removed it ✅

packages/openlmis/src/Utils.js
josephjclark: Can they not be defaulted in the configuration schema? Json schema has defaults right?

packages/openlmis/src/Adaptor.js
mtuchi: I have created a helper function in `Utils.js` called `urlMatchesBase(path, baseUrl)`. This function will throw an error `The URL ${path} does not match the base URL ${baseUrl}`);` If the path url doesn't match the baseUrl

packages/openlmis/src/Utils.js
mtuchi: Your right, i was testing using the old CLI version, I have tested with CLI `1.6.0` and the CLI logs the error nicely. 

I have removed the extra error logging

josephjclark:
@mtuchi I don't think that doc link works. At least, I can't work out anything useful from it.

If the system uses the query string to log in then, well, I guess that's just how it works! But I'm surprised, it seems super insecure to me -  putting an unencrypted password in the URL like that :shrug:

packages/openlmis/src/Utils.js
mtuchi: Yes i have put the default for both `clientId` and `clientSecret` in `configuration-schema`

packages/openlmis/src/Utils.js
josephjclark: @mtuchi  looking at this now, I still see:

```
console.error('Error authenticating with OpenLMIS');
console.log(err.body);
throw err
```

I don't think we want to log err.body?

packages/openlmis/src/Adaptor.js
```diff
@@ -0,0 +1,130 @@
...
+ * Options provided to the HTTP request
+ * @typedef {Object} RequestOptions
+ * @property {object} body - body data to append to the request.
+ * @property {object} query - An object of query parameters to be encoded into the URL.
+ * @property {object} headers - An object of headers to append to the request.
+ */
+
+/**
+ * Execute a sequence of operations.
+ * Wraps `language-common/execute`, and prepends initial state for odk.
```
josephjclark, line 24: :exclamation: This mentions ODK

Can we search for any `odk` strings in `packages/openlmis`?

packages/openlmis/src/Utils.js
josephjclark: But we can fix that! See the last comment in #578
```
import nodepath from 'node:path'
...
const safepath = nodepath.join('api', path)
```
`path.join` will build a safe path with slashes in the right place

As it stands at the moment, it'll still fail if there's a trailing slash on the baseUrl in config, which isn't robust enough

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+    ...otherOptions,
+  };
+
+  return commonRequest(method, path, opts);
+};
+
+function urlMatchesBase(path, baseUrl) {
+  const base = new URL(baseUrl);
+  const url = new URL(path, baseUrl);
```
josephjclark, line 86: Is this right?

If path is `https://www.google.com` and base url is `http://openlmis.org`, I suspect `new URL` will throw

I'd like to see a bunch of unit tests against this function to be satisfied that it's working correctly

packages/openlmis/src/Utils.js
mtuchi: Ooh i see, okay i have adapted that approach of creating safe path using nodepath

packages/openlmis/src/Utils.js
mtuchi: Yeah your right, i have remove this catch entirely. The error log that i am getting for `401` is more than enough

packages/openlmis/src/Utils.js
mtuchi: It does throw, I have added one unit test. Can you recommend more test scenarios ?

mtuchi:
@josephjclark this is a quick demo video for [how to generate an access_token](https://www.loom.com/share/e7c7489bdb164ca6b058bca5315c63b9?sid=ffdf7b17-4292-4728-a601-fe36c2ec975d). You can also use any Rest API client. like postman to 

Here is the cURL example 
```
curl  -X POST \
  'https://test.openlmis.org/api/oauth/token?grant_type=password&username=administrator&password=password' \
  --header 'Authorization: Basic dXNlci1jbGllbnQ6Y2hhbmdlbWU='
```

The basic auth is `user-client:changeme`

packages/openlmis/test/Adaptor.test.js
```diff
@@ -0,0 +1,256 @@
...
+    expect(error.statusMessage).to.eql('Bad Request');
+    expect(error.message).to.eql(
+      'POST to https://test.openlmis.org/api/programs returned 400: Bad Request'
+    );
+    expect(error.statusCode).to.eql(400);
+  });
+});
+
+describe('Url Matches BaseUrl', () => {
+  it('throw an error when url origin does not match baseUrl', async () => {
```
josephjclark, line 241: this test should be moved inside the `post` block. It should not be in its own describe

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,88 @@
+import nodepath from 'node:path';
+import { composeNextState } from '@openfn/language-common';
+import {
+  request as commonRequest,
+  makeBasicAuthHeader,
+} from '@openfn/language-common/util';
+
+export const authorize = state => {
```
josephjclark, line 8: I've been running some tests locally and I notice the auth step is super slow - at least when failing.

Can we add a log line to say "logging in as <username>" or "logging in with access token" or something?

I'm sure you added that originally? Maybe it was a different PR because I can't see it now.

Anyway, a little bit of feedback here would be super helpful

josephjclark: I wonder if we can be putting out a more useful error message as well. Can we print the client id and secret? Or warn if one is un-set?

josephjclark (changes requested):
I'm trying to run a few local tests but the test server is reporting 502 (bad gateway) errors. Even when I log in through the web UI.

So I'll have to try again later.

I've got a couple of little comments in the meantime.

packages/openlmis/src/Utils.js
mtuchi: I see 🤔, It's a tricky one 

I think the best way to think about it is as long as your authenticated to a service you can't use any absolute URL with exception of adaptor that already export common http. Those one can still make request to other service. Eg `http.get({url:'https://swapi.dev/api/'})`. 

I am saying authenticated because in http adaptor for example you can use the adaptor without a authentication to any service

packages/openlmis/src/Utils.js
```diff
@@ -0,0 +1,88 @@
...
+
+  return commonRequest(method, safepath, opts);
+};
+
+function urlMatchesBase(path, baseUrl) {
+  const base = new URL(baseUrl);
+  const url = new URL(path, baseUrl);
+
+  if (url.origin !== base.origin) {
+    throw new Error(`The URL ${path} does not match the base URL ${baseUrl}`);
```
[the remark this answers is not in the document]
mtuchi: Okay got you,  Which of the following is better 
```
`Requests are only allowed to the base URL ${baseUrl}. Please ensure you are sending requests to the correct URL.`;
```

or 
```
`Requests must be sent to the base URL: ${baseUrl}`
```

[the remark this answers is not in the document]
josephjclark: Let's do the second one, it's shorter while still being clear

`Requests must be sent to the base URL: ${baseUrl}`

packages/openlmis/src/Utils.js
mtuchi: I have added logs for `logging`, `clientId` and `clientSecret`

packages/openlmis/src/Utils.js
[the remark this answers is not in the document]
mtuchi: Done ✅

packages/openlmis/src/Utils.js
josephjclark: Yeah, agreed.

For any adaptor with a baseURL (including http), don't allow absolute URLs to be fetched at all.

I've worked out a plan this morning and I'm going to go off and action it ASAP. So we'll keep this here for now and if it changes in a few days, no problem

packages/openlmis/src/Utils.js
josephjclark: That's great, nice one

josephjclark (approved):

josephjclark:
@mtuchi this is ready to go and I've pushed the tag :tada: 

At the moment it's aimed at an epic branch, not main. If you want to release it now, I'll need to merge it into main. That cool?

mtuchi:
Hey @josephjclark yeah go ahead and merge.

PR #676, opened 2024-07-12
Msgraph oauth normalize

mtuchi:
## Summary

## Details

## Issues

## Review Checklist

### DIFF
--- a/.changeset/funny-hairs-clap.md
+++ b/.changeset/funny-hairs-clap.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-msgraph': minor
+---
+
+Normalize configuration keys
--- a/packages/msgraph/src/Adaptor.js
+++ b/packages/msgraph/src/Adaptor.js
@@ -1,5 +1,8 @@
 import { execute as commonExecute } from '@openfn/language-common';
-import { expandReferences } from '@openfn/language-common/util';
+import {
+  expandReferences,
+  normalizeOauthConfig,
+} from '@openfn/language-common/util';
 
 import {
   request,
@@ -43,6 +46,7 @@ export function execute(...operations) {
     return commonExecute(...operations)({
       ...initialState,
       ...state,
+      configuration: normalizeOauthConfig(state.configuration),
     })
       .then(cleanup)
       .catch(error => {


josephjclark (approved):
Well this is fine and I'm happy to release

But I'm curious  - why not just update the configuration schema? We've agreed on other adaptors that `access_token` is the way to go. Feels strange to work around it like this

mtuchi:
@taylordowns2000 @elias-ba have more insight on the lightning side, But i think configuration schema is not used for OAuth credentials on lightning. (I could be wrong 🤷🏽 ) @elias-ba do you know if configuration-schema is used for `OAuth` credentials on lighting ?

taylordowns2000 (approved):
I'm OK with this. Thank you.

Before merging, @josephjclark , can you confirm that we want to use this `normalizeOauthConfig` function to change `access_token` (which we get in the `.json` from most Ouath providers) to `accessToken` which _we prefer to use_ in the `.js` adaptor code?

I think that's the intention, but wanted to double check.

elias-ba:
@mtuchi json schema is not used for oauth in Lightning. We may need it in the future to dynamically render custom inputs like api_version or scope, ...

josephjclark:
@taylordowns2000 I think we need to have a chat about oauth (and schemas) and make sure all the rules are documented somewhere  in the adaptor wiki or readme. There's still a few things I'm not certain of.

I would prefer to just use `access_token` everywhere. It's a bit weird from a pure JS point of view but that's the oath standard name for the key - if we just use that, everything else gets a little simpler. Plus it's only used by like one line of code per adaptor.

If there's a rush I'm happy to use this mapping. If we have a bit of time I'd rather just get it right once and for all.

taylordowns2000:
> I would prefer to just use access_token everywhere

@josephjclark , agreed. I thought this _was_ the decision we made, and my comment here (see below) was probably me trying to re-write my memory and get on board with the new thing!

> confirm that we want to use this normalizeOauthConfig function to change access_token (which we get in the .json from most Ouath providers) to accessToken which we prefer to use in the .js adaptor code?

The adaptors expect a JSON object to be handed to them by `fetch:credential` (or loaded from a `configuration.json` file). the shape of those JSON objects are determined by the systems they connect to.... we shouldn't have _that much_ of a say in terms of what they look like.

In the case of OAuth2, we know we'll always get `access_token`. A few months back I think we updated most of the adaptors so that they expect `access_token` and we missed msGraph.

I'd like to merge this now and then open an issue to, once and for all*, fix all the adaptors so that even if they use responsible "whoopsie handling" code like `normalizeOauthConfig`, they're still built to expect `access_token`.



*well not really... i mean... things will keep evolving i'm sure 😄

josephjclark:
@taylordowns2000 Fine by me!

@mtuchi we have a failing test - we do need to fix that before merging please

You can fix the test, then bump versions and tags, and @ me for a final review before we merge

mtuchi:
@josephjclark i have fixed the broken tests

josephjclark:
Version bumps were not included here. I have fixed directly on main

Issue #569, opened 2024-05-17
commcare: add auto pagination to `get`


josephjclark:
At the moment, in commcare users have to handle pagination manually.

## API design

I think that by default we should autofetch. Ie, pull down all the records.

We should pull down in batches defined by `limit`, which users can customize but we default to something quite high like 1000.

If an offset is passed (including `offset: 0`), we do not auto-page and assume that users will take control

The callback function should be invoked for each page (check this - I think this is what other chunking style operations do?)

I wonder if we should have some kind of `maxlimit`, set to a million or something, and if there are more records than that limit, we throw an error and say "woah, there's too much data here, you need to do something". But I don't know, I'd rather we had a library, platform and runtime capable of processing all that data. And batching up should reduce the overhead anyway.

## How to do it today

I see two strong patterns to paginate commcare (thanks to Professor @mtuchi!)

1) Use edge conditions to run the job again

```
state.meta.next // if state.metada.next is truthy, there is more data and we should re-run from the next offset
```

This is a really nice way to use the workflow to loop.

2) Get the total, generate batches, and iterate over them

```
// fetch no data
get('cases', { limit: 0, offset: )`

// read the total count
fn((state) => {
  const count = state.meta.total_count

 // work out all the offsets we need to fetch one page at a time
  const offsets = [0]
  for(const i = 1000;i<total_count;i+=1000) {
    offsets.push(i)
  }
 state.data.offsets = offsets
 return statte
})

each($.data.offsets, get('case'), { offset: $.data )
```

PR #678, opened 2024-07-16
commcare: Implement pagination on commcare's get request

hunterachieng:
## Summary

Implement pagination on commcare`get`

## Details

A manual implementation of pagination of data fetched from commcare. Users can override the default pagination when they pass in `offset` and/or `limit` 

## Issues

#569 

## Review Checklist

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -126,6 +126,7 @@ export function get(path, params = {}, callback = s => s) {
           break;
         }
       } while (allowPagination);
+      
       return {
         ...nextState,
         data: result,
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -68,11 +68,7 @@ export function execute(...operations) {
  * @param {function} [callback] - Optional callback to handle the response
  * @returns {Operation}
  */
-export function get(
-  path,
-  params = { offset: 0, limit: 1000 },
-  callback = s => s
-) {
+export function get(path, params = {}, callback = s => s) {
   return async state => {
     const { domain } = state.configuration;
     const [resolvedPath, resolvedParams] = expandReferences(
@@ -81,42 +77,50 @@ export function get(
       params
     );
 
+    let { offset = 0, limit = 1000 } = resolvedPath;
+
     let allItems = [];
+    let items;
     let next;
     let nextState = state;
-    let offset = resolvedParams.offset;
-    const limit = resolvedParams.limit;
 
     try {
       do {
-        // fetch a page
+        let requestParams = {
+          ...resolvedParams,
+        };
+
+        if (next) {
+          requestParams = {
+            ...requestParams,
+            offset,
+            limit,
+          };
+        }
 
         const response = await request(
           state.configuration,
           `/a/${domain}/api/v0.5/${resolvedPath}`,
           {
             method: 'GET',
-            params: {
-              ...resolvedParams,
-              offset: offset,
-              limit: limit,
-            },
+            params: requestParams,
             contentType: 'application/json',
           }
         );
         nextState = prepareNextState(state, response, callback);
 
-        if (Array.isArray(nextState.data)) {
+        if (next) {
           allItems.push(...nextState.data);
         } else {
-          allItems.push(nextState.data);
+          items = nextState.data;
         }
         next = response?.body?.meta?.next;
+        limit = response?.body?.meta?.limit;
         offset = limit + offset;
       } while (next);
       return {
         ...nextState,
-        data: allItems,
+        data: items || allItems,
       };
     } catch (e) {
       if (e.statusCode === 404) {
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -90,12 +90,17 @@ export function get(
     try {
       do {
         // fetch a page
+
         const response = await request(
           state.configuration,
           `/a/${domain}/api/v0.5/${resolvedPath}`,
           {
             method: 'GET',
-            params: resolvedParams,
+            params: {
+              ...resolvedParams,
+              offset: offset,
+              limit: limit,
+            },
             contentType: 'application/json',
           }
         );
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -56,8 +56,8 @@ export function execute(...operations) {
 
 /**
  * Make a GET request to any commcare endpoint. The returned objects will be written to state.data.
- * Not adding an offset param when getting items will trigger an automatic pagination and the returned data is combined and written into state.data
- * If an offset is provided, no auto-pagination occurs
+ * Unless an `offset` is passed, `get()` will automatically pull down all pages of data if the response
+ * is paginated.
  * A `response` key will be added to state with the HTTP response and a `meta` key
  * @public
  * @example <caption>Get a list of cases</caption>
@@ -67,7 +67,7 @@ export function execute(...operations) {
  * @function
  * @param {string} path - Path to resource
  * @param {RequestQueries} params - Optional request params such as limit and offset.
- * @param {function} [callback] - Optional callback to handle the response. During pagination, the callback is called on each pagination request made
+ * @param {function} [callback] - Callback invoked once per page of data retrieved.
  * @returns {Operation}
  */
 export function get(path, params = {}, callback = s => s) {
@@ -83,6 +83,8 @@ export function get(path, params = {}, callback = s => s) {
 
     let nextState = state;
     let result;
+
+    // Automatically paginate if the user did not pass an offset
     let allowPagination = isNaN(resolvedParams.offset);
 
     try {
@@ -103,9 +105,9 @@ export function get(path, params = {}, callback = s => s) {
         );
 
         nextState = prepareNextState(state, response, callback);
+        // If the server tells us there's another page of data, setup
+        // the next request to get it
         if (response?.body?.meta?.next) {
-          // Check if next is present and make another call
-
           if (!result) {
             result = [];
           }
@@ -124,11 +126,10 @@ export function get(path, params = {}, callback = s => s) {
           } else {
             result = nextState.data;
           }
-          //  Break when next is null
+          // Exit the loop when no more data is available
           break;
         }
       } while (allowPagination);
-      // Make another call if offset is missing in the params
 
       return {
         ...nextState,
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -77,23 +77,48 @@ export function get(path, params = {}, callback = s => s) {
       params
     );
 
-    try {
-      const response = await request(
-        state.configuration,
-        `/a/${domain}/api/v0.5/${resolvedPath}`,
-        {
-          method: 'GET',
-          params: resolvedParams,
-          contentType: 'application/json',
-        }
-      );
+    let allItems = [];
+    let next;
+    let nextState = state;
+    let offset = 0;
+    const limit = 1000;
 
-      return prepareNextState(state, response, callback);
+    try {
+      do {
+        const requestParams = {
+          offset: offset,
+          limit: limit,
+          ...resolvedParams,
+        };
+        // fetch a page
+        const response = await request(
+          state.configuration,
+          `/a/${domain}/api/v0.5/${resolvedPath}`,
+          {
+            method: 'GET',
+            params: requestParams,
+            contentType: 'application/json',
+          }
+        );
+        nextState = prepareNextState(state, response, callback);
+
+        if (Array.isArray(nextState.data)) {
+          allItems.push(...nextState.data);
+        } else {
+          allItems.push(nextState.data);
+        }
+        next = response?.body?.meta?.next;
+        offset = limit + offset;
+      } while (next);
+      return {
+        ...nextState,
+        data: allItems,
+      };
     } catch (e) {
       if (e.statusCode === 404) {
         e.body = { error: `Resource ${path} not found` };
       }
-      throw e.body.error ?? e;
+      throw e;
     }
   };
 }
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -79,19 +79,19 @@ export function get(path, params = {}, callback = s => s) {
       params
     );
 
-    let { offset = 0, limit = 1000 } = resolvedParams;
+    let offset, limit;
 
     let nextState = state;
     let result;
     let allowPagination = isNaN(resolvedParams.offset);
 
-
     try {
       let requestParams = {
         ...resolvedParams,
       };
-  
+
       do {
+        // Make the first request
         const response = await request(
           state.configuration,
           `/a/${domain}/api/v0.5/${resolvedPath}`,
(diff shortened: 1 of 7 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -68,7 +68,11 @@ export function execute(...operations) {
  * @param {function} [callback] - Optional callback to handle the response
  * @returns {Operation}
  */
-export function get(path, params = {}, callback = s => s) {
+export function get(
+  path,
+  params = { offset: 0, limit: 1000 },
```
josephjclark, line 73: I'm not sure these defaults are set the correct way.

If the user does `get('blah', { offset: 10 })`, then limit won't be defaulted.

I think after we expand the params, we should deconstruct and default them.

We also need to change the params passed into the request to ensure it uses the deconstructed values

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +81,48 @@ export function get(path, params = {}, callback = s => s) {
...
     try {
-      const response = await request(
-        state.configuration,
-        `/a/${domain}/api/v0.5/${resolvedPath}`,
-        {
-          method: 'GET',
-          params: resolvedParams,
-          contentType: 'application/json',
+      do {
+        // fetch a page
```
josephjclark, line 92: This comment doesn't seem terribly useful

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +81,48 @@ export function get(path, params = {}, callback = s => s) {
...
+              ...resolvedParams,
+              offset: offset,
+              limit: limit,
+            },
+            contentType: 'application/json',
+          }
+        );
+        nextState = prepareNextState(state, response, callback);
+
+        if (Array.isArray(nextState.data)) {
```
josephjclark, line 109: I am a bit uncomfortable with this.

`get()` is generic. Whatever the server sends to us, we should write to `state.data`.

If we get a single case, state .data should not be an array. But this logic here forces it into an array.

Basically the `allItems` array isn't necessarily the answer here. We only want `allItems` if there's pagination.

packages/commcare/test/index.js
```diff
@@ -102,18 +101,17 @@ describe('getCases', () => {
   it('should fetch cases', async () => {
```
josephjclark, line 101: I don't expect this test to change in this PR. We shouldn't be adding limits and offsets.

packages/commcare/test/index.js
```diff
@@ -159,21 +157,21 @@ describe('getCases', () => {
...
     expect(data.length).to.equal(1);
     expect(data[0]).to.haveOwnProperty('case_id');
     expect(response.meta.limit).to.equal(1);
   });
-
   it('should fetch a single case', async () => {
     testServer
       .intercept({
-        path: `/a/${domain}/api/v0.5/case/12345`,
+        path: /\/a\/my-domain\/api\/v0\.5\/case\/12345/,
```
josephjclark, line 171: We shouldn't need a regex here really. In fact, when requesting a single case, we shouldn't add `offset` and `limit` to the URL at all.

This test has alerted me that there's something amiss in the implementation, and we need think about our strategy a little harder.

packages/commcare/test/index.js
```diff
@@ -216,19 +214,21 @@ describe('getCases', () => {
...
-    expect(data.properties.case_type).to.equal('pregnancy');
+    expect(data[0].case_id).to.equal('12345');
+    expect(data[0].properties.case_type).to.equal('pregnancy');
   });
 
   it('should fetch cases and call the callback', async () => {
     testServer
       .intercept({
-        path: `/a/${domain}/api/v0.5/case`,
+        path: /\/a\/my-domain\/api\/v0\.5\/case/,
```
josephjclark, line 224: I don't think we need a regex path AND a query object. Ideally path and query would be fine.

Tbh I still think the undici mock should match a path and ignore query parameters. I think we need to investigate that a little further.

packages/commcare/test/index.js
```diff
@@ -1,7 +1,6 @@
 import { expect } from 'chai';
 import { enableMockClient } from '@openfn/language-common/util';
 import { execute, submitXls, get, post } from '../src';
-import sinon from 'sinon';
```
josephjclark, line 4: Is sinon listed in this adaptors dev dependencies? And if so, can we remove it?

packages/commcare/test/index.js
```diff
@@ -319,6 +319,54 @@ describe('getCases', () => {
       },
     ]);
   });
+
+  it('should fetch cases with pagination', async () => {
```
josephjclark, line 323: I would like this test to also check that the callback is called three times, once for each item

packages/commcare/test/index.js
```diff
@@ -319,6 +319,54 @@ describe('getCases', () => {
...
+        hostUrl,
+        domain,
+        appId: app,
+        username: 'user',
+        password: 'password',
+      },
+    };
+
+    const { data, response } = await execute(
+      get('case', { limit: 1, offset: 0 })
```
josephjclark, line 364: If we're testing auto-pagination, we should not pass an offset here.

josephjclark (changes requested):
Ah I'm afraid we have quite a big problem here :(

The PR at the moment forces pagination, even on requests for single items, and even if the user passes the offset through.

Firstly, the design in the issue  is that if  the user passes an `offset`,  we don't auto page, because we assume the user is taking control. (We don't care about `limit`, we should still auto-paginate if the user specifies a limit). This behavior also needs a unit test.

Secondly, not all endpoints support pagination (presumably). We shouldn't force `offset` and `limit` to be passed on the first `get` request - we should only start doing that stuff if the response contains `meta.next`. When we see `meta.next`, we can start thinking about autopaging - but not before.

Also, There is no documentation to explain the auto pagination behaviour at all. Also the docs on the callback function haven't been updated to explain that, if paginating, the callback is called once per page.

packages/commcare/src/Adaptor.js
```diff
@@ -56,6 +56,7 @@ export function execute(...operations) {
 
 /**
  * Make a GET request to any commcare endpoint. The returned objects will be written to state.data.
+ * Adding an offset param will trigger an automatic pagination and the returned data is combined and written into state.data
```
josephjclark, line 59: I'm afraid these docs are incorrect. Check the spec in the issue - is this the correct behaviour?

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
       params
     );
 
+    let { offset = 0, limit = 1000 } = resolvedParams;
```
josephjclark, line 81: We should not default the limit here because it makes pagination weird (we only add the default on the second request, the first request will use the system default).

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+
+    let nextState = state;
+    let result;
+    let allowPagination = !isNaN(resolvedParams.offset);
```
josephjclark, line 85: I think this is backwards? If there's NO offset, we auto-paginate.

Unit tests should confirm this either way. Tests are passing so either the tests are wrong, or have insufficient coverage

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+
     try {
-      const response = await request(
-        state.configuration,
-        `/a/${domain}/api/v0.5/${resolvedPath}`,
-        {
-          method: 'GET',
-          params: resolvedParams,
-          contentType: 'application/json',
+      let requestParams = {
```
josephjclark, line 88: hmm. This won't use the default values we've assigned.... maybe that's ok?

packages/commcare/src/Adaptor.js
josephjclark: ah, actually, we basically need to delete this line entirely

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+          }
+        );
+
+        nextState = prepareNextState(state, response, callback);
+
+        if (response?.body?.meta?.next) {
+          if (!result) {
+            result = [];
+          }
+          offset = response?.body?.meta?.offset + response?.body?.meta?.limit;
```
josephjclark, line 109: We don't need to use the optional chaining operator `?` so much here (and in other lines).  If we get  to this line, we know that `response.body.meta.next` is an object

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+          limit = response?.body?.meta?.limit;
+
+          requestParams = {
+            ...requestParams,
+            offset,
+            limit,
+          };
+          result.push(...nextState.data);
+        } else {
+          if (result) {
```
josephjclark, line 119: My spidey sense looks at this block and says it's too complicated and badly structured. I wouldn't mind rethinking how `result` is handled.

The clue is that `result.push` is repeated. And anyway over this six line block, all we're really doing is assigning the result

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+            result.push(...nextState.data);
+          } else {
+            result = nextState.data;
+          }
+
+          break;
```
josephjclark, line 125: Why are we breaking here? I'd love a comment

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
-        {
-          method: 'GET',
-          params: resolvedParams,
-          contentType: 'application/json',
+      let requestParams = {
+        ...resolvedParams,
+      };
+
+      do {
+        const response = await request(
```
josephjclark, line 93: High level feedback: this do/while loop is kinda complicated. But if it's not too bad if you sort of know what it's doing (get data, check to see if we should auto-page it, if so go again and merge results, otherwise just return the data).  So I'd love to see one or two lines of comment which broadly explain the strategy to the next developer coming into the code.

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+            result = nextState.data;
+          }
+
+          break;
         }
-      );
-
-      return prepareNextState(state, response, callback);
+      } while (allowPagination);
+      return {
```
josephjclark, line 128: I would like to see a clear  line break between the `while` and the `return`.

The return is really important but because the code is too dense, it's a little hard to read.

packages/commcare/src/Adaptor.js
```diff
@@ -77,23 +78,62 @@ export function get(path, params = {}, callback = s => s) {
...
+        nextState = prepareNextState(state, response, callback);
+
+        if (response?.body?.meta?.next) {
+          if (!result) {
+            result = [];
+          }
+          offset = response?.body?.meta?.offset + response?.body?.meta?.limit;
+          limit = response?.body?.meta?.limit;
+
+          requestParams = {
```
josephjclark, line 112: I would probably not spread this, I would just mutate the values I want to change. I think that's a bit more declarative?

```
requestParams.offset = offset
requestParams.limit = limit
```

I'd maybe add a comment too, like "set up pagination for the the next request"

packages/commcare/test/index.js
```diff
@@ -319,6 +326,107 @@ describe('getCases', () => {
       },
     ]);
   });
+
+  it('should fetch cases with pagination', async () => {
```
josephjclark, line 330: This is in the `getCases` describe block, but it's not a `getCases ` test.

This is a test of `get`. 

Also, the test name is wrong. We don't care about `fetching cases` here, we care about `GET with pagination`.

Please restructure the tests so that the right functions are tested in the right place. And for any `get` tests, make sure the names are appropriate.

packages/commcare/test/index.js
```diff
@@ -319,6 +326,107 @@ describe('getCases', () => {
...
+            next: next,
+            offset: offset,
+            previous: null,
+            total_count: 3,
+          },
+          objects: [objects[offset]],
+        };
+      });
+
+    testServer
```
josephjclark, line 363: You really shouldn't need to duplicate this intercept block three times. What's going on here?

This is not just me being picky: the duplication of these intercepts makes it really hard to understand the test

packages/commcare/test/index.js
josephjclark: This comment still stands, and if this test is passing we have a problem. The `get()` function here should not paginate.

packages/commcare/test/index.js
josephjclark: I think we need many more tests on pagination too:

```
should not add limit and offset to the parameters of the first request if the user does not pass them
should fetch the next page if the response has a meta.next property
should not auto-fetch if the user sets an offset
should respect the user's limit while paginating
should call the callback once for a single request
should call the callback once per page
should stop paginating after the first error
should not force a non-paginated response into an array
```

Can you think of any more?

josephjclark (commented):
I've got some comments. Mostly stylistic at this point but I do have a couple of questions about whether this function is working as planned.

To that end I've started focusing more on the tests, and I've got a bunch of really important comments on the tests.

Why are tests important? Well, they have several really important jobs:

1) They ensure that specific parts of the code are working as designed
2) They reveal and explain the design of the code.

So if I, as a developer or user of this API, want to better understand how pagination in `get` works, one really good way to do that is to look at the tests. The name of the tests, the test code, the API usage and the assertions should all show me the expected usage of the API.

In this specific case, I've just reviewed `get()` and seen something in the code which makes me say "huh, is that really doing the right thing?". To answer that question,  I want to look at the test code and see whether this particular behaviour is covered.

And the really big problem I have right now is: because the test code is very poorly  organised, I struggle to even find the test test (let alone read and understand it).

So maybe the next step is to focus on the unit tests for a while - once those are right, we can be confident that the operation is behaving properly.

I've suggested a bunch of tests but you'll need to adjust the structure first

packages/commcare/test/index.js
```diff
@@ -226,9 +221,12 @@ describe('getCases', () => {
       .intercept({
         path: `/a/${domain}/api/v0.5/case`,
         method: 'GET',
+        query: {
```
josephjclark, line 224: I can't really tell where this is in the diff, but should this test have changed?

josephjclark (changes requested):
see comments above

packages/commcare/src/Adaptor.js
hunterachieng: @josephjclark  If we change this then the tests will change as well. Let me make it that if we do not have `offset`, then we autopaginate

packages/commcare/test/index.js
```diff
@@ -321,6 +409,256 @@ describe('getCases', () => {
...
+    const callback = state => {
+      callbackArgCount++;
+      return state;
+    };
+
+    await execute(get('case', {}, callback))(state);
+    expect(callbackArgCount).to.deep.equal(1);
+  });
+
+  it.skip('should stop paginating after the first error', async () => {
```
josephjclark, line 630: @hunterachieng why is this skipped? Is this something we need to talk about?

josephjclark:
@hunter just a heads-up that I've pushed a commit with some light documentation. You'll have to pull before you can push again.

josephjclark (approved):

josephjclark:
@hunterachieng I've made a couple of super light edit to the tests and I am happy with this

Just to note that I've made this a minor change, rather than a major one. The outward API hasn't changed at all. The behaviour of the function has changed a bit but I don't think I consider it breaking, and yet emotionally it fees like a feature-add rather than a patch. So I think minor.

Are you familiar with semver by the way? It's worth a real quick look at https://semver.org/

Issue #662, opened 2024-07-04
salesforce: update API


josephjclark:
New API design for salesforce, including adding `composeNextState` and removing old code.

## Delivery Plan

We should break this down into three parts:

* Remove old/unused functions, fix tests
* Use composeNextState and fix state.references, update tests
* Update API to new signatures, with basic docs changes

Each part should be treated like its own story. It should have its own PR which is reviewed and merged into an epic branch called `epic/salesforce`.

## Overview

All functions should use `composeNextState` and write to `state.data`.

`references` behaviour should be preserved, but `state.data` should be the primary usage from now on.

We should probably go on and fix #509 here because we're going to be working so close to that problem.

Only basic documentation is included in this issue. We'll do a deep dive on docs later.

We are going to be bold and REMOVE callbacks entirely from the API :scream: This may result in worse performance for this adaptor version on v1, but I think that's acceptable. And `then()` support, which is coming super soon, will empower this adaptor on v2.

It makes sense to use the new expandReferences API, but that's not technically part of the issue.

Note that the HTTP helpers (`request`, `get`, `post`) should use `connection.request`. Docs should make it super clear that these helpers only call _salesforce_ APIs. Later, we will add a generic HTTP API with common (see #668)

This work should only have minimal unit tests. We'll do a deep dive on tests in #666 

Errors should be thrown, not caught and logged , unless there is any specific messaging that can be added 

## API

```
bulkQuery(query, options)

bulk(operation, sObjectName, records, options)

create(sObjectName, records) // use records instead of attrs. Default to batch.

describe(sObjectName)

describeAll()

destroy(sObjectName, ids, options) // rename attrs (which is wrong) to ids

insert(sObjectName, records)

query(soql, options)

request(url, options)

get(url, options)

post(url, data, options)

retrieve(sObjectName, id)

update(sObjectName, records)

upsert(sObjectName, externalId, records)

toUTF8(as-is)

fnIf() // exported out of common
```
## Code Removal

To be clear, the list above is the whole API. The following functions should all be **removed**.

   - `cleanupState()`
   - `steps()`
   - `reference()`
   - `createIf()`, `upsertIf()` 
   - describeAll()  maybe?
   - remove axios 
   - remove beta from common
   - remove relationship

josephjclark:
This may fix #644

mtuchi:
Just a thought exercises 🧠 , Should we keep exporting the following function from `@openfn/language-common` 🤔 ?
- dataPath,
[...]

PR #682, opened 2024-07-19
`salesforce` API Update

mtuchi:
## Summary

Lot's of cleanup in salesforce adaptor. Ref #662 

## Details

- Update function signature for `bulk()`.
- Rename ~attrs~ to `records`
- Add a default for `options = {}`
- Remove axios dependency
- Remove old/unused functions. `relationship`, `upsertIf`, `createIf`,
  `reference`, `steps`, `beta`
- Standardize state mutation in all operation
- Change `bulk` signature to `bulk(operation, sObjectName, records, options)`
- Remove callback support
- Slightly clean up jsdocs 

## Review Checklist


mtuchi:
Hiya @josephjclark this is ~technically~  almost done, I need to create a changeset and update the PR description to cover what has been removed 
cc @aleksa-krolls

mtuchi:
Hiya @josephjclark this is ready for review, just a couple additional notes

- I have created an issue for moving `state.connection` to `let client = null` pattern, See #687 
- I also think we should remove `describeAll()` and add support for it on `describe()` function. Do you think this is a good idea ?
- I would love to remove `lodash` from salesforce. But i we don't have enough unit test for me to comfortably do that.  We are using this function `flatten` which i used in commonExecute to `...flatten(operations)`. I have only see this pattern in 3 adaptor. Salesforce, mogli and vtiger. I think we don't need it but i am not confident enough to remove it.

josephjclark:
@mtuchi I agree with you on `describeAll`. Let's kill it. if `describe()` has no arguments, it'll describe all

I'm not that interested in removing lodash tbh. Yes it's a bit annoying having it just for the sake of flatten (which we should easily be able to remove) but lodash is the sort of library we should have available to all adaptors. So I don't mind the dependency.

Maybe once we've added a better suite of tests, we can come back to it. Feel free to raise an issue to remind us

.changeset/gentle-oranges-crash.md
```diff
@@ -0,0 +1,24 @@
...
+'@openfn/language-salesforce': major
+---
+
+New API design for salesforce, including adding composeNextState and removing
+old code.
+
+### Major Changes
+
+- Remove axios dependency
+- Remove old/unused functions. `relationship`, `upsertIf`, `createIf`,
```
josephjclark, line 11: There's no migration guide for upsertIf and createIf, which I think is worth capturing

packages/salesforce/src/Adaptor.js
```diff
@@ -367,50 +64,50 @@ export function bulkQuery(qs, options, callback) {
...
  * @param {string} operation - The bulk operation to be performed.Eg "insert" | "update" | "upsert"
+ * @param {array} records - an array of records, or a function which returns an array.
  * @param {object} options - Options passed to the bulk api.
- * @param {integer} [options.pollTimeout=240000] - Polling timeout in milliseconds.
- * @param {integer} [options.pollInterval=6000] - Polling interval in milliseconds.
  * @param {string} [options.extIdField] - External id field.
  * @param {boolean} [options.failOnError=false] - Fail the operation on error.
- * @param {array} records - an array of records, or a function which returns an array.
+ * @param {integer} [options.pollInterval=6000] - Polling interval in milliseconds.
+ * @param {integer} [options.pollTimeout=240000] - Polling timeout in milliseconds.
```
josephjclark, line 92: `allowNoOp` needs documentation

packages/salesforce/src/Utils.js
```diff
@@ -0,0 +1,144 @@
...
+      );
+    }
+
+    // Wait for the polling interval before the next attempt
+    await new Promise(resolve => setTimeout(resolve, pollInterval));
+    attempt++;
+  }
+}
+
+export const prepareNextState = (state, result, data) => {
```
josephjclark, line 139: What is result? It looks new? How is it different to data?

josephjclark (changes requested):
This has been really hard to review. I've had to locally diff the main Adaptor.js and the new Adaptor.js to try and verify what has changed (and why), because there's too much work here. The diff is too dense and there are too many changes.

The original issue did specifically ask for the work to be broken down, and I really wish you'd done that.

But I think I've figured it out and left some comments.

We will need to look at tests and docs in a separate PR before we can release this. I think we've also got some other issues to add.

So let's get this basic refactor finished as soon as possible and then we can start building on top of it

packages/salesforce/src/Adaptor.js
```diff
@@ -488,9 +182,123 @@ export function bulk(sObject, operation, options, records) {
...
+ * @example <caption> Single record creation</caption>
+ * create("Account", { Name: "My Account #1" });
+ * @example <caption> Multiple records creation</caption>
+ * create("Account",[{ Name: "My Account #1" }, { Name: "My Account #2" }]);
+ * @function
+ * @param {string} sObject - API name of the sObject.
+ * @param {object} records - Field attributes for the new record.
+ * @returns {Operation}
+ */
+export function create(sObject, records) {
```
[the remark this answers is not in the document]
mtuchi: I think `sObjectName` makes more sense. let me update to that
For clarity 
- sObject: A general term for any object in Salesforce.
- sObjectName: The specific API name of the object, e.g., "Account."

packages/salesforce/src/Adaptor.js
```diff
@@ -526,46 +339,66 @@ export function destroy(sObject, attrs, options) {
         );
 
         const failures = result.filter(r => !r.success);
-        console.log('Failed to delete: ', JSON.stringify(failures, null, 2));
+        if (failures.length > 0)
```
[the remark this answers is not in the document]
mtuchi: Something like this 
```
if (failures.length > 0) {
  console.log("Failed to delete: ", JSON.stringify(failures, null, 2));
  if (failOnError) throw "Some deletes failed; exiting with failure code.";
}

```

packages/salesforce/src/Utils.js
mtuchi: Well sometime we have result that comes with other metadata for example 
If you run `describeAll()` this will return 
`{"encoding": "UTF-8",  "maxBatchSize": 200, "sobjects": [{}, {}]}`. So the `sobjects` that can be `state.data` and `state.result` will be everything returned by salesforce `{"encoding": "UTF-8",  "maxBatchSize": 200, "sobjects": [{}, {}]}`

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: I have updated the code to use the pattern above ⬆️

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Exactly - that's my understanding

.changeset/gentle-oranges-crash.md
```diff
@@ -0,0 +1,24 @@
+---
+'@openfn/language-salesforce': major
```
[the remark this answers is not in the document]
mtuchi: Thanks for the feedback, I have break down the changesets into 3 files, major, minor and patch changeset. 

I will apply this feedback in my next PR, i think the big take away from this is we should really breakdown issues into smaller chunks and deliver small set of PRs.

.changeset/gentle-oranges-crash.md
mtuchi: I have updated the changeset to include a migration guide

packages/salesforce/test/Adaptor.test.js
```diff
@@ -42,60 +26,7 @@ describe('Adaptor', () => {
         .then(state => {
           expect(spy.args[0]).to.eql([sObject, fields]);
           expect(spy.called).to.eql(true);
-          expect(state.references[0]).to.eql({ Id: 10 });
```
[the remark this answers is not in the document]
mtuchi: I will investigate this further, it seems like i am getting undefined when i use `state.references[0]`

.changeset/gentle-oranges-crash.md
josephjclark: It looks to be duplicated?

mtuchi: My bad 🙌🏽 , i have cleaned up the duplicates

.changeset/gentle-oranges-crash.md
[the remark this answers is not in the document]
josephjclark: > the big take away from this is we should really breakdown issues into smaller chunks and deliver small set of PRs.

Well, yes, that's almost always true. Remember I _did_ try to break this work down into three PRs, and the salesforce work has already been broken into issues.

There are many truths and many learnings. Best not to get too fixated on one, or to try and create some simple absolute truth.

You just have to be really mindful of everything you're doing. From when you commit, to when you `changeset`, to when you branch, to how you assemble a PR, to when and how you test. These things aren't formalities, they're not arbitrary rules or magic commands. They are tools and processes you use to structure and share your work. You have to _apply_ them to each situation, to think it through, not just copy and paste across branches.

packages/salesforce/src/Adaptor.js
```diff
@@ -526,46 +339,66 @@ export function destroy(sObject, attrs, options) {
...
+      return util.prepareNextState(state, result, sobjects);
     });
   };
 }
+/**
+ * Send a GET HTTP request using connected session information.
+ * @example
+ * get('/actions/custom/flow/POC_OpenFN_Test_Flow');
+ * @param {string} url - Relative to request from
+ * @param {object} options - Request query parameters
```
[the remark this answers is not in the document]
josephjclark: This text is still wrong

packages/salesforce/src/Utils.js
josephjclark: But it's new though? This doesn't exist on main right now, that metadata is discarded.

The problem is that this prepareNextState signature doesn't make a lot of sense.

It would accept this, if you _really_ think it's valuable:
```
export const prepareNextState = (state, data, extras {}) => {
  return {
    ...composeNextState(state, data),
    ...extras,
  };
};
```

.changeset/gentle-oranges-crash.md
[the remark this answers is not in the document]
mtuchi: This is well noted and very appreciated. Will put this to work on the next one

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: I have changed this to ` * @param {object} options - The options for the request.`

packages/salesforce/src/Utils.js
mtuchi: I think we should make this simple and do this 
```
export const prepareNextState = (state, data) => {
  return {
    ...composeNextState(state, result),
    result,
  };
};
```
So the result will be in `state.data` and `state.result`.

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: I think your commit edited the wrong line

The `get` options object allows you to set options and headers. Currently it still reads this:

image

Please don't rush this stuff :pray:

packages/salesforce/src/Utils.js
josephjclark: The code snippet isn't right (result is undefined). I assume you mean this:
```
export const prepareNextState = (state, result) => {
  return {
    ...composeNextState(state, result),
    result,
  };
};
```
In which case, you're just duplicating result and data? To what end?

mtuchi: Now that i am thinking about it, we don't really need to add result. I think we have that pattern in our adaptor which adds `state.result` as well. 

I think the idea was even if that `state.data` gets updated you still have `state.result`

I guess this should be like this 👇🏽  then 
```
export const prepareNextState = (state, result) =>  composeNextState(state, result)
```

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: I think i was looking at `request()` it has options as well. let me update the `get()`

packages/salesforce/src/Utils.js
josephjclark: I think `state.result` is an entirely new thing you've invented here :joy:  it's like `state.response` but I can't see it on main.

I vote to keep it simple and just do what's in your last comment

```
export const prepareNextState = (state, result) =>  composeNextState(state, result)
```

packages/salesforce/test/Adaptor.test.js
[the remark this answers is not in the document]
mtuchi: Hiya @josephjclark since we are using `composeNextState` the data will be `state.data`. If we want to test `state.references[0]` that mean we have to run another operation so that `state.data` gets moved to `state.references`

packages/salesforce/src/Utils.js
mtuchi: I have updated `prepareNextState` function

packages/salesforce/test/Adaptor.test.js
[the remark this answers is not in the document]
josephjclark: Ah gotcha, that makes sense. Thanks for checking this out!

packages/salesforce/src/Utils.js
```diff
@@ -0,0 +1,139 @@
...
+      );
+    }
+
+    // Wait for the polling interval before the next attempt
+    await new Promise(resolve => setTimeout(resolve, pollInterval));
+    attempt++;
+  }
+}
+
+export const prepareNextState = (state, data) => composeNextState(state, data);
```
josephjclark, line 139: Since `prepareNextState` isn't actually doing anything, can we just use `composeNextState` directly? So remove `prepareNextState` entirely

mtuchi: I have removed `prepareNextState` and use `composeNextState` directly

josephjclark (approved):

Issue #681, opened 2024-07-18
Epic: Resolve HTTP URL issues


josephjclark:
This is a cross-repo epic issue to track fixes to #665

The PROBLEM is that the "generic HTTP adaptors" accept absolute URLs and blindly append credentials, so that credentials may end up getting sent to another server.

The SOLUTION IS:
- HTTP Helpers should ONLY accept paths (not full URLs)
- We replace the common http helper:
  - rebase against common HTTP to make them consistnet
  -  Don't do any credential automation
  -  Provide some helper APIs to make basic auth and access tokens easy

## TODO list

here's all the work that needs to be done

* Re-write the `http` namespace in common
* Create (or export) some auth utils in common
* Expose `request` as a HTTP util (we could do this later)
* Expose native fetch in the runtime sandbox
* Make sure we have a good way to document this
  *  What I mean is, docs.openfn.org needs to be able to render the namespace properly
  * Also lightning's adaptor docs
* Identify all the leaky adaptors in this list and update them
  * Ensure that docs are clear about paths
  * Ensure common http is exported

josephjclark:
This is almost finished: I just need to get the two PRs rebased and merged in. Need to run some QA too.

I doubt it'll be done today but probably ready to ship tomorrow

PR #683, opened 2024-07-19
Don't allow common http helpers to call absolute URLs

### LINKED ISSUE #665: HTTP helpers: don't append auth if not using base url
Warning about an new antipattern that may be emerging in the new http adaptors

when you do a `get`, it'll assume you're using the backend and append headers and stuff. But you might be calling another system entirely.

Should the helpers assume a backend? Should there be a different http.get/common.get/util.get?

josephjclark:
This PR prevents http helpers on adaptors from calling absolute URLs if the baseURL is set.

Fixes #665

Part of the #681 epic

## Notes

* The risk of this isn't as bad as I feared: many adaptors (and docs) are strongly geared towards paths anyway
* I've tightened up the common `util.request` so that it throws if the request path has a different origin to the base URL. If no baseURL is passed, the function won't throw. This should automatically "fix" most adaptors.
* I've also added  an `assertRelativeURL` helper for adaptors to use before manipulating the path
* I've updated adaptors and docs
* commcare is already designed to take a path, not a full URL. This is clearly documented. No action needed there.
* The template has been updated to encourage better practice
* Unit tests have been added to ensure we throw for an absolute URL

## Example: asana

The Asana adaptor always sets a base URL. So any request to an absolute URL will try to append credentials.

Asana will now throw in this case.

Job code (no config needed):

```
request('https://jsonplaceholder.typicode.com/todos/1');
````

Output:
```
$ openfn tmp/http.js -ma asana

[CLI] ♦ Versions:
         ▸ node.js                   18.17.1
         ▸ cli                       1.6.1
         ▸ @openfn/language-asana    monorepo
[CLI] ✔ Loading adaptors from monorepo at /d/repo/openfn/adaptors
[CLI] ⚠ Skipping auto-install as monorepo is being used
[CLI] ✔ Compiled all expressions in workflow
Asana says:
[R/T] ✘ Failed step job-1 after 146ms
[R/T] ✘ {
  "source": "runtime",
  "name": "AdaptorError",
  "severity": "fail",
  "message": "Target origin does not match baseUrl origin",
  "details": {
    "code": "BASE_URL_MISMATCH",
    "description": "A request was attempted to an absolute URL, but a different base URL was specified. This is potential security violation.",
    "target": "https://jsonplaceholder.typicode.com/todos/1",
    "baseUrl": "https://app.asana.com/api/1",
    "fix": "Try using a generic HTTP function to access the target URL"
  }
}
[R/T] ✘ Check state.errors.job-1 for details.
[CLI] ✔ State written to tmp/output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 255ms
```

## Further work

The generic HTTP helpers need to be fixed. Note that any adaptors which re-export http require major version bumps in that work. Also in that work, adaptors which do not export http must do so.

We could probably release this standalone, but I'd quite like to do more work in this area in the epic

## QA Notes

I've updated all the adaptors that I think need updating, but maybe I've missed some?

### DIFF
--- a/.changeset/clever-cars-invent.md
+++ b/.changeset/clever-cars-invent.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-http': patch
+---
+
+In all functions, if baseUrl if set, path MUST be relative.
--- a/.changeset/lemon-rivers-speak.md
+++ b/.changeset/lemon-rivers-speak.md
@@ -1,0 +1,7 @@
+---
+'@openfn/language-asana': patch
+---
+
+Don't allow HTTP helpers to call out to different domains. This can cause a
+security violation where credentials are sent to external servers. Use generic
+HTTP helpers like `http.get` or `fetch` instead.
--- a/.changeset/perfect-badgers-appear.md
+++ b/.changeset/perfect-badgers-appear.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-asana': patch
+---
+
+Fix an issue where not passing a params argument would trigger an exception
--- a/.changeset/tough-items-compete.md
+++ b/.changeset/tough-items-compete.md
@@ -1,5 +1,6 @@
 ---
 '@openfn/language-odk': patch
+'@openfn/language-satusehat': patch
 ---
 
 Enforce that absolute urls should not be passed
--- a/.changeset/tough-items-compete.md
+++ b/.changeset/tough-items-compete.md
@@ -1,6 +1,7 @@
 ---
 '@openfn/language-odk': patch
 '@openfn/language-satusehat': patch
+'@openfn/language-surveycto': patch
 ---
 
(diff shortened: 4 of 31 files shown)


josephjclark:
When satisfied that the core technical problem is fixed,we can merge this in

josephjclark:
@mtuchi I would appreciate a review of this PR this week.

It'll be released alongside other common HTTP changes. I just want you to sanity check, maybe run a little bit of QA on a couple of adaptors, and note the structure of the PR and changesets. This shouldn't take you more than an hour so please timebox if you need to.

@hunterachieng Could I ask you to check this PR description and have a look and test at the Satusehat changes specifically? I'd like to verify that I haven't broken anything. It should only take you ~20 minutes.

PR #690, opened 2024-07-23
docs: support namespace exports

josephjclark:
This PR enables namespaces within an adaptor (like `dateFns` and `http`) to be documented properly on the doc site.

image

It also enforces that only functions with the `@public` tag are documented - which means that lots of internal functions have been removed from the listing (hurrah)

This is an essential dependency for re-writing the new common http operations (#681) and enabling the util pattern #522.

Some functions have now been documented for the very first time, and frankly the documentation isn't good. But this will be changing in a couple of days, so don't sweat that.

## Issues

https://github.com/OpenFn/docs/issues/447

## Screenshots

In Asana (for example), the `http` namespace is referenced without ().
image

The link takes you to the right place in common docs:

image

And the sidebar shows namespaces:

image

I've also added some very light documentation to date-fns:

image

## How to test

* Checkout the repo locally on this branch
* Set the env var `OPENFN_ADAPTORS_REPO` to the path to this repo, eg `~/repo/openfn/adaptors`
* Run the following commands to set up:
  * `pnpm install`
  * `pnpm build`
  * `pnpm docs:build`
*  (you may to to `npm install -g pnpm`)
* Check out the docs repo locally
* Run `yarn start:dev`

This should open up a browser window to `localhost:3000`, where the Adaptors page shows the docs from your local repo.


josephjclark:
One big change I'm about to push is: only `@public` functions will be included in docs

This is because some function definitions (like surveyCTO's `util.convertDate`) are accidentally getting picked up by the docs now. In production, these things are accidentally getting ignored because they happen to be declared as constants (not functions).

Generally speaking I am happy with only `@public` members getting documented. I think it's a rule we adopt (enforce?) anyway. But I need to compare branches and see.

josephjclark:
I'm a bit stuck with dateFns.

The rule is kind of that any exported item needs to have jsdocs attached to it - which make sense, because jsdoc wouldnt; generate stuff for jsdocs.

I think what we have to do in this case is either:
a) provide explicit documentation on each dateFns function
b) export an object and document that explicitly as a namespace

For production right now all we really need is a), so I'll just go ahead and do that

josephjclark:
TODO - I think state tables are (wrongly) showing under interfaces. http maybe. Check this!

ritazagoni (approved):
Looks great, much clearer!

Issue #549, opened 2024-05-07
Remove `axios`


josephjclark:
We should migrate all adaptors away from `axios`

Affected packages should either:

1) migrate to the new common http utils (preferable)
2) migrate to use undici as a direct replacement

This affects literally every adaptor because while `common` still supports the legacy http APIs, it depends  on axios (and so everyone does). 

These adaptors have direct dependencies on axios:
```
@openfn/language-common@1.13.1
@openfn/language-dhis2@4.0.3
@openfn/language-godata@3.4.0
@openfn/language-mailchimp@0.7.3
@openfn/language-openfn@1.3.10
@openfn/language-progres@1.3.10
@openfn/language-rapidpro@1.0.10
@openfn/language-salesforce@4.6.2
```

josephjclark:
Axios has been patched and improved.

For something like dhis2, maybe we should update axios rather than removing it entirely? Would that be less work?

PR #692, opened 2024-07-24
common: update http helpers

### LINKED ISSUE #521: common: expose addBasicAuth
Many adaptors are copying this function from the new common http helpers:

```
import {
  makeBasicAuthHeader,
} from '@openfn/language-common/util'

const addBasicAuth = (configuration = {}, headers) => {
  const { username, password } = configuration;
  if (username && password) {
    Object.assign(headers, makeBasicAuthHeader(username, password));
  }
};
```

Instead of exporting `makeBasicAuthHeader`, we should export `addBasicAuth`. Maybe we add options for the user/pass keys on config, just  in case.

josephjclark:
## Summary

This PR replaces common.http (the axios based operations) with new unidici based operations (based on the new request helper). It removes axios too.

Using common, you can now do:
```
http.get(www)
```
This uses the same underlying undici API as the HTTP adaptor.

Other adaptors can re-export this to expose generic HTTP operations.

## Versions

Versions are bumped in this PR which creates a lot of noise :(

Common is up to 2.0, which will trigger an update to many (but not all) adaptors.

Some adaptors have been version locked to common 1.15. This might produce some inconsistencies and we'll ahve to update those in future. This includes salesforce, which will be updated soon anyway, and kobotoolbox, for which there is an issue.

Adaptors which currently export `http` have had a major version bump (except salesfroce)

## docs

image

## Details

The common http helpers are just lightweight wrappers around the request helper.

This means that common is almost the same as HTTP, with the big difference being that HTTP will use your credential and accept relative URLs, while common will not. All calls to the common helper are absolute URLs.

The cool (?) thing here though is that because the common helper won't help you with auth, I've added some APIs to make auth easier.

Check this out:
```
http.get(
  'https://jsonplaceholder.typicode.com/todos',
  http.options().basic('user', 'pass')
)
```

http.options() will return a chainable object with helper functions to set the header to json or add a basic auth header.

I haven't added many helpers yet, we can add more in time.

Everything is beautifully documented as well:

## Issues

Part of #549
Maybe can close #521 through `options()`

## Review Checklist


josephjclark:
TODO: I want to run some real job tests, ensuring that the namespace is really working and that the auth helpers really work. I'll document the test jobs here

josephjclark:
Oh no, I've picked up a really big problem here: `http.get` doesn't actually compile properly!

Didn't I fix this already?

josephjclark:
Glorious: this now works:
```
fn(state => {
  state.config = {
    username: 'rita@...',
    applicationName: 'test-20190880',
    appId: '68df1b1d2a1f4aafbeee4a2e27ca1089',
    password: '...'
  };

  state.url =
    'https://www.commcarehq.org/a/test-20190880/api/v0.5/case/2c5eaa82-f818-4960-886e-a81e5bf3e326';

  return state;
});

http.get($.url, state =>
  http.options().basic(state.config.username, state.config.password)
);
```

josephjclark:
Ack! At least one adaptor depends on the old common.http (kobotoolbox).

Let me check what else this breaks...

josephjclark:
I have to bump versions on this branch to lock adaptors and get tests passing. It's a bit wierd but it'll have to do.

josephjclark:
I am going to version lock salesforce to common@1.15. 

There is a seperate branch to bump the salesforce major version, which can include the new http APIs

Issue #634, opened 2024-06-20
CHT adaptor for pulling (community health toolkit)


taylordowns2000:
Right now we integrate with the community health toolkit by having CHT push to OpenFn using webhooks. We should add an adaptor that pulls as well.

To start, this adaptor should: 
1. handle authentication 
2. include a generic GET helper so we can get any resource
3. include a generic POST so we can create any resource
4. include a generic PUT so we can update any resource

_Later_ we can consider adding more abstraction and specific helper functions. For now, the generic requests would be a great start. (FYI I think Hunter recently worked on a template for these generic http requests when she was working on Satusehat adaptor with Joe - maybe sync with her if you're not sure what next steps are.)

## Demo account 
See LP `CHT Demo Project` for login information. You'll find that there is already some dummy data in there.  

## Resources
https://docs.communityhealthtoolkit.org/apps/reference/api/

![rectangle](https://github.com/OpenFn/adaptors/assets/8732845/f7f592c5-985b-4458-9e6b-c2b5fb9ac03e)
![square](https://github.com/OpenFn/adaptors/assets/8732845/d5f1a887-51f2-4c5e-bf90-5bc5b488dbad)

aleksa-krolls:
Noted @taylordowns2000 - will slot in the 2/3rd week along with OpenLMIS

mtuchi:
I won't be able to work on this issue today, let's promote this to next week sprint

aleksa-krolls:
hey @mtuchi - we're going to give this issue to Hunter because of some shifts in HIH timelines

@hunterachieng pls watch [this vide](https://www.youtube.com/watch?v=Gxf0xNvZPG8&ab_channel=TechChange) and check out the [CHT website](https://communityhealthtoolkit.org/) for background. Also please book time on my calendar for Monday afternoon to get more context on this new app cc @AishaHassen

mtuchi:
Okay sounds good 👍🏽

hunterachieng:
@aleksa-krolls  I have boked a meeting

aleksa-krolls:
If we want to test this later...
1. CHT/Medic team _might_ grant us access to a sandbox cloud environment
2. _Or_ we could consider setting up our own CHT instance locally: https://docs.communityhealthtoolkit.org/hosting/4.x/app-developer/

PR #693, opened 2024-07-25
feat: implement cht adaptor

hunterachieng:
## Summary

Create a `Community Health Toolkit` adaptor

## Details

Implement the `cht` adaptor with basic `get`, `put`, and `post` methods

## Issues

#634 

## Review Checklist


packages/cht/README.md
```diff
@@ -0,0 +1,31 @@
...
+### Configuration
+
+View the
+[configuration-schema](https://docs.openfn.org/adaptors/packages/cht-configuration-schema/)
+for required and optional `configuration` properties.
+
+### A example expression using the `create` function
+
+```js
+post('patient', { name: 'Bukayo' });
```
josephjclark, line 19: Can we make this a more realisic/relevant example? The title will need updating too

packages/cht/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@openfn/language-cht",
+  "version": "1.0.0",
+  "description": "OpenFn cht adaptor",
```
josephjclark, line 4: `cht` should be uppercase `CHT`

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+ * @property {object} form - Pass a JSON object to be serialised into a multipart HTML form (as FormData) in the body.
+ * @property {object} query - An object of query parameters to be encoded into the URL.
+ * @property {object} headers - An object of headers to append to the request.
+ * @property {string} parseAs - Parse the response body as json, text or stream. By default will use the response headers.
+ * @property {number} timeout - Request timeout in ms. Default: 300 seconds.
+ */
+
+/**
+ * Make a GET request against the base URL.
+ * @example <caption>Get a list of contacts</caption>
```
josephjclark, line 27: This is excellent  documentation. Two good, concise, relevant and useful examples. Well done!

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+    "search": "jim"
+  }
+});
+ * @function
+ * @public
+ * @param {string} path - Path to resource
+ * @param {RequestOptions} options - Optional to configure the HTTP request
+ * @param {function} [callback] - Optional callback to handle the response
+ * @returns {Operation}
+ * @state {SHTHttpState}
```
josephjclark, line 41: Nice to see state tables too!

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+* @example <caption>Filter contacts given a name</caption>
+ * get("/api/v2/export/contacts", {
+  "filters": {
+    "search": "jim"
+  }
+});
+ * @function
+ * @public
+ * @param {string} path - Path to resource
+ * @param {RequestOptions} options - Optional to configure the HTTP request
```
josephjclark, line 38: Typo: should read `Options to configure the HTTP request`

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+ * @state {SHTHttpState}
+ */
+export function get(path, options, callback) {
+  return request('GET', path, null, options, callback);
+}
+
+/**
+ * Make a POST request against the base url
+ * @example <caption>Create a new person</caption>
+ * post("/api/v1/people", {
```
josephjclark, line 50: I really like that you've taken these examples from the official docs. That's really smart!

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * Make a PUT request against the base url
+ * @example <caption>Update settings</caption>
+ * put("/api/v1/settings",{overwrite:true});
+ * @function
+ * @public
+ * @param {string} path - Path to resource
+ * @param {RequestOptions} options - Optional to configure the HTTP request
```
josephjclark, line 75: Another typo on `optional` here

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+ * @state {SHTHttpState}
+ */
+export function put(path, options, callback) {
+  return request('PUT', path, null, options, callback);
+}
+
+/**
+ * Make a general HTTP request
+ * @example
+ * request("POST", "patient", { "name":"Bukayo" });
```
josephjclark, line 87: This example needs updating to match CHT

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {function} [callback] - Optional callback to handle the response
+ * @returns {Operation}
+ * @state {SHTHttpState}
+ */
+export function put(path, options, callback) {
+  return request('PUT', path, null, options, callback);
+}
+
+/**
+ * Make a general HTTP request
```
josephjclark, line 85: We should probably say `make a general HTTP request to CHT`

packages/cht/src/Adaptor.js
```diff
@@ -0,0 +1,129 @@
...
+ * request("POST", "patient", { "name":"Bukayo" });
+ * @function
+ * @public
+ * @param {string} method - HTTP method to use
+ * @param {string} path - Path to resource
+ * @param {object} data - Object which will be attached to the POST body
+ * @param {RequestOptions} options - Optional request options
+ * @param {function} [callback] - Optional callback to handle the response
+ * @returns {Operation}
+ * @state {HttpState}
```
josephjclark, line 96: And this should be CHTState, not HTTP state

packages/cht/src/Utils.js
```diff
@@ -0,0 +1,45 @@
+/**
```
josephjclark, line 1: Comment needs removing (maybe we should remove it from template too?)

packages/cht/test/Adaptor.test.js
```diff
@@ -0,0 +1,198 @@
...
+    const error = await request(
+      'GET',
+      '/api/v1/contacts'
+    )(state).catch(error => {
+      return error;
+    });
+    expect(error.statusMessage).to.eql('Forbidden');
+  });
+
+  it('should include basic auth', async () => {
```
josephjclark, line 63: it doesn't really matter, but I would probably remove this test. It's just the same as the first get test.

Thinking about it, it would be more interesting to make sure that there's one test for each request argument  the methods, data, path and options. And we should test that each of those arguments is expanded (ie, can take a function).

But I don't think it's worth spending a lot of time on this. We're built on well-tested code here.

I'll leave it up to you. If you have an hour and would like to spend it writing extra tests, then go for it! If you're pushed for time, I'm happy to just move on.

packages/cht/test/Adaptor.test.js
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    const finalState = await request('GET', '/api/v1/settings')(state);
+
+    expect(finalState.configuration).to.eql(state.configuration);
+  });
+});
+
+describe('get()', () => {
+  it('should make a simple get request', async () => {
```
josephjclark, line 90: I would actually like to see at least one test which includes query parameters.

You can either extend this test or add a second one like `should make a get with parameters`

packages/cht/test/Adaptor.test.js
```diff
@@ -0,0 +1,198 @@
...
+  it('should make a simple post request', async () => {
+    testServer
+      .intercept({
+        path: '/api/v1/people',
+        method: 'POST',
+        headers: {
+          Authorization: 'Basic aGVsbG86dGhlcmU=',
+        },
+      })
+      .reply(201, {
```
josephjclark, line 130: Ah, but this isn't the example response listed in the docs

What you want is to simulate this:

```
HTTP/1.1 200 OK
Content-Type: application/json

{
  "id": "71df9d25ed6732ea3b4435862510d115",
  "rev": "1-a4060843d78f46a60a6f41051e40e3b5"
}
```

packages/cht/test/Adaptor.test.js
```diff
@@ -0,0 +1,198 @@
...
+      },
+      data: {
+        name: 'Hannah',
+        phone: '+254712345678',
+        type: 'contact',
+        contact_type: 'patient',
+      },
+    };
+
+    const finalState = await post('/api/v1/people')(state);
```
josephjclark, line 151: Your post here doesn't include any data!

Which means this test is passing by accident. It's not actually exercising the API properly. In fact, if you comment out the `post`, the test will still pass.

No matter - when you update the response data, the test will fail, and then you can fix it

josephjclark (changes requested):
This is excellent @hunterachieng 

Mostly this is just typos and tweaks, although there's a couple of things in the tests that do need looking at.

If you have time today, I've suggested a few more tests. But they are optional.

packages/cht/src/Utils.js
hunterachieng: I have removed in both

packages/cht/test/Adaptor.test.js
```diff
@@ -0,0 +1,206 @@
...
+      },
+      data,
+    };
+
+    const finalState = await get('/api/v1/settings')(state);
+
+    expect(finalState.data).to.eql(data);
+    expect(finalState.response.method).to.eql('GET');
+  });
+  it('should make a get with parameters', async () => {
```
josephjclark, line 94: This test doesn't seem right...

1) there are no query parameters
2) it's returning XML for some reason

I'm quite happy to include an XML test, but put it under `request` tests and call it, like, `should return XML responses as a string"

The get test here needs to be like `get('/whatever', { query: { /*something*/ }  })`

josephjclark (approved):

PR #694, opened 2024-07-25
epic: common http & safe http

### LINKED ISSUE #665: HTTP helpers: don't append auth if not using base url

josephjclark:
This PR does two things:

* Re-write the common HTTP helper functions (the `http` namespace) #692
* Ensure that "generic http functions" in adaptors cannot call absolute URLs and leak credentials #683

Within that, here's a finer list of the changes

* [x] common: remove axios
* [x] common: expose http namespace, with new, totally generic helpers
* [x] various: document and enforce that http helpers are targeted at the datasource
* [x] various: major version bumps for any adaptor which exports http

I have not created any NEW http exports. Ie, if an adaptor does not expose `http` today, then it will not after this release.

## Versions

Versioning is a bit complicated. Here's a summary:

* Common has gone up to 2.0, which has triggered a bunch of patches
* Anything which exports `http` has had a major version bump
* Apart from salesforce, which will be getting a major bump soon
* A select few adaptors have been version locked to common@1.15 to avoid breakages
* Some adaptors explicitly do not have updates here (salesforce, twilio) because I would rather version lock in the repo and not publish an empty patch. The production releases of those adaptors will already be locked to safe common versions and don't need a patch.

## QA

I am not sure how to QA this! Some ideas

- we should test the common http functions to prove they work
- we should take one of the major changing adaptors - like primero - and check that the `http` namespace works
- We should check that the docs work well (on common and on someone who re-exports `http`)
- I will go through the list of adaptors before release and make sure the versions look correct

## Issues

Closes #665 #681


josephjclark:
Possible new issue for this epic: make sure that adaptors which use common, if possible, are on `workspace:*`. Because a LOT of adaptors are actually pinned to old versions and won't get the benefit

packages/rapidpro/src/Adaptor.js
```diff
@@ -2,7 +2,7 @@ import {
   execute as commonExecute,
   composeNextState,
   expandReferences,
-  http,
+  http, // Important: this is the OLD axios-based http.
```
josephjclark, line 5: Rapdpro is version locked to an old version of common and should not get a new release here

josephjclark:
Hi @mtuchi @hunterachieng 

I want to get this PR merged ASAP but it's a big one! Can I ask you to both review and QA parts of it?  It is important to do this quickly because there is a maintenance cost in keeping the PR open.

Note that you'll need the very latest CLI version to use `http.get` (it fixes a bug in the compiler)

@mtuch please test and review common, then take a look at everything from `asana` to `mailgun`

@hunterachieng please test and review `http`, then take a look at everything from `maximo` to `zoho`

Both of you should be checking:
- Does the new version number look correct? Packages that export common.http need a MAJOR bump. Everything else needs a patch bump. 
- Does the changelog say something sensible?
- Are there any other diffs in the package, and why?

Please check carefully becuase the changes are subtle! Ask questions if you're not sure

This should take you at least 1 hour. If it takes 2, you should stop (and let me know how far you got)

Thank you :pray:

packages/common/src/http.js
```diff
@@ -0,0 +1,152 @@
...
+ * http.get(
+ *  'https://jsonplaceholder.typicode.com/todos',
+ *  http.options().oauth($.configuration.access_token)
+ * )
+ * @param {string} url - URL to access
+ * @param {CommonRequestOptions} options - Request options
+ * @state {CommonHttpState}
+ * @returns {Operation}
+ */
+export function get(url, options) {
```
mtuchi, line 131: I have tested this in `asana` it works but it overwrites `state`

packages/common/src/util/http.js
```diff
@@ -77,31 +78,57 @@ const assertOK = async (response, errorMap, fullUrl, method, startTime) => {
   }
 };
 
+export const ERROR_ABSOLUTE_URL = 'Absolute URLs not suppored';
+
+// throws if a path is absolute
+export const assertRelativeUrl = path => {
+  if (/https?:\/\//.test(path)) {
```
mtuchi, line 85: what if  it's `http`?, I need to test this

mtuchi:
Hiya @josephjclark  i think i need more time to test this PR to give you good feedback. 
So far what i have tested the http common helper `http.get()` which works fine at top level but they over write state, so if i need to persist my state i will have to use this helper at the top of my job code.  I was hoping to use the helpers like this 👇🏽 
```
fn(state => {
  const result = http.get('https://jsonplaceholder.typicode.com/todos/1');

  console.log({ result }, 'Do something with the results');

  return state;
});
```
But the operation above doesn't work it returns empty object.

packages/common/src/util/http.js
josephjclark: Hey great question. The regex means the `https` is optional, but let me push a unit test to convince you

josephjclark: Ok check 975cd9b4a1fc765b8c0bda80367e8b9aba3d336c for tests

packages/common/src/http.js
josephjclark: Oh crap, you're right! I've done it wrong!

josephjclark: That should be fixed now, thank you :pray:

josephjclark:
@mtuchi That's my bad. It should be fixed now and work as per the documentation. Please retest in the morning!

hunterachieng:
@josephjclark  Tested `http` adaptor and it works with the JSON Placeholder api.

packages/beyonic/package.json
```diff
@@ -31,7 +31,7 @@
     "configuration-schema.json"
   ],
   "dependencies": {
-    "@openfn/language-common": "^1.15.1",
+    "@openfn/language-common": "^2.0.0",
```
mtuchi, line 34: Why not `"@openfn/language-common": "workspace:*",`?

packages/common/src/http.js
mtuchi: This is working as expected now 👍🏽

packages/common/test/util/http.test.js
```diff
@@ -81,6 +93,22 @@ describe('parseUrl', () => {
       expect(e.message).to.eql('Invalid URL');
```
mtuchi, line 93: Looks like we are checking if the url is invalid and not the checking if query parameters are extracted correctly 
Perhaps we should update the test name ?

packages/common/test/util/parse-date.test.js
```diff
@@ -7,102 +7,101 @@ import { startOfDay, subDays, subHours } from 'date-fns';
...
 
   it('should return 2 days ago', () => {
-    const target = startOfDay(subDays(new Date(), 2)).toLocaleString()
-    console.log(target)
-    const result = parseDate('2 days ago').toLocaleString()
-    console.log(result)
-    expect(result).eql(target)
-  })
+    const target = startOfDay(subDays(new Date(), 2)).toLocaleString();
+    console.log(target);
```
mtuchi, line 81: should we leave these `console.log`?

packages/beyonic/package.json
josephjclark: I think this is just what pnpm does - but you're absolutely right!

There are a few like this, I'll update them all.

packages/common/test/util/http.test.js
josephjclark: I think you're looking at a collapsed diff here. `should extract query parameters` and `should throw if path and no base` are two different tests

josephjclark:
Thank you for your testing @hunterachieng. Have you finished? Have you looked at the other adaptors on your list?

I'd love to exercise a more complicated API  like Satusehat, just to really be sure

hunterachieng:
@josephjclark  I have gone through the other adaptors and tested `odk` and `satusehat`, both are working well

josephjclark:
Thank you @hunterachieng . Can you approve the PR if you're happy? (slack me if you're not sure how to do this)

@mtuchi when you're ready can you re-check and close out the review?

Thanks both!

mtuchi:
Hiya @josephjclark i am happy with my testing so far, you can proceed. 

As i was testing `http.options().basic()` My thoughts are
It's very useful helper but it's not very clean when it's job code but way better and i might dare say safe than user trying to set Authorization headers in job code.

josephjclark:
What do you mean by it's not "clean" in job code? Can you elaborate on that?

mtuchi:
@josephjclark here is an example.

For this example let's use [Mockaroo API](https://www.mockaroo.com/docs), For Authentication API Key is to be passed with each request, either as a query string parameter called "key" or as an "X-API-Key" request header.

I would write my job code like this 👇🏽 
**Opt A**
```js
http.get('https://api.mockaroo.com/api/users', {
  query: { key: $.configuration.apiKey },
});
```
**Opt B**
```
http.get(
  'https://api.mockaroo.com/api//users',
  http.options({ query: { key: $.configuration.apiKey } })
);
```

I think option A is more cleaner than option B

josephjclark:
Oh, yeah I totally agree. The options helper is only useful if you use the chained helpers (.json() or .basic()). Just using `options()` is not helpful. I'd never recommend option B.

hunterachieng (approved):

Issue #530, opened 2024-04-25
`asana` run logs don't print error message


aleksa-krolls:
## Request
Update the Asana adaptor to print the error message returned by the Asana API. Right now it shows up as `data: { errors: [Array] }`... but I want to see the message that's in that array! 

This is not being done across functions. For example, [see L342 in this run failure](https://v1.openfn.org/projects/consosci/runs/0662743b-6133-7bb6-ac2d-d0db76125123). 
```
AxiosError: Request failed with status code 401
data: { errors: [Array] }
```

## To test
Run [this Asana job](https://github.com/OpenFn/ConSoSci/blob/master/form-sharing/4-upsertAsanaTask.js) with a fake personal access token as the credential to reproduce the `401` error example above.

josephjclark:
Hey @aleksa-krolls is this blocking anything? It's fine on V2 with the latest adaptor. Fixing an old adaptor on v1 is non trivial.

We can do it, but it's gotta be worth the effort

aleksa-krolls:
If the recommendation is just to update the adaptor version, then that's fine. (That wasn't clear before.)

josephjclark:
Platform needs updating to support the latest adaptor version. There are breaking changes so job might need migrating. And I'm not 100% it'll fix the issue on v1 anyway - can't test to see!

aleksa-krolls:
I've suggested a very quick v1 test [on this slack thread](https://openfn.slack.com/archives/C017ELVRSM8/p1722243955790539) to see. But if it's going to be a lot of work, then forget it and we can look to migrate the client's jobs later

josephjclark:
@aleksa-krolls I'm afk right now but I'll check the slack thread.

Thinking about it, I _think_ we could put out a 2.x patch with a quick fix and load that onto platform 

Back at my desk an in hour and I'll look closely at the options

josephjclark:
I've released a patch build 2.1.7. Once that's been released on platform I hope it'll close the issue

PR #698, opened 2024-07-29
fix: display error message

hunterachieng:
## Summary

Display the error messages in `asana`

## Details

The body data when an error is received is not visible in the run logs in `asana`.

## Issues

#530 

## Review Checklist

### DIFF
--- a/.changeset/clean-kids-double.md
+++ b/.changeset/clean-kids-double.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-asana': minor
+---
+
+Display the error body in the console
--- a/packages/asana/src/Utils.js
+++ b/packages/asana/src/Utils.js
@@ -37,7 +37,11 @@ export function request(state, path, params, callback = s => s) {
     .then(callback)
     .catch(err => {
       console.log('Asana says:');
-      logResponse(err);
+      const body = err.body;
+      logResponse({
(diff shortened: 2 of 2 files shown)


josephjclark:
That's strange, because the body is still sent to log response? And I'd prefer the body wasn't stringified - you should get a better log output if you just send the object.

Can you share a repro case? Will I get the error if I try to use the API without credentials?

hunterachieng:
@josephjclark  Here is the [link](https://app.zenhub.com/workspaces/implementation-engineering-600030af88f2bb00155d46a5/issues/gh/openfn/adaptors/530) to the issue. 
The repro case is [here](https://v1.openfn.org/projects/consosci/runs/0662743b-6133-7bb6-ac2d-d0db76125123)

josephjclark:
@hunterachieng This isn't going to work I'm afraid. I know it looks like a simple issue on the surface but it's a really really hard one.

The issue in question occurs on the 2.x version of the adaptor. This fix targets the 3.x version where quite a lot has changed.

I'm also not happy about the way we're stringifying the body here to log it, it's really untidy. In the v2 runtime we're going to lose a lot of control in how we deal with it.

I think we are going to have to create a patch against v2 and not merge it to main. I will deal with this.

Issue #644, opened 2024-06-24
salesforce: autofetch is broken


josephjclark:
The autofetch feature on the salesforce query sort of works but also not really.

This PR #478 introduced a few regressions:

* Referneces is nested, so you have to do `[0][0]`
* The adaptor pushes responses into references, not recerds, So the data isn't really even accurate

Basically autofetch needs to push the RECORDS into a single flat results array

Not to mention that the code is a bit quirky, to say the least.

I'd also ask whether callback should be triggered per batch, or at the end result. I think that when you're talking about a batch processing API it make sense to chunkup the callback or else you're likely to hit a problem downstream. Maybe this is an option. This is something that comes up on many APIs so I'd want to think carefully about what we do here.

josephjclark:
And there are no unit tests on any of this functionality

josephjclark:
The other metadata is not useful when using autofetch. We'll use that stuff internally but there's no point feeding it back to the user.

aleksa-krolls:
Agreed with Joe's comment.

PR #691, opened 2024-07-23
Epic: Salesforce next

josephjclark:
Epic PR to track Salesforce changes

* [x] Update API #662
* [x] docs review
* [ ] ~~tests review~~
* [x] anything else?

PR #699, opened 2024-07-29
`salesforce` Fix `autoFetch` in `query` 

mtuchi:
## Summary

This PR contain a fix `autoFetch` and code improvements for salesforce `query()` function.  All result records are now merged into `records` array

Ref #644 

## Review Checklist


josephjclark:
@mtuchi this is targeted at main, not the salesforce epic. Is that right?

josephjclark:
Ok, I think I get it.

In production, salesforce 4.8.1 will push an array of pages of results  into `state.references[]`.  So if there are two pages of data, you'll end up with something like this:
```
state.references: [
    [{
           done: true,
            totalSize: 6,
            records: [a, b, c],
    }, {
           done: true,
            totalSize: 6,
            records: [d, e, f],
    }]
]
```
The issue in #644 ins't necessarily a bug, just an unexpected design choice.

What this PR does is flattens those pages into a single item
```
state.references: [
    {
          done: true,
           totalSize: 6,
           records: [a, b, c, d, e, f],
    }
]
```

This still isn't the design I would choose, but we're fixing that in #691.

So this PR will fix the 4.x adaptor so that the old references design works a bit better, which is good for our legacy users.

But we'll need to port this fix over to the new 5.x branch.

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,86 @@ describe('Adaptor', () => {
...
+          expect(state.references[0]).to.eql({
+            done: true,
+            totalSize: 5713,
+            records: [{ Name: 'OpenFn' }],
+          });
+        })
+        .then(done)
+        .catch(done);
+    });
+    it('should return nextRecordsUrl if autoFetch is false and totalSize > 2000', done => {
```
josephjclark, line 250: I think the test name should be more like `should not fetch another page if autofetch is false`. Because that explains the behaviour, not the implementation details

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,86 @@ describe('Adaptor', () => {
...
+            nextRecordsUrl:
+              '/services/data/v47.0/query/0r8yy3Dlrs3Ol9EACO-2000',
+            records: [{ Name: 'OpenFn' }],
+          });
+        },
+      };
+      let state = { connection: fakeConnection, references: [] };
+      let spy = sinon.spy(fakeConnection, 'query');
+
+      query('select Name from Account', { autoFetch: true })(state)
```
josephjclark, line 265: Oh look, autofetch is `true` here!

Something is clearly off with this test

packages/salesforce/src/Adaptor.js
```diff
@@ -187,36 +187,38 @@ export function query(qs, options, callback = s => s) {
...
-      console.log(
-        'Done ✔ retrieved records',
-        result.map(ref => ref.records).flat().length
-      );
+    if (qResult.totalSize > 0) {
+      console.log('Total records:', qResult.totalSize);
+      await processRecords(qResult);
+      console.log('Done ✔ retrieved records:', fetchedRecords.length);
     } else {
       result.push(qResult);
```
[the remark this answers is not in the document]
mtuchi: If you get empty result salesforce returns 
```json
{
   "totalSize": 0,
   "done": true,
   "records": []
}
```
The result will not be `null`

packages/salesforce/src/Adaptor.js
```diff
@@ -187,36 +187,38 @@ export function query(qs, options, callback = s => s) {
...
-            console.log(`Error ${errorCode}: ${message}`);
-            throw err;
-          }
-        } else {
-          done = true;
+    const processRecords = async qResult => {
+      const { done, totalSize, records, nextRecordsUrl } = qResult;
+
+      fetchedRecords.push(...records);
+      result = [
```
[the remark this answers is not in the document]
mtuchi: the `fetchedRecords` array holds all the records each time we get results from salesforce. 
This is what salesforce returns 

For result <=2000 records
```json
{
  "totalSize": 3,
  "done": true,
  "records": [a,b,c]
}
```

For result >2000 records
```json
{
  "totalSize": 2100,
  "done": false,
  "nextRecordsUrl": "/services/data/v47.0/query/0r8yy3Dlrs3Ol9EACO-2000",
  "records": [a,b,c,...others]
}
```

If autoFetch is enable and records > 2000 we want 

```json
{
  "totalSize": 2100,
  "done": true,
  "records": [a,b,c,...others] // All 2100 records
}
```

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,86 @@ describe('Adaptor', () => {
       expect(result).to.eql('Nhamaonha 6a Classe 2023-10-09');
     });
   });
+
+  describe('query', () => {
```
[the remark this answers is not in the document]
mtuchi: Okay i have added a test that checks if the `fakeConnection.query` was called with the query right string

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,86 @@ describe('Adaptor', () => {
...
+          expect(state.references[0]).to.eql({
+            done: true,
+            totalSize: 1,
+            records: [{ Name: 'OpenFn' }],
+          });
+        })
+        .then(done)
+        .catch(done);
+    });
+    it('should fetch all records if autoFetch is true and totalSize > 2000', done => {
```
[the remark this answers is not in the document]
mtuchi: It's tough one but i think i got most of it, i have updated the test to inspect if `fakeConnection.request` will be called when `autoFetch: true`. Please have a look

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,86 @@ describe('Adaptor', () => {
       expect(result).to.eql('Nhamaonha 6a Classe 2023-10-09');
     });
   });
+
+  describe('query', () => {
+    it('should fetch all records', done => {
```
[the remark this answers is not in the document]
mtuchi: I am struggling a bit with `fetch all records in a single page` because it doesn't say what is the maximum number of records per single page 🤔 . What about this `fetch up to 2000 records` ?

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Ah, got it, it's pushing the "empty" results object.

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: I don't think this answers the question.

Every query should push a single result object into `state.references`. That result object may contain 0 or 5 or 100000 records. But there's only ever _one_ object.

But you're using an array of result objects, and spreading that array into the final next state's references. That's weird and misleading.

I would rather you  did this after the first query to set up your result:
```
const result = {
  done: true,
  totalSize,
  records: []
}
```

And then each `processRecords` callback pushes to that result array
```
result.records.push(...qResult.records)
```
And finally, we add the result to our references:

```
const nextState = {
      ...state,
      references: [result, ...state.references],
 };
```

In this strategy, you would NOT  push an empty qResult

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,124 @@ describe('Adaptor', () => {
...
+        references: [],
+        qs: 'select Name from Account',
+      };
+      let spy = sinon.spy(fakeConnection, 'query');
+
+      query('select Name from Account')(state).then(state => {
+        expect(spy.calledWith('select Name from Account')).to.eql(true);
+      });
+
+      query(state => state.qs)(state).then(state => {
```
josephjclark, line 222: Why is this duplicated?

if you're testing that references can be expanded, I think that's a different test entirely

A unit test should test one thing and one thing only. Every character of code in a test should support that one case (either setting it up or asserting it worked). Every unrelated character of code weakens the test.

packages/salesforce/test/Adaptor.test.js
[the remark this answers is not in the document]
josephjclark: I don't think it matters what the maximum number of records per page is.

What matters that if there is a another page, the function fetches it.

You've called that test `fetch up to 2000 records`, but I don't know why 2000 records is important. Nor does Hunter when this test starts failing on one of her feature branches.

The most important question here is: what are you actually trying to test?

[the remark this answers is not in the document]
josephjclark: The reason 2000 doesn't matter is that this behaviour is controlled by the salesforce server. It sends `done` and `nextRecordsUrl` to tell you there is more data (it also sends the total count, which you could choose to use)

The adaptor doesn't say "if there are more than 2000 records, fetch another page". The adaptor says: if the request is not `done`, fetch another page.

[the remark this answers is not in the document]
josephjclark: Are you trying to test that the function doesn't autofetch if there are no more items? Are you trying to test that the function can download 2000 items exactly per page? Are you trying to test that if there are many pages of data, the adaptor fetches them all?

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,124 @@ describe('Adaptor', () => {
...
+          expect(spy.called).to.eql(true);
+          expect(spyReq.called).to.eql(true);
+          expect(state.references[0]).to.eql({
+            done: true,
+            totalSize: 5713,
+            records: [{ Name: 'Open' }, { Name: 'Fn' }],
+          });
+        })
+        .then(done)
+        .catch(done);
```
josephjclark, line 286: Oh don't call `done()` if there's an exception. If there's an exception we need this test to fail!

The same goes for the other tests in this describe block

josephjclark:
Hey, passing `autofetch` and `autoFetch` is pretty subtle and very easy to get wrong.

Can we maybe:
a) console.log whether we're going to autofetch
b) support both forms of `autofetch`

josephjclark (changes requested):
Getting closer, but still a few more changes. Thanks!

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
mtuchi: I have updated the code to use the pattern you recommended above

packages/salesforce/test/Adaptor.test.js
mtuchi: I have separated the tests

packages/salesforce/test/Adaptor.test.js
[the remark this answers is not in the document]
mtuchi: I want to test the function can download 2000 maximum rows but not necessary assert 2000 records because depending on your query you might get less than 2000 results.

mtuchi:
I have made the suggested improvements, please have a look

packages/salesforce/src/Adaptor.js
```diff
@@ -187,38 +196,37 @@ export function query(qs, options, callback = s => s) {
...
-        'Done ✔ retrieved records',
-        result.map(ref => ref.records).flat().length
-      );
+    if (qResult.totalSize > 0) {
+      console.log('Total records:', qResult.totalSize);
+      await processRecords(qResult);
+      console.log('Done ✔ retrieved records:', result.records.length);
     } else {
-      result.push(qResult);
+      result = qResult;
```
josephjclark, line 229: Why this assignment? Is `qResult` ever any different than the result declaration above?

I know i'm being picky but I don't think it's a meaningful assignment, which means you can make result a `const` and the code just gets a bit clearer

packages/salesforce/src/Adaptor.js
```diff
@@ -164,19 +164,28 @@ export function retrieve(sObject, id, callback) {
...
     const [resolvedQs, resolvedOptions] = newExpandReferences(
       state,
       qs,
       options
     );
-    const { autoFetch } = { ...{ autoFetch: false }, ...resolvedOptions };
+
+    const { autoFetch: autoFetchOption, autofetch: autoFetchOptionAlias } =
+      resolvedOptions;
+    const autoFetch = autoFetchOption || autoFetchOptionAlias || false;
```
josephjclark, line 185: Maybe just `const autoFetch = options.autoFetch || options.autofetch;`

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,138 @@ describe('Adaptor', () => {
...
+            done: false,
+            totalSize: 5713,
+            nextRecordsUrl:
+              '/services/data/v47.0/query/0r8yy3Dlrs3Ol9EACO-2000',
+            records: [{ Name: 'OpenFn' }],
+          });
+        },
+      };
+      let state = { connection: fakeConnection, references: [] };
+      let spy = sinon.spy(fakeConnection, 'query');
```
josephjclark, line 316: Just to really drive this test home, can we add a mock `request` function (which does nothing) and assert that it isn't called?

And in fact, we don't need to assert that `query` is called at all. We know it's called because we're checking the result data (and anyway we tested that earlier)

If I set `autofetch` to true the test throws an exception, which is good because it means  it's working as designed. But a better assertion will give me confidence that the test works without having to run (and modify) the code

packages/salesforce/test/Adaptor.test.js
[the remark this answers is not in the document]
josephjclark: I still don't understand the value of that, but the way  I would usually test something like that is to test the bounds. Test with 2000 items returned by the server and assert on the correct behaviour, then test with 2001 items and assert on that. What happens if the server returns 2001 items anyway? (hint: nothing at all - the code doesn't care about this 2000 magic number). It would also be valuable to test with 0 items returned.

But because this 2000 limit is arbitrary, I don't think you need to go off and do that.

I'm closing this thread and starting a new one to move us forward.

packages/salesforce/test/Adaptor.test.js
```diff
@@ -194,4 +195,138 @@ describe('Adaptor', () => {
...
+        qs: 'select Name from Account',
+      };
+      let spy = sinon.spy(fakeConnection, 'query');
+
+      query(state => state.qs)(state).then(state => {
+        expect(spy.calledWith('select Name from Account')).to.eql(true);
+      });
+    });
+
+    it('should fetch up to 2000 records', done => {
```
josephjclark, line 244: I'm still not happy about this test :thinking: 

Based on your last comment in the other thread, I still don't think you have a clear vision of what you're trying to test here.

I know I'm pushing you hard on this but I really think we need to raise our test standards across our core adaptors in particular. I want 2 or 3 really good, sharp, focused, valuable tests. I want military precision. I don't want 10 or 20 vague, inconsistent and ineffective ones.

Let's forget about that for a moment: this test says `should fetch up to 2000 records`, but the mock returns only 1 record. So it isn't testing 2000 records at all.

I would be happy enough if you renamed the test to something like `should return some records` and we treat it as a general happy-path test of the function.

Alternatively, you could rename the test to `should fetch 1 record` and at least it's accurate.

mtuchi: I am happy with `should fetch 1 record`. It's a bit hard to create fixtures for more than `2000` records. Perhaps this is will require factories library like chancejs or [factory-girl](https://github.com/simonexmachina/factory-girl)

packages/salesforce/src/Adaptor.js
josephjclark: `{  done: true, totalSize: 0, records: [] }` is perfect for a result with no records. We don't care about `nextRecordsUrl`.

I don't see any need for re-assignment here

packages/salesforce/test/Adaptor.test.js
josephjclark: Well it's super easy to create a 2000 record test , just do `new Array(2000).fill({})`. The contents of the array are totally immaterial.

But fetch 1 is absolutely fine

packages/salesforce/src/Adaptor.js
josephjclark: But there are 0 records. We're specifically talking about the case where `qResult.totalSize === 0`.

We are overwriting the object `{ done: true, totalSize: 0, records: [] ` with  `{ done: true, totalSize: 0, records: [], nextRecordsUrl: ''}` . And it doesn't benefit anyone.

mtuchi: Ooh i think i got it 💡, i have pushed an updated that removes the `qResult` reassignment thingy. Please have a look

josephjclark: I don't understand that commit at all :joy_cat: :joy_cat: 

I'll sort this out and get this released

josephjclark (approved):

.changeset/ten-crews-complain.md
```diff
@@ -0,0 +1,8 @@
+---
+'@openfn/language-salesforce': major
```
josephjclark, line 2: I'm gonna call this a patch release.

it's technically a breaking fix.

But we have a major version coming soon which I expect jobs to migrate to. This patch just ensures that the 4.x branch is stable and behaves as originally intended.

Issue #700, opened 2024-07-29
commcare:some params throwing not found error


hunterachieng:
## Issue

The `commcare` adaptor contains the `get` method which takes in a few `RequestOptions`. Some param items are missing making the adaptor relatively unusable in some scenarios.

josephjclark:
I don't believe this is an adaptor bug. Closing.

PR #701, opened 2024-07-29
fix: add request params

hunterachieng:
## Summary

Add missing parameters for `commcare` `RequestQueries`

## Details

Missing param options in the `RequestQueries` for `commcare`

## Issues

#700 

## Review Checklist

### DIFF
--- a/.changeset/serious-hats-give.md
+++ b/.changeset/serious-hats-give.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-commcare': minor
+---
+
+Add more request queries for various `get` requests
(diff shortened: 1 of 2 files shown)


josephjclark:
Thank you for flagging this @hunterachieng but two really important things here:

1) This does not fix the issue you reported on slack
2) Ok, but it's good to have more docs right?? Well no actually - we've got this documentation wrong. I've raised https://github.com/OpenFn/adaptors/issues/702 to deal with that (which I'll pick up because I'm supposed to be documenting commcare anyway)

PR #707, opened 2024-08-06
Fix survey CTO and automate tags

josephjclark:
## Summary

SurveyCTO 2.2.1 has started throwing an error on `fetchSubmissions`:

```
TypeError: Cannot read properties of undefined (reading 'toString')
```

Turns out this is a bug introduced in the recent patch.

## Tag automation

I've snuck in a little tweak here to run `pnpm chaneset tag && git push --tags` after publish.

I have had trouble getting this to work in the past, but let's see if it work this time :crossed_fingers:  We'll just have to see.

## path.join and URLs

I should have clocked that this would be problematic, but we didn't have enough tests in the adaptor (and I didn't add them) so it got by me.

We've been using a new pattern of `path.join(a,c,b)` for safer URL assembly. It works great on paths - but when surveyCTO does it, it builds a whole URL, starting with `https://`. And when nodepath sees `//` it strips out the extra slash, so the URL becomes `https:/test.surveycto.com`, which is not a valid URL.

## Review Checklist

- [x] If this is a new adaptor, Added the adaptor on marketing website ?

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,6 +12,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
+          fetch-tags: true
       - uses: actions/setup-node@v3
         with:
           node-version: '18.17.1'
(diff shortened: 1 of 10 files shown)


.github/workflows/publish.yaml
```diff
@@ -12,6 +12,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
+          fetch-tags: true
```
josephjclark, line 15: I'm not sure but I think this is what I was missing before :thinking:

PR #708, opened 2024-08-06
Surveycto date format

josephjclark:
Fix survey CTO date format to `MMM dd, yyyy hh:mm:ss a`

Not sure if  I want to release this yet, trying to work out the source of truth


josephjclark:
I'm closing this down because, so far as we can tell, the format string we are using now is fine.

PR #715, opened 2024-08-09
`redis` Add `jGet` function

mtuchi:
## Summary

Add a  `jGet(key: string)`. Ref #714 

**Notes**
changeset is in `release/redis` branch

## Review Checklist

### DIFF
--- a/packages/redis/src/Utils.js
+++ b/packages/redis/src/Utils.js
@@ -48,3 +48,11 @@ export const assertSetArgs = (key, value) => {
     `Make sure both key and value are strings for set(): e.g., set('name', 'fela')`
   );
 };
+
+export const assertjGetArgs = key => {
+  assertArgType(
+    key,
+    'string',
+    `Make sure to pass a strings for jGet(): e.g., jGet('patient').`
+  );
+};
--- a/packages/redis/src/Utils.js
+++ b/packages/redis/src/Utils.js
@@ -14,12 +14,12 @@ export const asserthGetArgs = (key, field) => {
   assertArgType(
     key,
     'string',
-    `Ensure both key and field are strings for hget(): e.g., hget('patient', 'name').`
+    `Ensure both key and field are strings for hget(): e.g., hget('patient', 'name')`
   );
   assertArgType(
     field,
     'string',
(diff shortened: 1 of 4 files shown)


packages/redis/src/Utils.js
```diff
@@ -48,3 +48,11 @@ export const assertSetArgs = (key, value) => {
     `Make sure both key and value are strings for set(): e.g., set('name', 'fela')`
   );
 };
+
+export const assertjGetArgs = key => {
+  assertArgType(
+    key,
+    'string',
+    `Make sure to pass a strings for jGet(): e.g., jGet('patient').`
```
josephjclark, line 56: `pass a string` (string singular) would be better english. And don't end error/status/console messages in full stops please.

josephjclark (changes requested):
The get and set could have been in the same PR as they're very closely related functions. Also there's one issue for both functions.

But it's OK.

A changeset is needed here

josephjclark:
Oh, wait, the changeset is on the release branch? Why?

mtuchi:
>Oh, wait, the changeset is on the release branch? Why?

I wanted to have one minor version bump and not multiple changeset which will create multiple minor version bump

packages/redis/src/Utils.js
mtuchi: Fixed the typo

josephjclark:
@mtuchi I don't understand

If you release 3 minor changes at once, changeset will generate one minor version bump. That's the whole point of it, that's why we're using it.

By commiting the changeset to the release branch, you are promising that the release branch will evenually contain that feature. But it doesn't.

What if I decide to release the scan function today and these new functions next week? The release branch is currently in an inconsistent state and we all have to hope that I notice or remember that there are unimplemented changes.

Please keeps things simple. 

Whenever making a change that the user needs to be informed of, generate a changeset on that branch.

josephjclark (approved):

mtuchi:
@josephjclark thanks for the clarity, I think my thinking process hard some flaw 😞, I didn't want to create multiple versions so i thought the changeset should be in the release branch. I think i understand more now how to use the changeset, thanks for the clarification 🙌🏽

PR #729, opened 2024-08-16
Bump axios

### LINKED ISSUE #727: axios: bump version to 0.28
I'd like to bump axios versions to 0.28 across the repo. This will clear out some security notifications (it won't of course affect production releases).

The affected adaptors would be:

```
dhis2
godata
mailchimp
openfn
progres
rapidpro
salesforce
```

Some big tickets there. We can ignore salesforce as axios will be dropped soon anyway.

The 0.x changelog does report some breaking changes: https://github.com/axios/axios/blob/main/CHANGELOG.md

They look subtle. There's some stuff around formdata which makes me nervous.

I think the thing to do is to bump axios, and put a warning in the patch notes that the axios version changed. If any problems are introduced, we'll just have to fix them on demand.

josephjclark:
Bump axios to version 0.28.0.

This bump contains subtle breaking changes. I have no idea if this affects us. I don't think I can realistically test this. I recommend we just release and see what breaks. I think it'll be fine and, if there's trouble, it's easy to roll back to earlier adaptor versions.

Closes #727 

Fixes https://github.com/OpenFn/adaptors/security/dependabot/97

Fixes https://github.com/OpenFn/adaptors/security/dependabot/88

I've branched this off of the imminent release branch, so version bumps are already included. I've updated changelogs though.

### DIFF
--- a/packages/dhis2/CHANGELOG.md
+++ b/packages/dhis2/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Patch Changes
 
+- Update axios version to 0.28.0.
 - 8146c23: Fix typings in package.json
 - Updated dependencies [8146c23]
   - @openfn/language-common@2.0.1
--- a/packages/dhis2/package.json
+++ b/packages/dhis2/package.json
@@ -35,7 +35,7 @@
   ],
   "dependencies": {
     "@openfn/language-common": "workspace:*",
-    "axios": "^0.24.0",
+    "axios": "^0.28.0",
     "lodash": "^4.17.21",
     "qs": "^6.11.0"
   },
--- a/packages/godata/CHANGELOG.md
+++ b/packages/godata/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Patch Changes
 
+- Update axios version to 0.28.0.
 - 8146c23: Fix typings in package.json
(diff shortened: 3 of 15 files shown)


josephjclark:
Hey @stuartc @taylordowns2000 I wanna be brave and just release this - any thoughts?

taylordowns2000 (approved):
you've got my approval, but/and aren't we ditching `axios` entirely (for `undici` or something like it) soon?

edit: wait...  i could have sworn axios was unsupported, but now they've got a "platinum sponsor" and it's being maintained?!

josephjclark:
@taylordowns2000 haha yeah I noticed that - they have a whole 1.0 branch now :shrug: 

We are ditching axios but it'll take a while to migrate the last few adaptors over. Common is done, SF is in the works. I kind want to at least run with a safer, stabler version if we're gonna be stuck with it for a while

josephjclark:
So axios 0.28.0 doesn't support `paramsSerializer` (it should, but it does not), which means dhis2 breaks.

Makes you wonder what else has broken :thinking: 

This is pretty annoying but I am going to abandon ship here. I've lost all confidence in the update and I don't want to spend a lot of time on it. This was a supposed to be a quick fix...

Issue #696, opened 2024-07-26
fhir: Resource builders


josephjclark:
This is a concept to make it easy to generate FHIR resources.

# Builder Functions.

We can introduce "builder" functions which make it easy to create a FHIR resource. Every FHIR resource type should have a builder (of course we can phase them in as we need to - maybe we'd try and start with the level 3 concepts).

Builders are chainable and return a plain, serializable JSON object at the end.

The builders are semantically rich and backed by constants. They should be nicely compatible with the openfn style and have sensible defaults.

For example (super rough sketch):
```
patient($.data.id)
  .name("joe", "clark") // add a name
  .name("joseph", "clark") // add a second name!
  .gender(MALE) // set gender with a constant
```

The builders should be smart. Patients should be presumed alive. Calling `.deceased()` with an optional date would set the right internal fields on the patient object.

Builder functions would be backed by full type definitions, so you get full code assist in Lightning. This lets you build a resource without having the fhir docs open.

Each builder has a typed entrypoint like `patient(id, data)`, where data is just a blob of fhir JSON that will be assignged by default (if you would rather use json than the helper functions, go for it).

# Implementation Notes

The builders would have to use subclasses or mixins or something so that the fhir concepts can be layered

Eg you might have a Name builder and an Address builder, and the Patient builder subclasses or mixes in both of those. So we have a lot of internal compositionality.

## Opportunity #1

If we release the fhir builders as a standalone package, they could be used by any Javascript package using fhir resources. The fhir npm package is pretty generic.

## Opportunity #2 

A lot of this stuff is fairly rote, I think. Could we use AI to generate at least a starting point?

josephjclark:
# WIP Detailed pitch

I have been using AI this week to have a crack at a standalone fhir-builders package (unrelated to openfn) to understand the cost and size and feasibility of it.

My hope is that I can make a core solution - a pattern - and use AI to scale it up.

I'm going to have a crack at documenting the work here in this comment, then I can move it up into the main issue.

## API

I've set up a good basic pattern for the API with strong type support.

## Types

Bit of a sticking point here.

There is a @types/fhir package which we can build on. Hurrah!

Frustratingly though, every property is duplicated with an underscore, so the code assist experience is terrible:

![Screenshot from 2024-08-02 09-33-33](https://github.com/user-attachments/assets/fa540c6c-a982-42b2-8ea7-81c2b05fdaa7)

I am not sure what to do about this. We may need to duplicate the types for a good UX

## Versions

[...]

PR #730, opened 2024-08-16
Add typescript defs to fhir create

josephjclark:
## Summary

An experiment to add typescript support to `create`, giving code assist while using that particular function.

This was conceived as a quick alternative to the FHIR Builders issue. But it won't get us very far because each fhir implementation is different and the generic types aren't actually all that useful.

What we really need to be able to auto-generate a fhir adaptor based on a spec, and for the autogenerated mapper to include typings, cross-version mappings, and builders. All the information is there in the spec so it's possible.

Anyway I'm going to keep this open for a little while in case it provides a useful template for something later.

## TODO

This is left basically in a broken state.

The big problem right now (apart from a hack to shortcut the dts build in fhir) is that the typings depend upon `@types/fhir`. Lightning, in its current form anyway, won't actually load the external typings, so you'll get no benefit. What we need to do is use `dts-bundle` to include `@types/fhir` in-line in our adaptor's typings.

Oh, and related to that, `@types/fhir` does this really annoying thing where every prop is duplicated with an underscore. It's a nightmare for code assist because you can't find what you want.

We can fix that in build too though, by stripping out all the underscore props just before we bundle.

## Issues

Relates to #696

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
   "devDependencies": {
     "@changesets/cli": "2.27.5",
     "@openfn/buildtools": "workspace:^1.0.2",
+    "@openfn/language-common": "workspace:^",
+    "@openfn/language-dhis2": "workspace:^",
+    "@openfn/language-fhir": "workspace:^",
     "@openfn/metadata": "workspace:^1.0.1",
     "@openfn/parse-jsdoc": "workspace:^1.0.0",
     "@openfn/simple-ast": "0.4.1",
--- a/packages/fhir/build-dts.ts
+++ b/packages/fhir/build-dts.ts
@@ -1,0 +1,3 @@
+export default async () => {
+  console.log(' >> WOOP WOOP <<<');
+};
--- a/packages/fhir/src/index.ts
+++ b/packages/fhir/src/index.ts
@@ -2,14 +2,18 @@ import './types';
 
 // // export * as utils from './Utils';
 
-// import * as r5 from 'fhir/r5';
+import * as r5 from 'fhir/r5';
 
-// export declare function create(
-//   resourceType: 'Patient' | 'Bundle',
-//   resource: r5.Patient | r5.Bundle,
-//   params: any,
-//   callback: any
-// ): any;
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -28,6 +28,9 @@
   "devDependencies": {
     "@changesets/cli": "2.27.5",
     "@openfn/buildtools": "workspace:^1.0.2",
+    "@openfn/language-common": "workspace:^",
```
josephjclark, line 31: This is just hacking around for a test file. All these deps need removing.

packages/fhir/build-dts.ts
```diff
@@ -0,0 +1,3 @@
+export default async () => {
+  console.log(' >> WOOP WOOP <<<');
```
josephjclark, line 2: TODO: this my hook to use dts bundle instead of tsc for this one adaptor

Or I suppose I could consider using dts bundle for all adaptors and creating everything in one file? It'll make generation of type listings downstream a lot easier. I remember there was a discussion on this 2 years ago and non-bundling won for some reason

packages/fhir/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export declare function create<
+  R extends keyof VersionMap[V],
+  V extends keyof VersionMap = 'r5'
+>(
+  resourceType: R,
+  // Exclude the duplicated underscored props from @types/fhir
+  // insane! https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html
+  // But doesn't affected nested types
+  resource: Omit<VersionMap[V][R], 'resourceType' | `_${string}`>,
```
josephjclark, line 31: By the way this is SICK - you can include patterns in property names :metal:

packages/fhir/src/types.ts
```diff
@@ -0,0 +1,8 @@
+import * as r5 from 'fhir/r5';
```
josephjclark, line 1: Oh this whole file is not needed now - remove it

PR #741, opened 2024-09-06
Add cursor, dateFns and pagination

mtuchi:


packages/openmrs/src/Adaptor.js
```diff
@@ -175,19 +175,40 @@ export function post(path, data, callback = s => s) {
...
+      // Make the first request
+      const res = await request(
+        state,
+        'GET',
+        '/ws/rest/v1/patient',
+        {},
+        requestQuery
+      );
+
+      response
```
josephjclark, line 196: Sorry but I'm gonna veto this syntax. Put it in an if/else block please!

packages/openmrs/src/Adaptor.js
```diff
@@ -317,15 +338,37 @@ export function getEncounters(query, callback = s => s) {
...
     console.log('Fetching encounters by query', resolvedQuery);
 
-    const response = await request(
-      state,
-      'GET',
-      `/ws/rest/v1/encounter/`,
-      {},
-      resolvedQuery
-    );
+    let response;
```
josephjclark, line 341: This whole block looks duplicated?

josephjclark (changes requested):
There's one code quality issue which I'm going to insist on changing.

But I'm generally not happy that this adaptor has zero tests, and that this PR introduces a lot of code duplication. It's not great.

If you think you can find the time, I would like to see the following changes:

- pagination logic pushed down to the request helper. It'll paginate any request by default, based on `body.links`, unless offset is passed in.
- A set of unit tests on request and paginated requests. Maybe 4-6 tests in total?

If I was doing this, I'd probably ask for half a day. It is not worth spending more than a day on.

packages/openmrs/src/Adaptor.js
```diff
@@ -317,15 +338,37 @@ export function getEncounters(query, callback = s => s) {
...
+    do {
+      const res = await request(
+        state,
+        'GET',
+        `/ws/rest/v1/encounter/`,
+        {},
+        requestQuery
+      );
+
+      response
```
josephjclark, line 354: Oh, you know what, I'd also like to see a log line for every page of data. Maybe something like:
```
fetching data...
fetched 20 results
```

Which repeats for every page

mtuchi: Thanks @josephjclark i will work on the change request on Monday

mtuchi:
I have added couple of unit test and move the pagination implementation into util @josephjclark

packages/openmrs/src/Utils.js
```diff
@@ -21,18 +21,52 @@ export function request(state, method, path, data, params) {
...
-
     query: params,
-
     parseAs: 'json',
   };
 
   const url = `${instanceUrl}${path}`;
 
-  return commonRequest(method, url, options).then(response => logResponse(response));
+  return autoFetchRequest(method, url, options);
```
josephjclark, line 34: I don't think we need `request` and `autoFetchRequest` - it doesn't seem to be a helpful division of functionality. I'd just move the autofetch stuff inside the main function.

packages/openmrs/src/Utils.js
```diff
@@ -21,18 +21,52 @@ export function request(state, method, path, data, params) {
...
+  let allowPagination = isNaN(reqQuery?.startIndex);
+
+  do {
+    const response = await commonRequest(method, url, {
+      ...options,
+      query: reqQuery ?? {},
+    });
+    logResponse(response);
+
+    allResponses
```
josephjclark, line 49: This still needs to be made into an if-else statement.

A ternary expression should always be on the right-hand-side of some expression. It shouldn't be a statement by itself.

packages/openmrs/test/index.js
```diff
@@ -47,3 +59,239 @@ describe('execute', () => {
     });
   });
 });
+
+describe('request', () => {
+  it('should search for a patient', async () => {
```
josephjclark, line 64: I think this test should be called `should GET with a query`. Because you're not really testing the search API, you're testing that the query to request gets appended to the url

josephjclark: Otherwise this is a good test!

packages/openmrs/test/index.js
```diff
@@ -47,3 +59,239 @@ describe('execute', () => {
...
+        { ...jsonHeaders }
+      );
+
+    const state = { configuration };
+    const { body } = await request(
+      state,
+      'GET',
+      '/ws/rest/v1/patient',
+      {},
+      { q: 'Sarah', limit: 1, startIndex: 1 }
```
josephjclark, line 152: If you remove `startIndex` from this test,, does it still pass?

I'll bet it does - but it shouldn't!

You need to set up the interceptor to return a next page. That way you are properly testing that your adaptor code will ignore the next page IF the startIndex is set

packages/openmrs/test/index.js
```diff
@@ -47,3 +59,239 @@ describe('execute', () => {
...
+    );
+
+    expect(body.results[0].display).to.eql('Sarah 2');
+  });
+});
+
+describe('get', () => {
```
josephjclark, line 159: Oh, excellent! You didn't need to add tests for the all the operations, I didn't ask you to do that.

But it's great that we have them now - it'll help prevent regressions in future. Thank you!

packages/openmrs/test/index.js
```diff
@@ -47,3 +59,239 @@ describe('execute', () => {
...
+    const state = { configuration };
+
+    const { data } = await execute(get('encounter/123'))(state);
+
+    expect(data.uuid).to.eql('123');
+  });
+});
+
+describe('post', () => {
+  before(() => {
```
josephjclark, line 179: I know there's only one test in this `describe`, but by creating this interceptor i the before block, you're sort of declaring that this interceptor will work for all these tests. It implies that the interceptor is shared.

Which is fine, and I agree with that, but you need to add `.persist()` or else the interceptor will be destroyed after the first request. It'll create confusion for the next dev coming in. The same for the other tests.

josephjclark (changes requested):
I much prefer this approach :+1: And thanks for the extra test coverage. Just a few little comments to address.

packages/openmrs/test/index.js
mtuchi: Update the test

packages/openmrs/test/index.js
mtuchi: I have added the interceptor for the next page

packages/openmrs/test/index.js
mtuchi: Yeah just very basic test to set the ground if we need to expand more

packages/openmrs/test/index.js
mtuchi: Moved the testServer into it's own test because in each test the interceptor will be different so it doesn't make sense to add the mock in `before()`

packages/openmrs/src/Utils.js
mtuchi: Removed autoFetchRequest function

packages/openmrs/src/Utils.js
mtuchi: Made some improvements for the requestOptions

packages/openmrs/test/index.js
josephjclark: Ah, ok, but now we still need to assert that only one page was actually fetched. You need to check that `body.results.length === 1`

packages/openmrs/src/Utils.js
josephjclark: Ah maybe I wasn't clear - I'm talking about this:
```
if (allResponses) {
  allResponses.body.results.push(...response.body.results)
} else {
  allResponses = response
}
```

packages/openmrs/test/index.js
mtuchi: Done, i have added the assertion for results to be equal  to 1

packages/openmrs/src/Utils.js
mtuchi: Thanks i have update the code accordingly

Issue #751, opened 2024-09-12
commcare: Create a generic request function


hunterachieng:
### Background, context, and business value
Create a generic `request` function in the commcare adaptor

### The specific request, in as few words as possible
We currently do not have a common `request` function which would allow us to implement more features api calls using `commcare`

aleksa-krolls:
@hunterachieng as discussed, please implement this in the adaptor so we can send generic GET/POST/PUT requests. This will be good to generally implement, as well as unblock us on HIH.

hunterachieng:
We have created 2 functions, `request` for normal json body requests and `bulk` for both `submitxls` and `lookup-table` updates

aleksa-krolls:
hey @hunterachieng can we close this out?

PR #755, opened 2024-09-23
feat:implement dynamic request function in commcare

hunterachieng:
## Summary

Implement a generic `request` function in `commcare`

## Details

We have many distinct APIs in commcare that can be used to make different requests. Since we cannot implement each separately, we have a generic function that will be used to make different requests.

## Issues

#751 

## Review Checklist

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -333,7 +333,6 @@ export function request(method, path, body, options = {}) {
   return async state => {
     const [resolvedMethod, resolvedPath, resolvedBody, resolvedOptions] =
       expandReferences(state, method, path, body, options);
-    try {
       const response = await util.request(state.configuration, resolvedPath, {
         method: resolvedMethod,
         data: resolvedBody,
@@ -342,9 +341,7 @@ export function request(method, path, body, options = {}) {
       });
 
       return util.prepareNextState(state, response);
-    } catch (e) {
-      throw e;
-    }
+
   };
 }
 export {
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -4,7 +4,7 @@ import { Blob } from 'node:buffer';
 import js2xmlparser from 'js2xmlparser';
 import xlsx from 'xlsx';
 
-import { request, prepareNextState } from './Utils';
+import * as util from './Utils';
 /**
  * State object
  * @typedef {Object} CommcareHttpState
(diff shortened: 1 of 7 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -316,6 +316,37 @@ export function fetchReportData(reportId, params, postUrl) {
...
+    try {
+      const response = await util.request(state.configuration, resolvedPath, {
+        method: resolvedMethod,
+        data: resolvedBody,
+        params: resolvedOptions,
+        contentType: 'application/json',
+      });
+
+      return util.prepareNextState(state, response);
+    } catch (e) {
```
josephjclark, line 345: You can remove this try/catch block as we're not doing any processing on the error

packages/commcare/test/index.js
```diff
@@ -653,3 +653,37 @@ describe('createUser', () => {
     expect(response.statusCode).to.equal(201);
   });
 });
+
+
+describe('HTTP wrappers', () => {
```
josephjclark, line 658: hmm, since we already have a describe group for `get`, I think we should just call this block `request`

packages/commcare/test/index.js
```diff
@@ -653,3 +653,37 @@ describe('createUser', () => {
...
+        hostUrl,
+        domain,
+        appId: app,
+        username: 'user',
+        password: 'password',
+      },
+    };
+
+    const { data, response } = await 
+      request('GET', '/a/asri/api/v0.5/case')
```
josephjclark, line 683: It would be great to use a bit more of the API in this test. Do you think you could add a query or something just to make sure the params object gets fed through?

You'll need to update the interceptor to also match on the query pattern

josephjclark (changes requested):
Nice and simple, great! Just a couple of little comments

We also need a changeset please! This is a `minor` release because it adds a new function to the API

packages/commcare/test/index.js
```diff
@@ -653,3 +653,40 @@ describe('createUser', () => {
...
+  it('makes a GET request', async () => {
+    testServer
+      .intercept({
+        path: `/a/asri/api/v0.5/case`,
+        method: 'GET',
+        query: {
+          offset: 1,
+        },
+      })
+      .reply(200, () => {
```
mtuchi, line 668: Can we just do this 
```js
reply(200, () => ({ case_id: "1" }));
```

mtuchi (approved):
This looks good @hunterachieng

mtuchi:
@josephjclark i am merging this PR, incase there is any feedback we can create an issue and open another PR

PR #759, opened 2024-09-26
`mssql` remove process.exit(1)

mtuchi:
## Summary

Removed the catch logic on `execute()` function. The catch was using `process.exit(1)` which cause a crash and stop the whole workflow from running. 

## Issues

- https://github.com/OpenFn/health-partners/issues/24

## Review Checklist

### DIFF
--- a/.changeset/thin-icons-change.md
+++ b/.changeset/thin-icons-change.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-mssql': patch
+---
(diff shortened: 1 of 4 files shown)


josephjclark (approved):
Can you bump versions please? `pnpm changeset version`

mtuchi:
@josephjclark i have bump version and update changelog ?

PR #761, opened 2024-09-26
FHIR NDR Ethiopia

josephjclark:
## Summary

A new, mostly auto-generated adaptor for building FHIR NDR Ethopia resources.

## Details

This adaptor includes:
* A bunch of hand-written utility functions for creating basic fhir types (one day these should be moved down into `fhir`
* Auto-generated builder functions which make it easy to create NDR resources.

The idea is that the builder functions provide code-assist while you're editing, so it's easier to find and set the properties you need. The builder and utility functions also handle a bunch of boiler-plate code for you, like generating  default metadata or setting typed keys (like `valueDateTime` vs `valueReference`  - with the builder you just set `value`!)

Once created, the resources need to be send to the destination system. Right now the adaptor doesn't really help with that - use http or fhir.

See `tests/Resources.test.js` for a bunch of usage examples.

## Docs

At the moment only very basic documentation is generated. It'll appear on the doc site when merged.

We need to work out whether and how to specify all the allowed properties for each resource type. This feels like a more complex, detailed and verbose form of type docs than we've done before (the builder function design actually makes this a bit harder sadly)


josephjclark:
This PR has been spun out of https://github.com/OpenFn/adaptors/pull/734

PR #764, opened 2024-09-30
feat: implement bulk upload for lookup-tables and case-data

hunterachieng:
## Summary

Implemented a `bulk` upload for allowing to upload and update `lookup-tables` and `case-data`.

Fixes # #751 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -395,9 +395,9 @@ export function request(method, path, body, options = {}) {
  *     }],
  *   }
  * )
- * @param {string} type - case-data or lookup-tables
+ * @param {string} type - case-data or lookup-table
  * @param {array} data - Array of objects to upload
- * @param {Object} params - Input parameters, see {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143946459/Bulk+Upload+Case+Data CommCare docs} for case-data and {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143946023/Bulk+upload+Lookup+Tables Commcare Docs} for lookup-tables.
+ * @param {Object} params - Input parameters, see {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143946459/Bulk+Upload+Case+Data CommCare docs} for case-data and {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143946023/Bulk+upload+Lookup+Tables Commcare Docs} for lookup-table.
  * @state data - the response from the CommCare Server
  * @returns {Operation}
  */
--- a/packages/commcare/src/Adaptor.js
+++ b/packages/commcare/src/Adaptor.js
@@ -3,6 +3,9 @@ import { expandReferences } from '@openfn/language-common/util';
 import { Blob } from 'node:buffer';
 import js2xmlparser from 'js2xmlparser';
 import xlsx from 'xlsx';
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
 
 import * as util from './Utils';
 /**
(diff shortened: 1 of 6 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -347,6 +348,119 @@ export function request(method, path, body, options = {}) {
...
+ *    [
+ *      {name: 'Mamadou', phone: '000000'},
+ *    ],
+ *    {
+ *      case_type: 'student',
+ *      search_field: 'external_id',
+ *      create_new_cases: 'on',
+ *    }
+ * )
+ * @example <caption>Upload a single row of data for a look-up table</caption>
```
josephjclark, line 368: Can this example be made shorter? Do we really need quite so many fields in the data?

packages/commcare/src/Adaptor.js
```diff
@@ -347,6 +348,119 @@ export function request(method, path, body, options = {}) {
...
+ *        'field:Satuan': 'TABLET',
+ *        'field:Harga':'1375',
+ *        'field:kfa_codes': '92000372',
+ *        'field:satusehat_id': 'TAB',
+ *        'field:strength': '100',
+ *        'field:strength_unit': 'mg',
+ *     }],
+ *   }
+ * )
+ * @param {string} type - case-data or lookup-tables
```
josephjclark, line 398: `lookup-tables` or `lookup-table` ?

josephjclark (commented):
Couple of quick comments @hunterachieng but I'll let @mtuchi take a closer look at the code

Are we going to deprecate `submitXls` now?

packages/commcare/src/Adaptor.js
hunterachieng: @josephjclark  I have fixed it. Its `lookup-table`

packages/commcare/src/Adaptor.js
mtuchi: Something like this perhaps 
```js
bulk("lookup-table", {
  types: [
    {
      "field 1": "Company",
      "field 2": "Func",
    },
  ],
  obat: [
    {
      "field:Company": "OpenFn",
      "field:Func": "Workflow Automation",
    },
  ],
});
```

packages/commcare/src/Adaptor.js
```diff
@@ -347,6 +348,119 @@ export function request(method, path, body, options = {}) {
     }
   };
 }
+
+/**
+ * Bulk upload data to CommCare. Accepts an array of objects, converts them into
```
mtuchi, line 353: Currently the bulk upload works for `case-data` and `lookup-table` not all commcare bulk upload API, Maybe we should say `Bulk upload case-data or lookup-table to CommCare`

packages/commcare/src/Adaptor.js
```diff
@@ -347,6 +348,119 @@ export function request(method, path, body, options = {}) {
...
+
+      return util.prepareNextState(state, {
+        ...response,
+        body: {
+          message: response.body.message.replace(/\n/g, ''),
+          code: response.body.code,
+        },
+      });
+    } catch (e) {
+      throw e.body ?? e;
```
mtuchi, line 460: Just out of curiosity, nothing to change here. 

Is CommCare going to throw without `body` in some edge cases ? I am trying to understand why we doing `e.body ?? e;`

packages/commcare/test/index.js
```diff
@@ -679,11 +678,131 @@ describe('HTTP wrappers', () => {
...
+    expect(data.code).to.equal(200);
+    expect(data.message).to.equal('success');
+
+    // And the adaptor should have uploaded a reasonable looking formdata object
+    expect(formdata.get('case_type')).to.not.be.undefined;
+    expect(formdata.get('case_type')).to.equal('student');
+    expect(formdata.get('search_field')).to.equal('external_id');
+    expect(formdata.get('create_new_cases')).to.equal('on');
+  });
+  it('should successfully submit a lookup-table file', async () => {
```
mtuchi, line 736: Just being picky, your not really submitting a file, the file gets created by the adaptor behind the scene. So may be `should successfully bulk upload lookup-table data`

packages/commcare/test/index.js
```diff
@@ -679,11 +678,131 @@ describe('HTTP wrappers', () => {
...
 
     expect(data).to.haveOwnProperty('case_id');
     expect(response.statusCode).to.equal(200);
   });
-});
\ No newline at end of file
+});
+
+describe('Bulk', () => {
+  it('should submit JSON as an xls file', async () => {
```
mtuchi, line 692: Not really submitting an xls file, That's what the adaptor is doing being the scene. Maybe something very specific like, `should bulk upload case-data`?

packages/commcare/src/Adaptor.js
hunterachieng: I can use a different example but yes it can get as long as the keys in the lookup-table

packages/commcare/test/index.js
```diff
@@ -679,11 +678,131 @@ describe('HTTP wrappers', () => {
...
+      .reply(200, req => {
+        // save out the form data that was upladed
+        formdata = req.body;
+
+        // simulate a return from commcare
+        return { code: 200, message: 'success' };
+      });
+
+    const state = {
+      configuration: {
```
mtuchi, line 709: Since this is use in couple of places you can tidy this up my putting 
configuration in it's own const at the top
```js
const configuration = {
        hostUrl,
        domain,
        appId: app,
        username: 'user',
        password: 'password',
    };
 ```
 Then you can setup state
 ```js
 const state = { configuration, data: {//some initial data} }
 ```

File: packages/commcare/src/Adaptor.js (hunk 1 above)
Comment (mtuchi) [reply packages/commcare/src/Adaptor.js:368]: Yeah sure, as long as it's lookup-table

File: packages/commcare/src/Adaptor.js [hunk 8]
```diff
@@ -347,6 +348,109 @@ export function request(method, path, body, options = {}) {
...
+
+      return util.prepareNextState(state, {
+        ...response,
+        body: {
+          message: response.body.message.replace(/\n/g, ''),
+          code: response.body.code,
+        },
+      });
+    } catch (e) {
+      throw  e;
```
mtuchi, line 450: Okay i don't know why this changed from `throw e.body ?? e` Because now there is no use for catching and throw the error as is. Maybe we should make this simple and try not to use `try.. catch` here

```js
const response = await util.request(state.configuration, apipath, {
  method: "POST",
  data: form,
});

return util.prepareNextState(state, {
  ...response,
  body: {
    message: response.body.message.replace(/\n/g, ""),
    code: response.body.code,
  },
});
```

packages/commcare/src/Adaptor.js
```diff
@@ -347,6 +348,109 @@ export function request(method, path, body, options = {}) {
...
+
+    try {
+      const response = await util.request(state.configuration, apipath, {
+        method: 'POST',
+        data: form,
+      });
+
+      return util.prepareNextState(state, {
+        ...response,
+        body: {
```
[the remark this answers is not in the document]
hunterachieng: The body only has message and code. The message being send when we use lookup-table looks like this: 
`S\nu\nc\nc\ne\ns\ns\nf\nu\nl\nl\ny\n \nu\np\nl\no\na\nd\ne\nd\n \n1\n \nt\na\nb\nl\ne\ns\n.` so I had to format it.  cc @mtuchi

mtuchi (approved):
This looks good @hunterachieng 👍🏽 , I have added unit test on bulk operation for checking the error if type is not `case-data` or `lookup-table`

packages/commcare/src/Adaptor.js
```diff
@@ -344,6 +344,105 @@ export function request(method, path, body, options = {}) {
...
+    const [json] = expandReferences(state, data);
+    let path, file;
+
+    const workbook = xlsx.utils.book_new();
+
+    if (type.toLowerCase() === 'lookup-table') {
+      path = `/a/${domain}/fixtures/fixapi/`;
+      file = 'file-to-upload';
+      // append types and lookup-table name xlsx
+      Object.keys(json).forEach(sectionName => {
```
josephjclark, line 403: `for (key in json) {` is probably a simply way to to looks like this (you can use `async` in it as well - not useful here but useful in other places)

packages/commcare/src/Adaptor.js
```diff
@@ -344,6 +344,105 @@ export function request(method, path, body, options = {}) {
...
+export function bulk(type, data, params) {
+  return async state => {
+    const { domain } = state.configuration;
+
+    const [json] = expandReferences(state, data);
+    let path, file;
+
+    const workbook = xlsx.utils.book_new();
+
+    if (type.toLowerCase() === 'lookup-table') {
```
josephjclark, line 399: `type.toLowerCase()` is duplicated. I'd probably prefer either a case-insensitive regex (maybe too complicated in this case), or type being saved out into a new variable and lowercased

packages/commcare/src/Adaptor.js
```diff
@@ -344,6 +344,105 @@ export function request(method, path, body, options = {}) {
...
+      // append types and lookup-table name xlsx
+      Object.keys(json).forEach(sectionName => {
+        const sectionData = data[sectionName];
+
+        const newSheet = xlsx.utils.json_to_sheet(sectionData);
+        xlsx.utils.book_append_sheet(workbook, newSheet, sectionName);
+      });
+    } else if (type.toLowerCase() === 'case-data') {
+      path = `/a/${domain}/importer/excel/bulk_upload_api/`;
+      file = 'file';
```
josephjclark, line 411: I think file could just be a const `file` in both cases? The actual file name doesn't seem important. If it is important we  probably need an option to take the filename from the user

josephjclark: Also we actually call the file `data.xlsx` anyway :thinking:

josephjclark (commented):
Hey @hunterachieng Just a couple of little comments for future reference.

Nothing here needs to be actioned, but maybe you could check the file name thing and raise an issue? I would be more comfortable if the file name was consistent across the API call. Even if it doesn't matter at the backend, it means the code looks buggy to anyone reading it

PR #765, opened 2024-10-01
`openmrs` add `fhir.get()` function

mtuchi:
## Summary

Add `fhir.get()` function

## Details

Using [OpenMRS FHIR Module](https://fhir.openmrs.org/index.html) I have added another function (`fhir.get()`) which can be used to fetch data from OpenMRS FHIR endpoints. This implementation does not interfere with the current implementation of OpenMRS adaptor which uses [OpenMRS Rest API](https://rest.openmrs.org/?javascript#openmrs-rest-api). 

Since OpenMRS FHIR Module is the implementation of FHIR, I have tried to compile a list of query params that are most relevant when making a get request with search params and include links to the original documentation.

**Important Consideration**
- I could not find the documentation for paging in OpenMRS FHIR Module. But i have tried two endpoints `Encounter` and `Practitioner` and both work as expected in auto-paging.
>`fhir.get('Encounter', { count: 100, lastUpdated: `ge${$.cursor}`)`
>`fhir.get('Practitioner', { count: 2, lastUpdated: `ge${$.cursor}`)`

## AI Usage

- [ ] Strategy / design
- [x] Other
- [ ] I have not used AI

## Review Checklist


mtuchi:
Probably there is a better way of handling auto pagination when fetching encounters from `/ws/fhir2/R4/Encounter` endpoint but currently i have design this function to validate my workflow design https://github.com/OpenFn/openfn-lime-pilot/pull/42

mtuchi:
#### **Notes for Joe**

**why `fhirGet`**
- Currently the adaptor append `/ws/rest/v1/` in each operation so i can't make a get request to an endpoint like `/ws/fhir2/R4/Encounter` because the request url will be `{instanceUrl}/ws/rest/v1/`ws/fhir2/R4/Encounter` which is invalid url. Because i didn't to do a major change i created a `fhirget()` to validate my requirement.

**why `fhirRequest` ?**
- The response for fetching encounters from `/ws/fhir2/R4/Encounter` is different from `/ws/rest/v1/encounter`. So auto paging doesn't work with the current implementation of `util.request()`. I created a copy of `util.request()` and add auto paging 

**Query param mapping.** Eg 🫱  `fhirGet('/ws/fhir2/R4/Encounter', { _count: 100, _lastUpdated: `ge${$.cursor}`});`
- It's hard finding API documentation for `/ws/fhir2/R4/Encounter`. Currently i am using chatgpt to figure out what query params are supported and also looking at the response to see other possible query params

josephjclark:
Ah this is weird :grimacing: 

First, a quick note, your "notes for joe" should really be part of the main PR body. That's important and explains why you're doing this work. When I bug you about doing your github admin or adding detail to PRs, this comment is exactly the kind of thing I'm looking for.

Do we have time to think about this more? I'd much rather do a major version bump and add a fhir "mode" to the existing get functions (or maybe, for now, just to get)

It's a bummer about the documentation. Does this help? https://openmrs.atlassian.net/wiki/spaces/projects/pages/26935963/FHIR+Swagger+Documentation  You should find JSON there which describes the supported query parameters.

mtuchi:
Well noted @josephjclark on `notes for joe` i will keep em coming 
Let me read the FHIR Swagger Docs, I will share feedback. This is the [OpenMRS  Rest API Docs](https://rest.openmrs.org/) that we used to implement the adaptor

mtuchi:
Hiya @josephjclark unfortunately the documentation was not help full because it only show 3 parameters for `/Encounter` and currently i am using `_count and _lastUpdate` in the query parameter. Eg `fhirGet('/ws/fhir2/R4/Encounter', { _count: 100, _lastUpdated: ge${$.cursor}})`. I tried to see if i can find the documentation from the current OpenMRS instance that i am using but i was getting 404 when trying to access [http://{serverHost}:{serverPort}/openmrs/module/fhir/rest/swagger.json](https://openmrs.atlassian.net/wiki/spaces/projects/pages/26935963/FHIR+Swagger+Documentation#). 
There is a sense of urgency on this function because it's needed in one of the msf workflow

mtuchi:
@josephjclark i have cleaned up the implementation a bit and add unit test. I am happy to start with this implementation and modify later when we have a better design. For the time being this will unblock our implementation in msf - https://github.com/OpenFn/openfn-lime-pilot/issues/34

josephjclark:
I'm really not happy about this 😐️

Since you say it's urgent, and I can't do anything about it until next week, I'll do a quick review and release now.

What we're doing here is adding more tech debt to an incredibly long (and serious) list. It's making tomorrow _harder_, which is really the opposite of why we work. It's also been expensive because there's a lot of documentation, and a test, which is going to end up being wasted effort. Actually I don't think we should be duplicating FHIR documentation like this either.

It's a really unsustainable way of developing and we should not be adding short term functions like this.

josephjclark (approved):

mtuchi:
From our call today with @josephjclark we agreed we are going to mark this function as private, So that it will be hidden from docs since we don't want to introduce a major breaking change. Then we will come up with the better design on the next implementation when this function is needed 

cc @aleksa-krolls

PR #768, opened 2024-10-03
`hive` update `hive-driver` package

mtuchi:
## Summary

Update `hive-driver ` package to `1.0.0`

Fixes

- https://github.com/OpenFn/adaptors/security/dependabot/96

## Details
- Recursively update `ws`, `pnpm update ws -r`
- Then `pnpm i && pnpm build`
- Then `pnpm test`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/kind-students-sleep.md
+++ b/.changeset/kind-students-sleep.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-hive': patch
+---
+
+Update hive-driver dependency
--- a/packages/hive/package.json
+++ b/packages/hive/package.json
@@ -29,7 +29,7 @@
   ],
   "dependencies": {
     "@openfn/language-common": "workspace:*",
-    "hive-driver": "0.2.0"
+    "hive-driver": "1.0.0"
   },
   "devDependencies": {
     "@openfn/simple-ast": "0.4.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8496,7 +8496,7 @@ packages:
       source-map: 0.6.1
       wordwrap: 1.0.0
     optionalDependencies:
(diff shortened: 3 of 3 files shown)


josephjclark:
Hmm. I was a bit surprised to even see a dependency on `ws`, because it's a web socket server library and we don't have any interest in creating a websocket server.

Digging in, it looks like this comes from `hive-driver@0.2.0`. I think you can update hive-driver to `1.0.0` (there are no breaking changes listed in the [changelog](https://github.com/lenchv/hive-driver/blob/9a80e888408cd32e197edfa567d1c4625c86c121/CHANGELOG.md)) and everyone will be happy.

Because we don't run any websocket servers I'm happy to dismiss the alert. It doesn't affect us. But a bump on `hive-driver` would be nice as it should clear out any alerts/vulnerabilities/warnings associated with it.

Also: there is no AI disclosure in this PR - please add one ( you can just add a line saying "I did not use AI in this PR", if you like).

mtuchi:
@josephjclark I have updated the `hive-driver` and the PR description . I don't have unit tests though for `hive` adaptor 😞

josephjclark:
@mtuchi are you able to run a manual test from back when you wrote the adaptor?

I am very confident that the upgrade is safe, but if you have a credential and can re-run a basic job, that would take me from 90% to 99% confidence. Worth it if you can do it in 10 minutes. If not I'll merge.

mtuchi:
@josephjclark unfortunately i don't have any way of testing this. It require an apache hive database which i no longer have in my pc.

josephjclark:
@mtuchi Ok that's cool - thanks for clarifying.

Like I say, I'm very confident in the update (no breaking changes listed). So let's just punch it in

PR #789, opened 2024-10-16
Collections

### LINKED ISSUE #758: Build a collections adaptor
## Overview

Create a Collections adaptor that speaks to the Lightning Collections API.

It should probably be suffixed by the backend data store, like `collections-postgres` or `collections-lightning` or something, so that later it's easier to introduce new collection types, like `collections-redis`.

The collections API is a bit unusual in that it will be loaded as a second adaptor to jobs. It's a second class citizen in the job code. So everything needs to be namespaced - collections and operations and stuff. So `collections.get` rather than `get`. Otherwise there's a risk for clashing with the main adaptor namespace.

## API
```
/**
 * Get one more values from a collection. For large collections, use each.
 * options can be a string key name (including wildcards), or an object
 * { key, created_after, created_before, inserted_before, inserted_after }
 * Throws if created AND inserted are specified
 * If no wild card or query included, the value will be written to state.data (or throw if not found)
 * If a wild card or query is included, an array of values will be written to state.data
 */
collections.get(name, query) // writes to state.data

/**
 * Upserts one or more values, as a { key, value } pair, to the named collection
 * If any errors are returned by the server, this will be thrown
 */
collections.set(name, data, options) // throws if errors

/**
 * Remove one or more values from the collection
 * Options can be a string key or a query object
 */
collections.remove(name, query, options) 

/**
 * Iterate over values in a collection which match the query
 * Query can be a wildcard string or object
 * The callback will be invoked with { key, value } on state.data
 * Or what if we do `(state, key, value)` ? because the first thing you need to do
 * is deconstruct anyway
 */
collections.each(name, query, (state) => {})
```

## Configuration

```
{
  collections_key: /* a JWT */
  collections_endpoint: /* set by the worker  */
}
```

Note that config will be set by the worker automatically. Maybe later Lightning will take more control.

## Later work

Maybe add APIs for administering a collection

josephjclark:
## Summary

Adds  a new Collections adaptor

Fixes #758

## TODO

* [x] Pass query options like `updated_before`
* [x] Add proper documentation
* [x] Add richer logging output
* [ ] Maybe mock date filters for better tests? Doesn't really feel material...

## Details

The adaptor provides a fairly simple API.

This implementation and tests rely heavily on a little mock collections server, which uses undici interceptors.

At the moment, because there's no backend and no worker support, the adaptor really only works through the mock interface.

Here's a test job which works through the CLI:
```
// Set up the test
fn((state = {}) => {
  // set up a mock
  const server = mock.createServer();
  collections.setMockClient(server);

  server.api.createCollection('collection');

  state.configuration = {
    collections_token: 'abc',
  };

  state.data = [
    { key: 'a', value: { id: 'a' } },
    { key: 'b', value: { id: 'b' } },
    { key: 'c', value: { id: 'c' } },
  ];

  return state;
});

collections.set('collection', $.data);

collections.each('collection', '*', (state, key, value) => {
  console.log(key, value);
});

```

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


josephjclark:
Here's a thought: do we need to do something to stop the collections adaptor appearing in the lightning drop-down list? :thinking:

PR #796, opened 2024-10-24
update `each` function examples

mtuchi:
## Summary

I have added couple of examples that shows how to use `each` function  and build the docs. 

## Details

The current example for each function is very confusing especially for beginners, Often times i see mistakes between the use of `$.data` and `$.data[*]`.  And more often how to access data in the operation when using each.
Screenshot 2024-10-24 at 3 43 21 PM

 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


mtuchi:
@josephjclark i am not sure if docs changes need a changeset, should i add one ?

packages/common/src/Adaptor.js
```diff
@@ -267,12 +267,32 @@ export function asData(data, state) {
...
+ *       patient_id: $.data.case_id
+ *     }
  *   )
  * )
+ * @example <caption>Inserting patitent data with custom transformations. (Only in v1)</caption>
+ * each(
+ *   $.data,
+ *   insert("patient", (state) => ({
+ *     patient_id: state.data.case_id,
+ *     patient_name: state.data.properties.case_name.toUpperCase(),
```
josephjclark, line 284: I think there are too many props here which bloat the example. Maybe I'd just set `patient_id` and then spread `state.data`

packages/common/src/Adaptor.js
```diff
@@ -267,12 +267,32 @@ export function asData(data, state) {
  * the state's references.
  * @public
  * @function
- * @example
- * each("$.[*]",
- *   create("SObject",
- *     field("FirstName", sourceValue("$.firstName"))
+ * @example <caption>Inserting patitent data using lazy state. (Only in v2)</caption>
+ * each($.data,
+ *   insert("patient",
```
josephjclark, line 272: As you pointed out on slack, the data thing is weird.

What about we add a comment?
```
each($.data,
   // Inside the callback operation, `$.data` is scoped to the item under iteration
   insert("patient",
     {
       patient_name: $.data.properties.case_name,
       patient_id: $.data.case_id
     }
  )
 )

File: packages/common/src/Adaptor.js [hunk 3]
```diff
@@ -267,12 +267,32 @@ export function asData(data, state) {
...
+ * @example <caption>Inserting patitent data with custom transformations. (Only in v1)</caption>
+ * each(
+ *   $.data,
+ *   insert("patient", (state) => ({
+ *     patient_id: state.data.case_id,
+ *     patient_name: state.data.properties.case_name.toUpperCase(),
+ *     patient_age: calculateAge(state.data.properties.birthdate),
+ *   }))
+ * );
+ * @example <caption>Inserting patitent data with custom transformations. (v1 and v2)</caption>
```
josephjclark, line 288: This one uses JSON path, rather than lazy state, which needs to be pointed out in the example

josephjclark (changes requested):
The `each` explanation really isn't good, which is the bigger crime here to be honest. I'll take a look at that if you can fix the examples per my comments.

Regarding changeset - we technically don't need it. But it will result in the git tag being out of date, and it means that `main` has a diff from `production`.

So I would add a patch-level changeset on this.

packages/common/src/Adaptor.js
```diff
@@ -267,12 +267,32 @@ export function asData(data, state) {
  * the state's references.
  * @public
  * @function
- * @example
- * each("$.[*]",
- *   create("SObject",
- *     field("FirstName", sourceValue("$.firstName"))
+ * @example <caption>Inserting patitent data using lazy state. (Only in v2)</caption>
```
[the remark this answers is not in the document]
mtuchi: $.data is lazy state loading and it doesn't work in v1. So i am bit worried there. I am not sure when officially we are going shutdown v1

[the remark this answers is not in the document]
josephjclark: V1 will keep running for ages, but no-one is writing new jobs on it. Check with Aleksa.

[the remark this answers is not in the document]
josephjclark: Paging @aleksa-krolls 🔔Can we document stuff that won't work on v1?

 I think we should be focusing our docs on V2. I also think talking about v1 and V2 is confusing for users. V1 of what?

packages/common/src/Adaptor.js
mtuchi: I don't understand what this mean, can you elaborate more ?

packages/common/src/Adaptor.js
mtuchi: Updated the caption

packages/common/src/Adaptor.js
mtuchi: I have update the example with a comment

packages/common/src/Adaptor.js
```diff
@@ -267,12 +267,32 @@ export function asData(data, state) {
...
+ * each($.data,
+ *   insert("patient",
+ *     {
+ *       patient_name: $.data.properties.case_name,
+ *       patient_id: $.data.case_id
+ *     }
  *   )
  * )
+ * @example <caption>Inserting patitent data with custom transformations. (Only in v1)</caption>
```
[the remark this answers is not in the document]
mtuchi: I have updated the caption

josephjclark:
Thanks @mtuchi , I'll take this over from here

mtuchi:
Noted 👍🏽 @josephjclark but just a heads up, I didn't create a changeset for this, i am not sure if we need changeset for doc changes ?

josephjclark:
I'm gonna hold off release until I've heard back from @aleksa-krolls about the v1 stuff (I'm 98% confident on that thought)

I've bumped versions but because a few packages are version locked to common  - salesforce being the big one - not all adaptors will see this change

josephjclark (approved):

packages/fhir-ndr-et/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openfn/language-fhir-ndr-et",
-  "version": "0.1.4",
+  "version": "0.1.5",
```
josephjclark, line 3: Wait, does this adaptor even export each? If it does, does it actually document it?

I think I need to double check the patches on this PR because I'm not sure that every adaptor is actually affected

packages/common/src/Adaptor.js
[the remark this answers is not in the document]
aleksa-krolls: @josephjclark I agree that we should focus the docs on v2... so no need to say `Only in v2` as that should be assumed. 

That said, I _do_ think it's okay to document when an old example is from `from/supported in openfn v1` ... because this helps our team and v1 users understand why old job code may look different than what they're seeing on the latest docs. But I'll leave this up to you and @mtuchi to decide.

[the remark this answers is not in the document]
josephjclark: That's a good steer, thanks. I'll double check this and see if I want to flag anything before merging, although I suspect not in this work

mtuchi:
@josephjclark seems there is a merge conflict in `packages/fhir-ndr-et/CHANGELOG.md`

josephjclark:
@mtuchi please don't merge this. I'm looking into a couple of things. I'll handle it!

josephjclark (approved):

PR #806, opened 2024-10-30
satusehat: add adaptor examples and update docs

### LINKED ISSUE #509: Potential bug - Salesforce double nesting response array on query?
It seems like the new salesforce adaptor puts the array response from query inside another array. Maybe this is some sort of incompatibility with `composeNextState`?

Anyway, it feels awkward. To find my records I need to type `state.references[0][0].records` rather than `state.references[0].records`

If it _did_ get broken (not sure if it did) somewhere along the way, maybe this is the moment to finally move to the standard composeNextState where we send results into `data`.

hunterachieng:
## Summary

Implementing more examples and update function docs for `satusehat` adaptor

Fixes [#509 ](https://github.com/OpenFn/docs/issues/509)

## Details

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/satusehat/src/Adaptor.js
```diff
@@ -64,10 +68,18 @@ export function get(path, params = {}, callback = s => s) {
 
 /**
  * Make a POST request to Satusehat
- * @example
+ * @example <caption>Make a POST request to create an Organization</caption>
  * post(
  *   "Organization",
- *  { "resourceType": "Organization", "active": true,
+ *  { "resourceType": "Organization",
```
josephjclark, line 74: The formatting here doesn't look quite right

packages/satusehat/src/Adaptor.js
```diff
@@ -64,10 +68,18 @@ export function get(path, params = {}, callback = s => s) {
...
+ *   ...state.data
+ *  }
+ * );
+ * @example <caption>Make a POST request to create an Encounter</caption>
+ * post(
+ *   "Encounter",
+ *  { "resourceType": "Encounter",
```
josephjclark, line 81: Formatting is off here too

I would actually argue that there's no point in creating an Encounter AND Organisation resources, because they're so similar. I know you've been asked for both but I don't think there's any point - not while we have nice, short examples like this

packages/satusehat/src/Adaptor.js
```diff
@@ -103,7 +115,7 @@ export function post(path, data, params = {}, callback = s => s) {
 
 /**
  * Make a PUT request to Satusehat
- * @example
+ * @example <caption>Make a PUT request to update an Organization</caption>
  * put(
  *   "Organization/123",
  *  { "resourceType": "Organization", "active": false,
```
josephjclark, line 121: Out of curiosity, if you're updating an existing organisation, presumably you don't need to send the resourceType? Would it be better or worse practice to exclude it?

packages/satusehat/src/Adaptor.js
```diff
@@ -142,7 +154,7 @@ export function put(path, data, params = {}, callback = s => s) {
 
 /**
  * Make a PATCH request to Satusehat
- * @example
+ * @example <caption>Make a PATCH request on an Organization</caption>
```
josephjclark, line 157: What does a PATCH request do? The example would be clearer if it read like "Make a PATCH request to update an organisation", or whatever the correct verb is

josephjclark (changes requested):
A couple of formatting things which should be fixed, but then some more open questions

josephjclark:
@hunterachieng  I think this needs more thought.

The ask here was to make it clearer how the GET functions relate to the postman examples - basically, how does `get()` map to [Encounter by subject](https://www.postman.com/satusehat/satusehat-public/request/ef2ilan/encounter-by-subject)

What isn't at all clear right now, for docs and examples, is that the second argument, the `params` object, maps to query parameters.  It may also not be clear that all paths are relative to the baseURL specified in configuration (the credential).

For GET, I would also add an example of [getting a resource by id](https://www.postman.com/satusehat/satusehat-public/request/guz33kn/encounter-by-id) (this is nice because it makes the `path` argument look a bit more like a URL).

Also, is it clear that the payloads are FHIR in the examples? Does Satusehat accpet any non-FHIR payloads? (I don't know 

For the methods which have a `data` argument,  should we describe the data as something like "the JSON FHIR resource to create"? That makes it clear that the `data` argument is your actual resource. The current description is a bit woolly. What I'm not sure about is whether any POST requests include non-fhir data, in which case this is presumptuous

Something else that would be useful is to add `@state` annotations to make it clear how the adaptor uses state.data. Search docs for `@state` for example (maybe look at HTTP or commcare(

packages/satusehat/src/Adaptor.js
hunterachieng: @josephjclark  yes it accepts the resource type in the body for a `PUT` request

packages/satusehat/src/Adaptor.js
```diff
@@ -34,17 +42,22 @@ export function execute(...operations) {
...
  *   identifier:'https://fhir.kemkes.go.id/id/nik|9271060312000001'
  * });
+ * @example <caption>Get a specific Encounter by subject</caption>
+ * get('/Encounter', {
+ * subject:100000030009  // This gets an encounter that maps to a specific patient whose id is:100000030009
+ * });
  * @function
  * @param {string} path - Path to resource
- * @param {object} params - Optional request params such as name.
+ * @param {object} params - Optional request params such as identifier.
```
josephjclark, line 57: The description needs to be like `Object of query parameters to include in the request`. Because it's not clear that this parameter represents query parameters, and it's not clear that `{ identifier: 'joe' }` maps to `/Organization?identifier=joe`.

That mapping between the `params` object and the actual query parameters is one of the missing pieces that we're trying to address in this work generally.

packages/satusehat/src/Adaptor.js
```diff
@@ -85,10 +98,11 @@ export function get(path, params = {}, callback = s => s) {
  * @function
  * @public
  * @param {string} path - Path to resource
- * @param {object} data - Object or JSON which defines data that will be used to create a given instance of resource
+ * @param {object} data - JSON FHIR data to create a resource
```
josephjclark, line 101: Maybe we can say `object` instead of `data`

packages/satusehat/src/Adaptor.js
```diff
@@ -80,13 +80,8 @@ export function get(path, params = {}, callback = s => s) {
...
  * Make a POST request to Satusehat. Use this to send resources directly to Satusehat REST API.
  * You can pass Satusehat body data as a JSON FHIR object.
- * @example <caption>Create a resource encounter</caption>
- * post(
- *   "Encounter",
- *  { "resourceType": "Encounter",
- *   ...state.data,
- *  }
- * );
+ * @example <caption>Create a resource encounter. Equivalent to POST `<baseUrl>/Encounter</caption>
```
josephjclark, line 83: 1. This should say "create an encounter resource"
1. Missing backtick here

packages/satusehat/src/Adaptor.js
```diff
@@ -166,15 +158,13 @@ export function put(path, data, params = {}, callback = s => s) {
...
- *    [{
- * "op": "replace", // Operation - `replace` is the only one used to change a specific property or element
- *  "path": "/language", // Path - The name of property/element of resource to be replaced
- *  "value": "id" // Value- The value to be replaced
- *}]
- *
- * );
+ * patch('Organization/abcde', [
+ * {
+ *  op: 'replace', // Operation - `replace` is the only one used to change a specific property or element
```
josephjclark, line 163: hmm. I'm not crazy about all the  comments here - it makes the example very busy :thinking: 

Every word in an example should be meaningful. Critical, even, to understanding the example. And I'm just not sure these comments are? Although I  can see why you've done it because you're right, it's not an intuitive example

I think if the caption said "Change a property of a resource" (instead of just "update a resource"), the data structure here would make less sense and you can get away with fewer comments.

josephjclark: As a side note - is the `op` in a `patch` ALWAYS `replace`? I just checked the docs and that seems to be true.

In that case we should default it so that users don't need to specify it :) Let's not do that now, I'll raise an issue for it

josephjclark (approved):

Issue #207, opened 2023-02-28
DHIS2 integration tests are broken


josephjclark:
* `cd adaptrs/dhis2`
* `pnpm integration-tests`

Tests fail for a wild variety of reasons.

Part of it is an ESM thing (which I expect to get fixed with the magic functions work), but it looks like even when you get the tests running, they go on to fail.

PR #819, opened 2024-11-11
dhis2: Update tracked entities endpoints to V42

### LINKED ISSUE #754: `DHIS2` new tracker api
**Description**

Since DHIS2 version 2.36, the tracker API has been re-implemented and the old API endpoints will be removed in `v2.42`
[Learn More here](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html)

Tracker API Warning

**Tasks**
- Re-implement our `upsert` operation to use the new tracker api 
- Maybe use common.http ?
- what else needs to change ?

**Demo System**
We should probably try building this on the latest version like `v41.1.0` - see all available demo sites: 
https://play.dhis2.org/

For all play instances, the username is `admin` and pw is `district`

hunterachieng:
## Summary

Update tracked entities endpoints to the new v42.

Fixes #754

## Details

Update `get` , `create`, and `update` endpoints to the new tracker endpoints

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


.changeset/wild-dodos-rush.md
```diff
@@ -0,0 +1,4 @@
+---
+'@openfn/language-dhis2': patch
+---
+Removed support for DHIS2 v42
```
josephjclark, line 4: This changeset needs updating. Is this the old changeset from the last fixes?

Anyway this needs to be a major update

packages/dhis2/src/Adaptor.js
```diff
@@ -146,7 +146,7 @@ axios.interceptors.response.use(
  *   orgUnit: 'DiszpKrYNg8',
  *   status: 'COMPLETED',
  * });
- * @example <caption>a trackedEntityInstance</caption>
+ * @example <caption>a trackedEntityInstance. Create a tracker instance. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Create tracker docs}</caption>
```
josephjclark, line 149: In the example, I would call it a `trackedEntity`.

I'm starting to think we should drop support for the term `trackedEntityInstance`.

I'm not very good at DHIS2 but is  "Create a tracker instance" the right language? I don't see that language in the DHIS2 API docs. I suspect you're not creating a tracker, I think you're creating an entity

packages/dhis2/src/Adaptor.js
```diff
@@ -229,20 +229,45 @@ export function create(resourceType, data, options = {}, callback = false) {
   return state => {
     console.log(`Preparing create operation...`);
 
+    let mappedOptions;
+
+    if (shouldUseNewTracker(resolvedResourceType)) {
```
josephjclark, line 234: We need to do resolvedOptions before we create `mappedOptions`. Otherwise, if `options` is passed as a function, things will blow up.

If you're to clear on how and why we resolve options, let me know and I can walk you through it

packages/dhis2/src/Adaptor.js
```diff
@@ -279,8 +304,8 @@ export function create(resourceType, data, options = {}, callback = false) {
  *   storedBy: 'admin',
  *   dataValues: [],
  * });
- * @example <caption>a trackedEntityInstance</caption>
- * update('trackedEntityInstances', 'IeQfgUtGPq2', {
+ * @example <caption>a trackedEntityInstance. Update a tracker instance. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Update tracker docs}</caption>
```
josephjclark, line 307: This caption is wierd.

We're not going to re-write all the captions in this  work. I suggest we simplify and just leave the example the same (but say `a tracked entity` rather than `a trackedEntityInstance`)

packages/dhis2/src/Adaptor.js
```diff
@@ -397,26 +422,51 @@ export function update(
...
+      };
+    } else {
+      mappedOptions = options;
+    }
+
     const resolvedResourceType = expandReferences(resourceType)(state);
     const resolvedPath = expandReferences(path)(state);
     const resolvedData = expandReferences(data)(state);
-    const resolvedOptions = expandReferences(options)(state);
+    const resolvedOptions = expandReferences(mappedOptions)(state);
```
josephjclark, line 439: Again, the order is wrong on the options resolution here

packages/dhis2/src/Adaptor.js
```diff
@@ -425,10 +475,10 @@ export function update(
 
 /**
  * Get data. Generic helper method for getting data of any kind from DHIS2.
- * - This can be used to get `DataValueSets`,`events`,`trackedEntityInstances`,`etc.`
+ * - This can be used to get `DataValueSets`,`events`,`trackers`,`etc.`
  * @public
  * @function
- * @param {string} resourceType - The type of resource to get(use its `plural` name). E.g. `dataElements`, `trackedEntityInstances`,`organisationUnits`, etc.
+ * @param {string} resourceType - The type of resource to get(use its `plural` name). E.g. `dataElements`, `tracker/trackedEntities`,`organisationUnits`, etc.
```
josephjclark, line 481: I am tempted to rename this to `path` instead of `resourceType`. Maybe that's too bige a change. Maybe I'll canvas for opinions tomorrow

packages/dhis2/src/Adaptor.js
```diff
@@ -442,10 +492,17 @@ export function update(
  * });
  * @example <caption>all programs for an organization unit</caption>
  * get('programs', { orgUnit: 'TSyzvBiovKh', fields: '*' });
- * @example <caption>a single tracked entity instance by a unique external ID</caption>
- * get('trackedEntityInstances', {
- *   ou: 'DiszpKrYNg8',
- *   filter: ['flGbXLXCrEo:Eq:124', 'w75KJ2mc4zz:Eq:John'],
+ * @example <caption>a single tracked entity instance by a unique entity ID. Either `program`, `trackedEntityType` or `trackedEntities` should be specified in the params. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#tracked-entities-get-apitrackertrackedentities TrackedEntities docs}</caption>
```
josephjclark, line 495: I think these examples look a bit too complicated and we should do more explanation in the main docs body. Let me think about that

packages/dhis2/src/Adaptor.js
```diff
@@ -485,10 +542,10 @@ export function get(resourceType, query, options = {}, callback = false) {
  * @returns {Operation}
  * @example <caption>Example `expression.js` of upsert</caption>
  * upsert('trackedEntityInstances', {
- *  ou: 'TSyzvBiovKh',
+ *  orgUnits: 'TSyzvBiovKh',
  *  filter: ['w75KJ2mc4zz:Eq:Qassim'],
  * }, {
- *  orgUnit: 'TSyzvBiovKh',
+ *  orgUnits: 'TSyzvBiovKh',
```
josephjclark, line 548: If I look at the [docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#tracked-entities), they say `orgUnit` (singular). Can we double check? and make sure this is accurate?

packages/dhis2/src/Adaptor.js
```diff
@@ -508,33 +565,52 @@ export function upsert(
...
+      const { configuration } = state;
+      return callNewTracker(
+        'create',
+        configuration,
+        { importStrategy: 'CREATE_AND_UPDATE' },
+        nestArray(resolvedData, resolvedResourceType),
+        params,
+        requestConfig
+      ).then(result => {
+        console.log(`Performed a "composed upsert" on ${resolvedResourceType}`);
```
josephjclark, line 582: Since the `.then()` is the same, let's re-use the promise, like we do with create

packages/dhis2/src/Adaptor.js
```diff
@@ -508,33 +565,52 @@ export function upsert(
   return state => {
     console.log(`Preparing upsert via 'get' then 'create' OR 'update'...`);
```
josephjclark, line 566: This log line doesn't really make sense on the new tracker - let's move it into the `else`

packages/dhis2/src/Adaptor.js
```diff
@@ -781,6 +880,34 @@ export function dv(dataElement, value) {
   return { dataElement, value };
 }
 
+export function shouldUseNewTracker(resourceType) {
+  return (
+    resourceType === 'trackedEntityInstances' ||
```
josephjclark, line 885: This would be more efficient and concise with a regex.

packages/dhis2/src/Adaptor.js
```diff
@@ -781,6 +880,34 @@ export function dv(dataElement, value) {
...
+    resourceType === 'tracker' ||
+    resourceType === 'tracker/trackedEntities' ||
```
josephjclark, line 887: Hmm I think the logic is backwards here.

If the resourceType is `tracker/`, `trackedEntityInstance, `enrollments, `events` or `relationships`, we should use the new tracker. OTHERWISE we use the old API.

A few unit tests on this function would be super helpful so that we can easily see how it works! They're easy to create as well because we don't need to mock anything.

packages/dhis2/src/Adaptor.js
josephjclark: Yeah same as https://github.com/OpenFn/adaptors/pull/819#discussion_r1838628253

Always resolve options first, before doing anything else

packages/dhis2/src/Adaptor.js
hunterachieng: @josephjclark  in this comment by the old api you mean the older versions of dhis2 for `tracker`?

josephjclark: Yep, exactly!

packages/dhis2/src/Adaptor.js
```diff
@@ -781,6 +907,29 @@ export function dv(dataElement, value) {
   return { dataElement, value };
 }
 
+export function shouldUseNewTracker(resourceType) {
+  return /^(trackedEntityInstances|tracker\/(enrollments|relationships|events|trackedEntities)|enrollments|relationships|events|trackedEntities)$/.test(
```
josephjclark, line 911: This regex looks over complicated! 

I think this is all you need:
```
return /(tracker\/|enrollments|relationships|events|trackedEntities)/.test(resourceType)
```

Now anything with `tracker/` in the resource type will use the new tracker, plus the 4 specific resources we're looking for.

.changeset/breezy-tigers-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-dhis2': major
+---
+
+Update trackedInstances to use new v42 implementation
```
josephjclark, line 5: We can be a bit more helpful to users with these release notes.

What changes? What breaks? What do users need to know?

The headline feature is probably: `Add support for new Tracker and remove support for API versions <36` (or whatever)

And by the way, somewhere in this PR, if the API version is too low - anything before the tracker was released - we should put out a warning (sort of the opposite of what we're doing in the other one)

.changeset/wild-dodos-rush.md
josephjclark: If this changelog doesn't apply it needs to be removed, because it'll create an inaccurate release note and confuse users

packages/dhis2/src/Adaptor.js
```diff
@@ -146,7 +146,7 @@ axios.interceptors.response.use(
  *   orgUnit: 'DiszpKrYNg8',
  *   status: 'COMPLETED',
  * });
- * @example <caption>a trackedEntityInstance</caption>
+ * @example <caption>a tracker entity. Create a tracker entity instance. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Create tracker docs}</caption>
```
josephjclark, line 149: Docs are still weird here. The link is fine but this should continue to say `tracker entity`, so that it's in the same style as the other examples here

packages/dhis2/src/Adaptor.js
```diff
@@ -236,13 +236,30 @@ export function create(resourceType, data, options = {}, callback = false) {
...
 
-    return request(configuration, {
-      method: 'post',
-      url: generateUrl(configuration, resolvedOptions, resolvedResourceType),
-      params,
-      data: nestArray(resolvedData, resolvedResourceType),
-      ...requestConfig,
-    }).then(result => {
+    let promise;
+    if (shouldUseNewTracker(resolvedResourceType)) {
```
josephjclark, line 240: hmm, hang on, are we sure `request` should use the new tracker?

Request is a low level API and it should do _exactly_ what you tell it. If you do `request('trackedEntityInstance')` it should return an error.

Oh, also request is a private function, not used directly in job code. So we really shouldn't be adding intelligence to it.

packages/dhis2/src/Adaptor.js
```diff
@@ -279,8 +296,8 @@ export function create(resourceType, data, options = {}, callback = false) {
  *   storedBy: 'admin',
  *   dataValues: [],
  * });
- * @example <caption>a trackedEntityInstance</caption>
- * update('trackedEntityInstances', 'IeQfgUtGPq2', {
+ * @example <caption>a tracker entity. Update a tracker entity instance. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Update tracker docs}</caption>
```
josephjclark, line 299: You should remove `Update a tracker entity instance.` from here, it's kinda wierd next to the other docs

packages/dhis2/src/Adaptor.js
josephjclark: Still thinking about this, but I need to hold the whole dhis2 API in my head and I can't do that at the moment :(

packages/dhis2/src/Adaptor.js
```diff
@@ -472,6 +513,56 @@ export function get(resourceType, query, options = {}, callback = false) {
   };
 }
 
+/**
+ * Post data. Generic helper method for posting data of any kind to DHIS2.
+ * - This can be used to create `DataValueSets`,`events`,`trackers`,`etc.`
```
josephjclark, line 518: Please remove the `-` and indent at the start of this line. Please also remove the backticks around `etc`

packages/dhis2/src/Adaptor.js
```diff
@@ -695,7 +804,7 @@ export function patch(
  * @param {{apiVersion: number,operationName: string,resourceType: string}} [options] - Optional `options` for `del` operation including params e.g. `{preheatCache: true, strategy: 'UPDATE', mergeMode: 'REPLACE'}`. Run `discover` or see {@link https://docs.dhis2.org/2.34/en/dhis2_developer_manual/web-api.html#create-update-parameters DHIS2 documentation}. Defaults to `{operationName: 'delete', apiVersion: state.configuration.apiVersion, responseType: 'json'}`
  * @param {function} [callback] - Optional callback to handle the response
  * @returns {Operation}
- * @example <caption>a tracked entity instance</caption>
+ * @example <caption>a tracked entity instance. Delete a tracker instance. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Delete tracker docs}</caption>
```
josephjclark, line 807: Again we need to remove ` Delete a tracker instance.` from here to keep the docs consistent

packages/dhis2/src/Adaptor.js
```diff
@@ -781,6 +907,29 @@ export function dv(dataElement, value) {
...
+  options,
+  data = null,
+  params,
+  requestConfig
+) {
+  return request(configuration, {
+    method: 'post',
+    url: generateUrl(configuration, options, 'tracker'),
+    params,
+    data: type !== 'delete' ? data : {},
```
josephjclark, line 928: Is this necessary? Can we not just pass data through, even if it's not defined?

packages/dhis2/src/Adaptor.js
```diff
@@ -781,6 +907,29 @@ export function dv(dataElement, value) {
...
+export function shouldUseNewTracker(resourceType) {
+  return /^(trackedEntityInstances|tracker\/(enrollments|relationships|events|trackedEntities)|enrollments|relationships|events|trackedEntities)$/.test(
+    resourceType
+  );
+}
+
+export function callNewTracker(
+  type = 'update',
+  configuration,
+  options,
```
josephjclark, line 919: I think `callnewTracker` should set the `importStrategy` option based on the `type`, unless it's already been set.

So if  `type` is `create`, the importStrategy defaults to `CREATE`, etc.

This makes life easier for functions call call into `callNewTracker`, because they don't have to work out the verb AND strategy.

packages/dhis2/test/index.test.js
```diff
@@ -142,6 +149,13 @@ describe('get', () => {
   });
 });
 
+describe('helperfunctions', () => {
+  it('should use the new tracker for enrollments', () => {
```
josephjclark, line 153: More tests please!

I want a test for each of the 4 resource types, one for `tracker/<whatever>`, and at least 2 tests for values which return false (like `dataValueSet`)

packages/dhis2/test/index.test.js
```diff
@@ -173,6 +189,7 @@ describe('create', () => {
       });
 
     const finalState = await execute(create('events', state.data))(state);
+    console.log({ finalState });
```
josephjclark, line 192: This log needs to be removed

packages/dhis2/test/index.test.js
```diff
@@ -413,26 +394,63 @@ describe('upsert', () => {
...
-                },
-              ],
             }
           )
         )(state),
       'Cannot upsert on Non-unique attribute. The operation found more than one records for your request.'
     );
   });
+
+  it('should make a post when new tracker is called', async () => {
```
josephjclark, line 411: Good test, thank you. Small suggestion, can we rename the test to `should make a post only when new tracker is called` (I've added the word `only`)

josephjclark (changes requested):

packages/dhis2/src/Adaptor.js
hunterachieng: @josephjclark  do you mind expounding on this?

josephjclark: `request()` is a basic HTTP request API. It should call whatever URL you pass it. It should not try to be backwards compatible, and should not try to map URLs for you.

The new tracker stuff should ONLY apply to higher-order functions like `create` and `update`, which abstract the URL from the user.

Please remove `if(shouldUseNewTracker(resolvedResourceType)` from this function!

hunterachieng: @josephjclark  I only used the new tracker on `upsert`, `create` , and `update` not any other function. The code you are pointing to above is for the `create` function

josephjclark: Ah you are right! I'm sorry, I was confused by the diff

packages/dhis2/src/Adaptor.js
```diff
@@ -442,10 +475,17 @@ export function update(
  * });
  * @example <caption>all programs for an organization unit</caption>
  * get('programs', { orgUnit: 'TSyzvBiovKh', fields: '*' });
- * @example <caption>a single tracked entity instance by a unique external ID</caption>
- * get('trackedEntityInstances', {
- *   ou: 'DiszpKrYNg8',
- *   filter: ['flGbXLXCrEo:Eq:124', 'w75KJ2mc4zz:Eq:John'],
+ * @example <caption>a single tracked entity instance by a unique entity ID. Either `program`, `trackedEntityType` or `trackedEntities` should be specified in the params. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#tracked-entities-get-apitrackertrackedentities TrackedEntities docs}</caption>
```
josephjclark, line 478: I don't think this example is correct

Looking at the docs link you (helpfully!) provided, if I wanted to get a single tracked instance by id, I would do:
```
get('tracker/trackedEntities/F8yKM85NbxW')
```

(we later show this in another example, which is actually kinda confusing - which format should I use, and when?)

Regarding the required parameters... well I don't see that stuff mentioned in the docs and generally I don't think we should get into that. Definitely not in this example, and probably not at all in the `get()` docs. 

Because it's really complicated, we can never capture it all, and if we make mistakes we'll just confuse users. Users who want to use the REST API directly need to refer to the dhis2 docs and build the URL they need.

I think overhauling the docs is way out of spec for this work. All I'll ask for this PR as the new examples we provide are super accurate and useful.

josephjclark (commented):
I think we're about there @hunterachieng. Sorry this is late, and I won't be able to get to it now until Wednesday.

I'm a bit worried about some of the docs. Big docs changes are out of scope for this work but I also think there are some innaccuracies.

Unless anyone else volunteers before Wednesday, I'll do a light sweep of the docs, do a final check, and we can release.

josephjclark:
@hunterachieng have you called the actual DHIS2 sandbox with the new APIs, or have you only tested against unit tests?

packages/dhis2/src/Adaptor.js
hunterachieng: @josephjclark  The path you have displayed above still works

hunterachieng:
> @hunterachieng have you called the actual DHIS2 sandbox with the new APIs, or have you only tested against unit tests?

@josephjclark Yes I did

josephjclark:
@hunterachieng the PR body is very out of date with the changes here, which is a big problem for audit trails and working in the open.

I'm going to start getting more strict about this -  it's super important that PRs actually do what they say they do when they are merged.

I've updated the description to be more accurate.

packages/dhis2/src/Adaptor.js
```diff
@@ -279,8 +299,8 @@ export function create(resourceType, data, options = {}, callback = false) {
  *   storedBy: 'admin',
  *   dataValues: [],
  * });
- * @example <caption>a trackedEntityInstance</caption>
- * update('trackedEntityInstances', 'IeQfgUtGPq2', {
+ * @example <caption>a tracker entity. See {@link https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import Update tracker docs}</caption>
+ * update('trackedEntityInstances', '', {
```
josephjclark, line 303: @hunterachieng I'm quite worried about this example :thinking: 
* It should be `update('trackedEntities')`. Although we (should) technically support the old format, we need examples which need the new stuff.
* The actual data structure in the example is outdated. For example, `trackedEntityInstace` (the "id") should now be `trackedEntity`. And `created` should now be `createdAt`. See the [migration guide](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker-deprecated.html#webapi_tracker_migration)
* There is no ID or path in the argument. But we should still need it. The API should not change. This makes the example misleading and wierd.

But also! Looking over the implementation, we don't nest or flatten the input data. So if we POST to `/tracker/` then the server won't understand what we're posting. Also, when we nest data for the new tracker, we need to ensure that the payload includes the id field. We should NOT need the id to be in the payload.

So if you were to run this example through the adaptor on this branch, I'm pretty sure it would fail (for a bunch of reasons!)

We need to take a really close look at all the examples to make sure they're up to date. This is important for users, but also important for us because it should verify that our adaptor works like we think it does!

josephjclark:
@hunterachieng I've found a big problem here and I wish I'd seen this earlier.

This adaptor includes a WONDERFUL set of integration tests. But they're disabled, see this ticket: https://github.com/OpenFn/adaptors/issues/207

I am really worried generally about this PR, I'm not at all sure we're doing the right thing generally, and I am low in confidence that we've implemented what we THINK we've implemented technically.

Getting the integration tests working would be a huge boost to my confidence and would help me feel better about releasing.

I will take a look this afternoon to see if I can restore the integration tests against main

hunterachieng:
> @hunterachieng I've found a big problem here and I wish I'd seen this earlier.
> 
> This adaptor includes a WONDERFUL set of integration tests. But they're disabled, see this ticket: #207
> 
> I am really worried generally about this PR, I'm not at all sure we're doing the right thing generally, and I am low in confidence that we've implemented what we THINK we've implemented technically.
> 
> Getting the integration tests working would be a huge boost to my confidence and would help me feel better about releasing.
> 
> I will take a look this afternoon to see if I can restore the integration tests against main

@josephjclark  I might have missed the integration tests but I would be happy to run and fix them until everything works as intended

josephjclark:
@hunterachieng I'd totally forgotten we even HAD integrations tests, they're so rare! They've been disabled since forever.

Let me get them in a basic working state, then I'll let you go through and fix/update them on this branch

josephjclark:
@hunterachieng I have fixed the integration test suite against main - see https://github.com/OpenFn/adaptors/pull/828

You should download that branch locally and merge it into your work (there might be one conflict on `update` so please check  carefully).

Run integration tests with `pnpm test:integration`. I expect them to pass on main but fail on your branch.

hunterachieng:
@josephjclark  and @mtuchi  I have managed to fix a few items in the adaptors and integration tests.
1. I added `resolvedPath` on the `update` function for both the `tracker` and the other resources
2. I have checked and updated tests on `create` and `update` but not yet done on `update` as I am now getting a `503`

josephjclark:
TODO: remove all support for `trackedEntityInstances` - ie, we should NOT map it to the new tracker, and we should have examples or unit tests which use this key.

This is because the datastructure for an entity on the new tracker is different to the datastructure to the old one. It's not worth pretending to maintain support.

josephjclark:
TODO: `nestArray` does NOT nest new tracker objects into the right format. In the create etc we need to ensure the data is in an array. Use `util.ensureArray`

josephjclark:
TODO: on the new tracker, we need to pass `async: false` so that the operation completes synchronously and returns data to us.

TODO: raise an issue to later support `async: true` as an option. This is kind of a big deal because we need to poll the endpoints and then return data. We need to make this seamless so that we just return the completed data to the job code.

mtuchi:
Hiya @josephjclark, please give this another round
- I have updated the tracker API implementation 
- Added https://github.com/OpenFn/adaptors/issues/836

.changeset/breezy-tigers-attend.md
```diff
@@ -0,0 +1,24 @@
...
+'@openfn/language-dhis2': major
+---
+
+Migrates the adaptor to the new Tracker API (v36+) for `trackedEntities`,
+`enrollments`, `events` and `relationships`.
+
+This release is designed for compatibility with DHIS2 v42, which drops support
+for a number of endpoints.
+
+The `create`, `update`, `upsert` and `destroy` functions will automatically map
```
josephjclark, line 11: This is out of date after your changes @mtuchi - can you please review and update these change notes?

packages/dhis2/src/Utils.js
```diff
@@ -1,4 +1,13 @@
 import { composeNextState } from '@openfn/language-common';
+import _ from 'lodash';
+
+export const { indexOf, isObject, isArray } = _;
```
josephjclark, line 4: Hey this is kinda weird- why are we re-exporting some lodash functions from this file?

It's cool to use lodash but why not just import the functions we need, where we need them? :thinking: 

Also, these are very low-value imports. What's wrong with `Array.isArray()` or  `[].indexOf` ?

mtuchi: @josephjclark 
My thinking was since lodash already have `isArray()` function we can use that instead of creating our own `isArray()` function. 

So i moved lodash to `Util.js` and export the functions that are needed for both `Util.js` and `Adaptor.js`

josephjclark: But Javascript has an isArray function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

Works great and doesn't require any complicated exports

packages/dhis2/src/Adaptor.js
mtuchi: Updated the example

.changeset/breezy-tigers-attend.md
mtuchi: Updated changeset

packages/dhis2/src/Adaptor.js
```diff
@@ -36,10 +34,16 @@ export function execute(...operations) {
 
   return state => {
     const version = state.configuration?.apiVersion;
-    if (+version >= 42)
+
+    if (+version < 42) {
+      console.warn(
+        `WARNING: This adaptor uses the new tracker APIs for create, update, upsert and destroy. Options and parameters may change, see https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html`
```
josephjclark: @mtuchi I asked for this stuff but I find myself questioning its value this morning. Do you think we  should bother with this warning?

My consideration is: if you've got an existing workflow which works for `create('events')`, and then you update the adaptor version, the options you pass into `create()` may no longer work. Because we'll use the new tracker API behind-the-scenes.

This is a serious worry - but maybe it's for the changelog and not the console?

josephjclark: Sorry I've answered my own question there :joy: I'll deal with it

josephjclark (approved):

PR #822, opened 2024-11-14
dhis2 better regex

josephjclark:
This PR shows simpler regex and a few more tests

### DIFF
--- a/packages/dhis2/test/index.test.js
+++ b/packages/dhis2/test/index.test.js
@@ -149,19 +149,40 @@ describe('get', () => {
   });
 });
 
-describe('helperfunctions', () => {
+describe.only('helperfunctions', () => {
   it('should use the new tracker for enrollments', () => {
     const result = shouldUseNewTracker('enrollments');
     expect(result).to.be.true;
   });
 
   it('should use the new tracker for events', () => {
+    const result = shouldUseNewTracker('events');
+    expect(result).to.be.true;
+  });
+
+  it('should use the new tracker for relationships', () => {
+    const result = shouldUseNewTracker('relationships');
+    expect(result).to.be.true;
+  });
+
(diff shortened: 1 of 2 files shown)


packages/dhis2/test/index.test.js
```diff
@@ -149,22 +149,43 @@ describe('get', () => {
...
+    const result = shouldUseNewTracker('relationships');
+    expect(result).to.be.true;
+  });
+
+  it('should use the new tracker for new tracker events', () => {
     const result = shouldUseNewTracker('tracker/events');
     expect(result).to.be.true;
   });
 
+  it('should use the new tracker for trackedEntityInstance', () => {
```
josephjclark, line 173: This test fails right now, but I think it should pass.

I don't think we have support right now to map `trackedEntityInstances` to `tracker/trackeEntities`. But I think we should?

We support `create('enrollments')` and `create ('trackedEntities')`, so why not `create ('trackedEntitiesInstances)`?

packages/dhis2/test/index.test.js
```diff
@@ -149,22 +149,43 @@ describe('get', () => {
...
+    const result = shouldUseNewTracker('trackedEntityInstance');
+    expect(result).to.be.true;
+  });
+
   it('should use the new tracker for trackedEntities', () => {
     const result = shouldUseNewTracker('trackedEntities');
     expect(result).to.be.true;
   });
 
+  it('should use the new tracker for new tracker generally', () => {
```
josephjclark, line 183: Hmm. Question mark on this actually.

`create` should take a resourceType - not a path. So `create('enrollments')`, not `create('tracker/enrollments')`. Because create has opinions and does stuff for you.

If you want a custom create with the new tracker, do `post('tracker/enrollments')`

PR #883, opened 2025-01-08
`salesforce` Add helper for validating and format input data

### LINKED ISSUE #556: Salesforce: Issue with mapping and upserting a nested object using Salesforce adaptor
Mapping for an object using the Salesforce adaptor that has a relationship with another object(`Openfn_Project__r`) no longer works as per below:

`'OpenFn_Project__r.Metrics_ID__c': 'id_value'`

Getting the following error when doing the mapping like the above and using the upsert function from the salesforce adaptor

```
Upserting Deal__c with externalId Contract_ID__c : {
  Contract_ID__c: 'FCA2023',
  Start_Date__c: '2023-04-02T21:00:00.000Z',
  End_Date__c: '2023-06-29T21:00:00.000Z',
  Next_Steps__c: 'No',
  Asana_Link__c: 'https://app.asana.com/0/1204270101347732/1204277345786609',
  'OpenFn_Project__r.Metrics_ID__c': '4183'
}
[R/T] ✘ Failed step job-1 after 2.95s
[R/T] ✘ AdaptorError: No such column 'OpenFn_Project__r.Metrics_ID__c' on sobject of type Deal__c
```
I tried using both the bulk function([failed job code](https://app.openfn.org/projects/159f6761-afc2-45cb-a6f2-a8f94323feba/w/c7da7ac2-927c-4b2f-aaf4-0a5db4d0e1b3?s=4b805e75-ce65-4894-b14d-7dcb15a2e372&m=expand) and [job run](https://app.openfn.org/projects/159f6761-afc2-45cb-a6f2-a8f94323feba/w/c7da7ac2-927c-4b2f-aaf4-0a5db4d0e1b3?a=59ed3d30-f6dd-4e1a-a792-edc1307f2cfe&m=expand&s=4b805e75-ce65-4894-b14d-7dcb15a2e372) using bulk function with dot operator for the mapping) and the upsert function([failed [job code](https://app.openfn.org/projects/159f6761-afc2-45cb-a6f2-a8f94323feba/w/c7da7ac2-927c-4b2f-aaf4-0a5db4d0e1b3?s=1a15d335-7227-48a6-a601-0834bcdf4a73&m=expand) and [job run](https://app.openfn.org/projects/159f6761-afc2-45cb-a6f2-a8f94323feba/w/c7da7ac2-927c-4b2f-aaf4-0a5db4d0e1b3?a=05fdb936-b187-4e5f-a86d-8b026800415c&m=expand&s=1a15d335-7227-48a6-a601-0834bcdf4a73) using upsert function with dot operator for mapping). 

The above method of mapping seems to be working in this project([checkout this blob](https://github.com/OpenFn/akf/blob/40f1a43c1266c02e2ea8c49ecdad4c8b3008acfc/jobs/loadFinancialstoSf.js#L28 ))

Here is a [successful job](https://app.openfn.org/projects/159f6761-afc2-45cb-a6f2-a8f94323feba/w/c7da7ac2-927c-4b2f-aaf4-0a5db4d0e1b3?s=ca59a364-4e1e-413e-a345-4f8dae1980ef&m=expand) that is using the following format for mapping nested objects:
`  OpenFn_Project__r: { Metrics_ID__c: 'id_value' }`

1. Is it a bug on the adaptor version? Can we modify the adaptor so that mapping with a dot operator is also acceptable?
2. Is it an intended feature, if yes can we modify the documentations?

mtuchi:
## Summary

Add utils helper `validateNoDotKeys` and `flattenData` to help with validation of input data when using `upsert`, `update` or `create` and to transform nested object into a dot notation syntax when using `bulk` operation.

Fixes #556 

## Details

When mapping a salesforce data model there is a bit of confusion on when to use the `dot notation` and when not to. Since the best practice when mapping objects is to always use a nested object for relationship, it's important that the adaptor adds a safe check when you forgot and use the `dot notation` to map salesforce relationship. This PR adds two things 

1. A validation step for `upsert`, `update` and `create` function. This is achieved by the new util helper `validateNotDotKeys`
2. A transformer to a dot notation syntax when using `bulk` function. This is done by the `flattenData` util helper

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -107,7 +107,7 @@ export function execute(...operations) {
 }
 
 /**
- * Create and execute a bulk job.
+ * Create and execute a bulk job. Nested relationships will be flattened to dot notation automatically.
  * This function uses {@link https://sforce.co/4fDLJnk Bulk API},
  * which is subject to {@link https://sforce.co/4b6kn6z rate limits}.
  * @public
@@ -129,6 +129,20 @@ export function execute(...operations) {
  *       vera__Geographic_Area__c: "Uganda",
  *       "vera__Indicator__r.vera__ExtId__c": 1001,
  *       vera__Result_UID__c: "1001_2023_Uganda",
+ *     },
+ *   ],
+ *   { extIdField: "vera__Result_UID__c" }
+ * );
+ * @example <caption>Bulk upsert with a nested relationship</caption>
+ * bulk(
+ *   "vera__Beneficiary__c",
+ *   "upsert",
+ *   [
+ *     {
+ *       vera__Reporting_Period__c: 2023,
+ *       "vera_Project": {
+ *         "Metrics_ID__c": "jfh5LAnxu1i4na"
+ *       }
  *     },
  *   ],
  *   { extIdField: "vera__Result_UID__c" }
@@ -302,7 +316,7 @@ export function bulkQuery(query, options = {}) {
 }
 
 /**
- * Create one or more new sObject records.
+ * Create one or more new sObject records. Relationships in the record should be nested and not use dot-notation syntax
  * @public
  * @example <caption> Single record creation</caption>
  * create("Account", { Name: "My Account #1" });
@@ -314,6 +328,13 @@ export function bulkQuery(query, options = {}) {
  *     Name: account.label
  *   })
  * ));
+ * @example <caption>Update a record with a relationship</caption>
+ * create("Account", {
+ *   Name: "My Account #1" ,
+ *   "Project__r": {
+ *     "Metrics_ID__c": "jfh5LAnxu1i4na"
+ *   }
+ * });
  * @function
  * @param {string} sObjectName - API name of the sObject.
  * @param {(Object|Object[])} records - Field attributes for the new record, or an array of field attributes.
@@ -643,7 +664,7 @@ export function query(query, options = {}) {
 }
 
 /**
- * Create a new sObject record, or updates it if it already exists.
+ * Create a new sObject record, or updates it if it already exists. Relationships in the record should be nested and not use dot-notation syntax
  * @public
  * @example <caption> Single record upsert </caption>
  * upsert("UpsertTable__c", "ExtId__c", { Name: "Record #1", ExtId__c : 'ID-0000001' });
@@ -652,6 +673,13 @@ export function query(query, options = {}) {
  *   { Name: "Record #1", ExtId__c : 'ID-0000001' },
  *   { Name: "Record #2", ExtId__c : 'ID-0000002' },
  * ]);
+ * @example <caption>Update a record with a relationship</caption>
+ * upsert("UpsertTable__c", {
+ *   Name: "Record #1",
+ *   "Project__r": {
+ *     "Metrics_ID__c": "jfh5LAnxu1i4na"
+ *   }
+ * });
  * @function
  * @param {string} sObjectName - API name of the sObject.
  * @magic sObjectName - $.children[?(!@.meta.system)].name
@@ -687,8 +715,9 @@ export function upsert(sObjectName, externalId, records) {
 }
 
 /**
- * Update an sObject record or records.
+ * Update an sObject record or records. Relationships in the record should be nested and not use dot-notation syntax
  * @public
+ * @function
  * @example <caption> Single record update</caption>
  * update("Account", {
  *   Id: "0010500000fxbcuAAA",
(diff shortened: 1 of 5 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -110,17 +110,18 @@ export function bulk(sObjectName, operation, records, options = {}) {
       pollInterval = 6000,
     } = resolvedOptions;
 
-    if (allowNoOp && resolvedRecords.length === 0) {
+    const flattenRecords = util.flattenData(resolvedRecords);
```
josephjclark, line 113: This not good English (because flatten is a verb form and you should use an adjective here ) - I would rename the variable `flatRecords` I  think.

packages/salesforce/src/Utils.js
```diff
@@ -134,3 +134,69 @@ export async function pollJobResult(conn, job, pollInterval, pollTimeout) {
...
+ * Flattens an array of nested objects to a maximum of two levels deep.
+ * - Each object in the array is flattened to have dot-separated keys.
+ * - Throws an error if the input is not an array or if a nested object exceeds two levels deep.
+ *
+ * @param {Array<Object>} input - An array of objects to flatten.
+ * @returns {Array<Object>} - An array of flattened objects.
+ * @throws {Error} - Throws an error if the input is not an array of objects or if nesting exceeds two levels.
+ */
+
+export function flattenData(input) {
```
josephjclark, line 148: Open question - should this be `flatten` or `removeNestings`?

When you talk about flattening an array, I think about taking an array of arrays and "flattening" them into a single array of elements.

But what we're doing here is removing the nesting of an object. 

The inverse operation would be to take an array of objects and nest them - detect any dot referenced properties and convert them into a nested object, broken on the dot. Talking about `nesting` or `validateNoNesting` is easier to understand than `validateNoDotKeys`

And generally, we  can talk about nesting objects, un-nesting objects (a bit awkward bit it works), or removing nesting of objects.

Phew :sweat_smile: 

Having written this out I'm pretty sure we should rename this to `removeNestings`

packages/salesforce/src/Utils.js
```diff
@@ -134,3 +134,69 @@ export async function pollJobResult(conn, job, pollInterval, pollTimeout) {
...
+};
+
+/**
+ * Validates that no keys in the input object or array of objects contain dots ('.').
+ * Throws an error if a key containing a dot is found.
+ *
+ * @param {Object|Array<Object>} input - The object or array of objects to validate.
+ * @throws {Error} - Throws an error if a key containing a dot is found or if the input is invalid.
+ */
+export function validateNoDotKeys(input) {
```
josephjclark, line 186: This is sort of thing (testing something and throwing) is usually known as an assertion - so I'd prefer `assertNoDotKeys` or even better `assertNoNesting`

packages/salesforce/src/Utils.js
```diff
@@ -134,3 +134,69 @@ export async function pollJobResult(conn, job, pollInterval, pollTimeout) {
...
+  const hasDotInKeys = obj => {
+    for (const key of Object.keys(obj)) {
+      if (key.includes('.')) {
+        throw new Error(`Invalid key "${key}" contains a dot.`);
```
josephjclark, line 190: We need to think about this error message

If Hunter comes along next week and happens to have an nested property name, she'll see the error `invalid key "name.first" contains a dot`

How likely is she to know what the problem is?

I recommend you import `{ throwError} ` from `@openfn/language-common`* and use that to carefully word the error, a description of the problem, and the solution. Like we do for UNEXPECTED_RELATIVE_URL in common.http.request

* If it's available on this version of salesforce? I've lost track!

packages/salesforce/test/Utils.test.js
```diff
@@ -0,0 +1,108 @@
+import { expect } from 'chai';
+import { flattenData, validateNoDotKeys } from '../src/Utils';
+
+describe('flattenData', () => {
```
josephjclark, line 4: Great tests, thank you!

I was a bit unsure about what was going on here - but I was able to look at these tests and they explained things beautifully.

packages/salesforce/src/Utils.js
mtuchi: Yeah it's hard naming function, i think `removeNestings` is the best fit for this use case although the keyword `remove` sounds like we are completely removing the keys from the object 🤷🏽

mtuchi: I have rename the function to `removeNestings`

packages/salesforce/src/Utils.js
mtuchi: Changed to `assertNoNesting`

packages/salesforce/src/Utils.js
mtuchi: I have updated the common package to `workspace*` and use the `throwError` instead of `throw new Error()`. I have also improved the error description a bit but the error code i am not sure if it's clear enough

josephjclark: Fantastic! I love it.

josephjclark (approved):

josephjclark:
Hey @mtuchi we need super clear docs on this to avoid confusion in the  future.

I've updated the changeset to remove irrelevant stuff (users don't need to know about private internal APIs we've added) and to explain the changes. I've also added docs and example to each affected function.

Can you just check that this looks all correct? Any other suggestions as to how to best communicate this?

PR #865, opened 2024-12-11
`common` Exporting common util 

mtuchi:
## Summary

Fixes namespacing of util functions in common and exports the utils properly for use by other adaptors.

## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to know!):

- [ ] Code generation (copilot but not intellisense)
- [ ] Learning or fact checking
- [ ] Strategy / design
- [ ] Optimisation / refactoring
- [ ] Translation / spellchecking / doc gen
- [ ] Other
- [x] I have not used AI

You can read more details in our [Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

PR #888, opened 2025-01-09
re-organize helpers and util

mtuchi:
## Summary

Fixes #

## Details

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/http/src/util.js
+++ b/packages/http/src/util.js
@@ -1,144 +1,4 @@
-import { composeNextState } from '@openfn/language-common';
-import {
-  request as commonRequest,
-  makeBasicAuthHeader,
-  expandReferences,
-  logResponse,
-  encode,
-  decode,
-  uuid,
-} from '@openfn/language-common/util';
-
-import * as cheerio from 'cheerio';
-import cheerioTableparser from 'cheerio-tableparser';
-
-export function addAuth(configuration, headers) {
-  if (headers.Authorization) {
-    return;
-  }
-
-  const { username, password, access_token } = configuration ?? {};
-
-  if (access_token) {
-    Object.assign(headers, { Authorization: `Bearer ${access_token}` });
-  } else if (username && password) {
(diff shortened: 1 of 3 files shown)


packages/http/src/util.js
```diff
@@ -148,7 +8,7 @@ export {
    * @param {string} data - The string to be encoded.
    * @returns {string} - The Base64 encoded string.
    * @example <caption>Encode a string</caption>
-   * const encoded = Util.encode('Hello World');
+   * const encoded = util.encode('Hello World');
```
josephjclark, line 11: Good catch!

mtuchi:
Fair point. I wish there was a better way of doing this because if i add documentation to any function in util i have to add `@private` if i don't that function to appear but still i will be able to use it in my job code. 

Let me revert and fix the typo only for this PR so that we can get #865 released

josephjclark:
>  I wish there was a better way of doing this because if i add documentation to any function in util i have to add @private if i don't that function to appear but still i will be able to use it in my job code.

Wait are you sure?

My understanding (and I've written a lot of code around this stuff, but I can definitely be wrong) is that a function needs `@public` and `@function` before it's documented.

Because we DO use private internal functions. So I think things should be private by default, and only made public when someone's made the effort to properly document them.

I'll have to take a look at this to be certain before moving on.

josephjclark:
The docgen in tools says this:
```
// Filter items which are not marked as @public
  templateData = templateData.filter(
    data => data.kind === 'typedef' || data.access === 'public'
  );
```

Let me assert that access is `private` by default though

josephjclark:
Yeah, confirmed: `common.alterState` is not currently documented:

image

It has no explicit access (so it's private by default)

```
/**
 * alias for "fn()"
 * @function
 * @param {Function} func is the function
 * @returns {Operation}
 */
export function alterState(func) {
  return fn(func);
}
```

josephjclark (approved):

PR #894, opened 2025-01-09
878 result objects

mtuchi:
## Summary

Fixes #

## Details

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


packages/salesforce/src/Adaptor.js
```diff
@@ -338,9 +338,11 @@ export function create(sObjectName, records) {
 
     return connection
       .create(resolvedSObjectName, resolvedRecords)
-      .then(recordResult => {
-        console.log('Result : ' + JSON.stringify(recordResult));
-        return composeNextState(state, recordResult);
+      .then(result => {
```
josephjclark, line 341: Shouldn't we use exactly the same structure for a single item?

It still makes perfect sense, and makes documentation and code cleaner (including in job code!)

packages/salesforce/src/Utils.js
```diff
@@ -134,3 +134,36 @@ export async function pollJobResult(conn, job, pollInterval, pollTimeout) {
     attempt++;
   }
 }
+
+export function formatResults(input) {
```
josephjclark, line 138: The integration tests are cool but I would value unit tests against this function way more!

I won't block the PR for tests on this though (especially as I've requested a change). If you have time for a few tests that's great. If you don't, don't worry.

What I'm really trying to say is: consider unit testing before integration testing. Also when you write a nice clean standalone util function like this, it's a great candidate for unit tests!

PR #895, opened 2025-01-09
common: bump versions

josephjclark:
Checking version diffs before merging


josephjclark:
Most adaptors bumped here aren't technically affected by any of the changes in common. But they get a slightly more secure base and it's not worth the time to cherry-pick which versions to bump. So we'll roll with it!

PR #897, opened 2025-01-10
Support PAT for dhis2

### LINKED ISSUE #697: Support PAT for DHIS2
@FaraiMutero , if PAT support is important for the DHIS2 community I think _this_ is where it would need to end up

1. add an optional `Personal Access Token` field to the configuration schema for dhis2 here - https://github.com/OpenFn/adaptors/blob/main/packages/dhis2/configuration-schema.json
2. then add some logic here that uses a PAT if provided, but defaults back to username/password if there's not PAT - https://github.com/OpenFn/adaptors/blob/main/packages/dhis2/src/Client.js#L15

One more Q here: how about Oauth2? We support generic Oauth2 clients, so maybe we should also extend the DHIS2 adaptor to support Oauth2

PiusKariuki:
## Summary

Fixes #697

## Details

1. Added a personalAccessTokenField to the configuration schema for the dhis2 adaptor
2. Added a conditional, that uses the PAT to authenticate if it id provided but defaults back to username/password if it it not

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/dhis2/configuration-schema.json
+++ b/packages/dhis2/configuration-schema.json
@@ -46,6 +46,15 @@
       "examples": [
         "v2"
       ]
+    },
+    "personalAccessToken": {
+      "title": "Personal Access Token",
+      "type": "string",
+      "description": "Personal access token",
+      "minLength": 1,
+      "examples": [
+        "d2pat_vreYExmi0yh92Soy8CCF2zHG0wRjID494035128247"
+      ]
     }
   },
   "type": "object",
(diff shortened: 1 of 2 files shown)


packages/dhis2/configuration-schema.json
```diff
@@ -46,6 +46,15 @@
       "examples": [
         "v2"
       ]
+    },
+    "personalAccessToken": {
```
josephjclark, line 50: A cleaner key would be nice here.

I would prefer we used `accessToken`, `apiToken` or even just `pat` as the key name. I think `pat` is my favourite!

PiusKariuki: I agree `pat` is better

josephjclark (approved):

PR #898, opened 2025-01-11
Common add assert() function

### LINKED ISSUE #876: common: add `assert()` function
Assert should take some expression or function and throw an error if either resolves to `false`.

This allows us to abort execution in the event of a fail state. It's a good replacement for eg salesforce's `failOnError`, and can be applied equally to all jobs and adaptors. It doesn't need special API support. Although the operation in question does need to provide enough feedback to make a useful assertion on - ie, it must write something useful to state.data.

It should take a second argument as an error message which will be returned.

Eg

```
assert($.data.errors === 0, "Errors detected in upsert")
assert($.data.length, "No patient data found!")
```

PiusKariuki:
## Summary

Added an 'assert' function to  the common adaptor

Fixes #876

## Details

Created an assert function that aborts  execution when it fails. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -903,3 +903,27 @@ export function cursor(value, options = {}) {
     return state;
   };
 }
+
+/**
+ * An assert function that throws an error when an expression or function resolves to false.
+ * Otherwise, it returns true.
+ * @public
+ * @function
+ * @example
+ * assert("a" === "b")
+ * @param expression
+ * @param errorMessage
+ * @returns {(function(*): (boolean|undefined))|*}
+ */
+export function assert (expression, errorMessage) {
+  return state => {
+    const resolvedValue = expandReferences(expression)(state);
+
+    if(resolvedValue){
+      return true;
+    }
+    else{
+      throw new Error(errorMessage);
+    }
+  }
+}
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -917,7 +917,7 @@ export function cursor(value, options = {}) {
  */
 export function assert (expression, errorMessage) {
   return state => {
-    const resolvedValue = expandReferences(expression)(state);
+    const [resolvedValue] = newExpandReferences(state,expression);
 
     if(!resolvedValue){
       throw new Error(errorMessage);
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -910,10 +910,10 @@ export function cursor(value, options = {}) {
  * @public
  * @function
  * @example
- * assert("a" === "b")
- * @param expression
- * @param errorMessage
- * @returns {(function(*): (boolean|undefined))|*}
+ * assert('a' === 'b', '"a" is not equal to "b"')
+ * @param {any} expression  - The expression or function to be evaluated.
+ * @param {string} errorMessage - The error message thrown in case of a failed state.
+ * @returns {boolean} Always returns true if the assertion passes.
  */
 export function assert (expression, errorMessage) {
   return state => {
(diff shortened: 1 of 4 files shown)


packages/common/src/Adaptor.js
```diff
@@ -903,3 +903,26 @@ export function cursor(value, options = {}) {
...
+ * @function
+ * @example
+ * assert('a' === 'b', '"a" is not equal to "b"')
+ * @param {any} expression  - The expression or function to be evaluated.
+ * @param {string} errorMessage - The error message thrown in case of a failed state.
+ * @returns {boolean} Always returns true if the assertion passes.
+ */
+export function assert (expression, errorMessage) {
+  return state => {
+    const resolvedValue = expandReferences(expression)(state);
```
josephjclark, line 920: This is a bit confusing I know but can you use `newExpandReferences` instead? It's a new function which does exactly the same thing but with a cleaner signature:
```
const [value] = expandReferences(state, expression)
```

This expandReferences function is due to be deprecated (we just haven't pressed on and done that yet)

packages/common/src/Adaptor.js
```diff
@@ -903,3 +903,26 @@ export function cursor(value, options = {}) {
...
+
+    if(!resolvedValue){
+      throw new Error(errorMessage);
+    }
+
+    return true;
```
josephjclark, line 926: This needs to return state, otherwise the state object will be lost to the rest of the job.

For example, a job like this won't work properly
```
fn(state => {
   state.x = 10;
   return state;
});
assert((state)  => state..x > 1)
fn(state => {
  console.log(state) // !! state here will not be what you expect! (probably undefined or empty object)
  return state
})
```

To test this adaptor properly, you can use the CLI :)

- Set an env var  `OPENFN_ADAPTORS_REPO` to point to your local `adaptors` repo
- If you haven't already, install the cli `npm install -g @openfn/cli`
- Build the common adaptor (from `packages/common`, run `pnpm build` or `pnpm build --watch`)
- Run a test job with `openfn -ma common path/to/expression.js`. The `-m` flag tells the CLI to load adaptors from this adaptors monorepo. Make sure adaptors are built before you run!

packages/common/test/index.test.js
```diff
@@ -1046,3 +1047,47 @@ describe('group', () => {
     expect(result.data).eql({ a: [{ x: 'a', y: { z: 'a' } }] });
   });
 });
+
+describe("assert", () => {
```
josephjclark, line 1051: Thank you for including unit tests! :pray:

packages/common/src/Adaptor.js
```diff
@@ -903,3 +903,26 @@ export function cursor(value, options = {}) {
...
+ * @param {any} expression  - The expression or function to be evaluated.
+ * @param {string} errorMessage - The error message thrown in case of a failed state.
+ * @returns {boolean} Always returns true if the assertion passes.
+ */
+export function assert (expression, errorMessage) {
+  return state => {
+    const resolvedValue = expandReferences(expression)(state);
+
+    if(!resolvedValue){
+      throw new Error(errorMessage);
```
josephjclark, line 923: If the user does not provide a message, what should the default be? :thinking:  We should have something generic like `assertion statement failed with ${resolvedValue}`

josephjclark (changes requested):
Thanks @PiusKariuki - what a pleasant surprise! Thank you for such a fully-featured and comprehensive PR, we really appreciate it.

There's one big issue with the return which we'll need to resolve, plus a couple of minor comments which it would be nice to address :pray:

packages/common/src/Adaptor.js
PiusKariuki: Oh yeah makes total sense. I will resolve this

josephjclark (approved):
Thank you @PiusKariuki!

As on the other PR, can you please generate a changeset file - `pnpm changeset`. This will also be a minor release.

Tomorrow I'll merge both of these changes to a branch in the repo, run a bit of QA, and release to the world!

PiusKariuki:
Thank you for the very helpful reviews @josephjclark !
I'll keep contributing!

PR #906, opened 2025-01-14
common: request throws if parseAs is json and the response body is empty

### LINKED ISSUE #689: common: request throws if parseAs is json and the response body is empty
In `common`, we provide a set of HTTP helper functions which are re-used by many adaptors - see `common/src/util/http.js`. Note that these are functions, not operations. They're designed to be used by other adaptors, not in job code directly.

These helper functions are wrappers around `request`, which  has some very useful automation to parse a response body into a JSON object. It'll do this if the response content type is application/json, or if the `parseAs` option is set to `json`.

This is all fine, but if the response body is empty, then `readResponseBody()` will throw an exception.

Empty response bodies frequently occur on POST and PUT requests, and for APIs which return `204`  statuses

We need to do two things:

1) If readResponseBody throws for any reason, we need to catch it and log a good error to the console. Maybe something like "Error parsing the response body [from URL, if we have it]", and include the parseAs flag, the content-type header on the response, and the body length. Use the `util/throwError` function to this.

2) if the response body is empty (I think you can do `response.body.length === 0`, we should NOT try to parse the response body at all. Just return `undefined` from `readResponseBody`

# Reproduction  & Test

You should be able to create a unit test in `common/test/util/http.test.js`. Look for the `request function` block and create a  test that if parseAs is 'json', and the server response is `204`, the `result.body` should be undefined

hunterachieng:
## Summary

An error is thrown when we use `parseAs=json` if the response body is undefined when `PUT` or `POST` requests are made

Fixes #689

## Details

In `common`, we mainly parse the response body, but since we might get an empty body with different requests, we are returning `undefined` for the body else parse as expected

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


josephjclark:
Hi @hunterachieng, can you update the PR description please? You need to:
- Include a brief description of the fix
- update the AI disclaimer
- remove any template sections that aren't relevant to this PR

.changeset/gold-kings-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-common': minor
```
josephjclark, line 2: This is probably a patch fix because we are not introducing breaking changes (so it's not a major), nor are we adding stuff to the API (so it's not a minor)

packages/common/src/util/http.js
```diff
@@ -195,7 +196,14 @@ export async function request(method, fullUrlOrPath, options = {}) {
 
   await assertOK(response, errors, url, method, startTime);
 
-  const responseBody = await readResponseBody(response, parseAs);
+  const responseBody = await readResponseBody(response, parseAs).catch(() => {
+    throwError(response.statusCode, {
```
josephjclark, line 200: I was expecting this error handling to inside `readResponseBody`.

Remember `readResponseBody` is called in two places. We don't want to duplicate the error handling.

josephjclark (changes requested):

packages/common/test/util/http.test.js
```diff
@@ -215,6 +215,41 @@ describe('request function', () => {
...
+    const response = await request('PUT', 'https://www.example.com/api', {
+      parseAs: 'json',
+      body: { id: 2 },
+    });
+
+    expect(response.statusCode).to.eql(200);
+    expect(response.body).to.eql(undefined);
+  });
+
+  it('should throw an error if there is no content-length header and an empty response body', async () => {
```
josephjclark, line 237: This is an EXCELLENT unit test because it demonstrates some really important behaviour in this function.

Looking at this unit test, I'm thinking, "we haven't really got this behaviour right yet". Because if the body is empty we really shouldn't throw an error, no matter what the state of the HTTP headers.

However, I understand WHY we have that behaviour and I don't think I'm going to ask you to change anything. I just want to highlight how useful this test is in explaining and documenting the behaviour of the functoin.

josephjclark (commented):
Almost there - just a couple of comments left to address. It's worth getting this right: change to common can affect every other adaptor, so getting it right now will save us problems in other adaptors in the future (if we'd done this right in the first place, you wouldn't be having to do this at all!)

josephjclark:
Hi @mtuchi - can you run `pnpm changeset version` on this PR, push the diffs, and merge? Hunter is blocked by it.

Remember we don't have to push tags any more :) Just push the version bumps and merge to release. Don't worry about the warnings when you version.

Thanks! I'll be online around 10am GMT if there are any problems. You'll be online way before me and can unblock Hunter quicker :pray:

mtuchi:
I have push the version bump and fix merge conflict.  I will force merge this PR to unblock hunter since it's not approved yet but Joe gave me a green light ✅

PR #907, opened 2025-01-14
`salesforce` Add support for `asyncMode` in bulk operation

### LINKED ISSUE #286: Add option to `salesforce` for `asyncMode: true/false` on bulk(...)
Right here, if the user has specified `asyncMode` in their bulk options, we should exit, returning `batchInfo` to state.references.

If they have not provided this extra `asyncMode` option, we should continue with the `.then()` and return all record info to state as we currently do. (No change there.)

mtuchi:
## Summary

Add `asyncMode` option in bulk operation. If you don't provided this extra `asyncMode` option, we return all record info to `state.data` as we currently do.

Fixes #286 

## Details

- If `asyncMode` is `true` we return `batchInfo`
- Formatted the results to `{ success: true, batches: [], errors: []}` if `asyncMode: true`
- Add docs for `state.data.batches`
- Add integration test when `asyncMode: true`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -66,7 +66,7 @@ import flatten from 'lodash/flatten';
  * @property {boolean} [failOnError=false] - Fail the operation on error.
  * @property {integer} [pollTimeout=240000] - Polling timeout in milliseconds.
  * @property {integer} [pollInterval=6000] - Polling interval in milliseconds.
- * @property {boolean} [asynMode=true] - Asynchronous mode.
+ * @property {boolean} [asynMode=false] - Use asyn mode.
  */
 
 /**
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -160,6 +160,7 @@ export function execute(...operations) {
  * @param {array} records - an array of records, or a function which returns an array.
  * @param {BulkOptions} [options] - Options to configure the request. In addition to these, you can pass any of the options supported by the {@link https://bit.ly/41tyvVU jsforce API}.
  * @state {SalesforceResultState}
+ * @state data.batches - Array of batch Infomation
  * @returns {Operation}
  */
 export function bulk(sObjectName, operation, records, options = {}) {
@@ -178,7 +179,7 @@ export function bulk(sObjectName, operation, records, options = {}) {
       allowNoOp = false,
       pollTimeout = 240000,
       pollInterval = 6000,
-      asyncMode = true,
+      asyncMode = false,
     } = resolvedOptions;
 
     const flatRecords = util.removeNestings(resolvedRecords);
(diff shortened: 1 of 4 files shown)


packages/salesforce/src/Adaptor.js
```diff
@@ -220,36 +223,54 @@ export function bulk(sObjectName, operation, records, options = {}) {
...
-              .on('queue', function (batchInfo) {
-                const batchId = batchInfo.id;
-                var batch = job.batch(batchId);
-                batch.poll(pollInterval, pollTimeout);
-              })
-              .then(async res => {
+            if (asyncMode) {
+              return batch.on('queue', async function (batchInfo) {
+                console.info('Batch queued. Batch ID:', batchInfo.id);
+                resolve({ batchInfo });
```
josephjclark, line 229: A promise can only resolve once. If there are multiple queue events, for multiple chunks, then they'll never report back to user code. In other words, we'll only ever report back the first batch here.

mtuchi: I have tested multiple batches by reducing the batch size to 1 and run integration test that insert two records with the option `asynMode: true`, I can see that i am getting two batches records after the bulk operation is done.

packages/salesforce/src/Adaptor.js
```diff
@@ -66,6 +66,7 @@ import flatten from 'lodash/flatten';
  * @property {boolean} [failOnError=false] - Fail the operation on error.
  * @property {integer} [pollTimeout=240000] - Polling timeout in milliseconds.
  * @property {integer} [pollInterval=6000] - Polling interval in milliseconds.
+ * @property {boolean} [asynMode=false] - Use asyn mode.
```
josephjclark, line 69: Typo

josephjclark:
@mtuchi where are we on this? I'm still not comfortable with this promise that returns multiple times.

When you call `resolve()`, you are saying to the Javascript: "I have finished with my promise, please continue executing". Until then, the runtime is patiently waiting for the promise to end.

In your code, every time the queue event is called, you're trying to resolve the promise again. 

There are two really important things here:
1) A promise can only return once. If you do `promise.then()` twice, you'll get the same value back twice. Your code is sort of working because it returns the same array twice - it's just that the array itself has changed
2) Each Operation is a Promise, and our Runtime waits for each Operation to resolve before executing the next one (or moving on to the next workflow step). Your adaptor code is "exiting" the operation after the first queue call, and so our runtime carries on. What happens when the promise resolves a second time? I actually don't know - but we can't trust the order of events that follows next.

I can't merge this as-is, it's too unsafe.

What we actually want to do is resolve the promise after the last queue event has triggered. But we have no way of knowing that, right? jsforce gives us no "queue finished event.

See the issue for a possible solution, although I'm not totally sure I trust that either :thinking:

josephjclark:
[promises demo](https://playcode.io/2240271)

mtuchi:
Hiya @josephjclark i have forgotten to close this PR,  from our last conversation we don't want this implementation and we agreed to put more effort on updating jsforce instead.

PR #916, opened 2025-01-16
commcare: Update request documentation for query params and fix response body error

### LINKED ISSUE #903: commcare: documentation issues on `request()`
Image

We've hit some issues on commcare's `request` docs
* The `options` parameter is misleading. The value passed as `options` will be passed as the URL query. So actually, we should rename this parameter to `params` (which is consistent with `commcare.get`). A good description would be `An object of query parameters to be encoded into the URL.`. We should include an example of a request with a query and show how that maps to the actual URL that will be called (like we do with get)
* RequestOptions is not a link. Turns out there is no typedef for RequestOptions. But actually there should not be - `options` should have a type of `object`, not RequestOptions

There is an argument that `options` should actually be options, in which case it accepts `{ headers, query, parseAs ...}` options which get fed into common http.

But that's not how the API works right now (I suspect this is a bug) and I don't want to raise a breaking change today)

hunterachieng:
## Summary

Updated the `options` sections in the `request` function to query parameters, and fix the response body when result is undefined.

Fixes #903 

## Details

The `options` section in the docs is misleading as the items are passed to the URL, and therefore changed to `params`. When a request is made and the response is an empty body, an error is thrown. Therefore a fix for handling the undefined aspect in the body was implemented

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/yellow-lions-double.md
+++ b/.changeset/yellow-lions-double.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-commcare': patch
----
-
-Update commcare request docs to read query parameters instead of options, and fix the empty body response received when specific requests are made.
--- a/.changeset/yellow-lions-double.md
+++ b/.changeset/yellow-lions-double.md
@@ -2,4 +2,4 @@
 '@openfn/language-commcare': patch
 ---
 
-Update commcare request docs to read query parameters instead of options
+Update commcare request docs to read query parameters instead of options, and fix the empty body response received when specific requests are made.
--- a/.changeset/yellow-lions-double.md
+++ b/.changeset/yellow-lions-double.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-commcare': patch
+---
+
+Update commcare request docs to read query parameters instead of options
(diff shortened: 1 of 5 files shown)


josephjclark:
Hi @hunterachieng, the PR title and body are quite misleading. Can you please update them to be more specific? This actually fixes the 204 thing right?

It's OK to mention fixes which don't have an issue behind them. But it _really_ scares me when PRs listed as "just documentation" include behaviour/code  changes.

hunterachieng:
> Hi @hunterachieng, the PR title and body are quite misleading. Can you please update them to be more specific? This actually fixes the 204 thing right?
> 
> It's OK to mention fixes which don't have an issue behind them. But it _really_ scares me when PRs listed as "just documentation" include behaviour/code changes.

@josephjclark  I have fixed the descriptions

.changeset/yellow-lions-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-commcare': patch
+---
+
+Update commcare request docs to read query parameters instead of options
```
josephjclark, line 5: This needs updating. The behavioural change is way more important than the docs change

hunterachieng: @josephjclark I have fixed it

josephjclark (approved):

PR #920, opened 2025-01-17
Common add as() function

### LINKED ISSUE #800: common: add something like a `save()` or `as()` function
A common use case is to run an operation and then to save the result back to state on a key other than data

Like this:
```
get('fixture/?fixture_type=option_set_mapping', {}, state => {
  state.lookup_table.option_set = state.data;
  return state;
});
```

This is really quite painful.

What if we had a wrapper operation that was able to take any state.data object and proxy/alias/push it to some other key:

```
save('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'))
```

The implementation is something like this:
```
function save(key, op) {
  return async (state) => {
       const data = state.data // save the incoming state
       const result  = await(op(state))
       state[key]=result;
       state.data = data;       
       return state;
  }
}
```
I don't like name though. Some ideas:

```
keyby('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'))
key('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'))
alias('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'))
proxy('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'))
```

PiusKariuki:
## Summary

Performs and operation and saves the result in a pre-defined key while preserving any values in the `data` key.

Fixes #800

## Details

I save the initial state `data` value in a variable then after I perform the operation I return that value bak in the `data` key in state.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -932,3 +932,35 @@ export function assert(expression, errorMessage) {
     return state;
   };
 }
+
+
+/**
+ * Runs an operation and saves the return value to a pre-defined key in state
+ * @public
+ * @function
+ * @example
+ * as('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'));
+ * @param {string} key - The key used to store the return value in the state object
+ * @param {any} op - The operation being executed
+ * @returns {operation}
+ */
+export function as (key, op) {
+  return async state => {
+    const data = state.data; // save incoming state before it is mutated by the operation
+
+    const [resolvedKey] = newExpandReferences(state, key);
+    try {
+      const result = await op(state);
+
+      const [resolvedResult] = newExpandReferences(state, result);
+
+      state[resolvedKey] = resolvedResult;
+    } catch (e) {
+      console.log(`Error executing operation: ${e.message}`);
+    } finally {
+      state.data = data;
+    }
+
+    return state;
+  }
+}
--- a/packages/common/test/index.test.js
+++ b/packages/common/test/index.test.js
@@ -5,6 +5,7 @@ import { request, MockAgent, setGlobalDispatcher } from 'undici';
 import testData from './fixtures/data.json' assert { type: 'json' };
 import {
   arrayToString,
+  as,
   chunk,
   combine,
   cursor,
@@ -33,6 +34,7 @@ import {
   assert as assertCommon,
 } from '../src/Adaptor';
 import { startOfToday } from 'date-fns';
+import { expandReferences as newExpandReferences } from '../src/util';
 
 const mockAgent = new MockAgent();
 setGlobalDispatcher(mockAgent);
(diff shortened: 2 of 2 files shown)


PiusKariuki:
@josephjclark What do you think about this?

josephjclark:
@PiusKariuki That's actually really nice - `as('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'));` reads really nicely.

This has been sitting at the back of my mind for a while and I didn't think it would work well. But seeing it here I love it!

Let me circulate for some feedback within the team. Thank you!

packages/common/test/index.test.js
```diff
@@ -1110,3 +1112,57 @@ describe('assert', () => {
...
+      return state;
+    };
+  };
+
+  it('it stores the result in the defined key', async() => {
+    let state = { data: 'Lorem Ipsum' };
+
+    const { foo: {data: fooData} } = await as('foo', mockOperation('https://localhost:1/as'))(state);
+
+    expect(fooData).to.eql('this data object has been changed');
```
josephjclark, line 1149: I actually expect this to be more like:
```
const result = await as('foo', () => ({ data: 'bar' }))(state);
expect(result).to.eql({ foo: 'bar' });

```

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,35 @@ export function assert(expression, errorMessage) {
...
   };
 }
+
+
+/**
+ * Runs an operation and saves the return value to a pre-defined key in state
+ * @public
+ * @function
+ * @example
+ * as('lookup_table.option_set', get('fixture/?fixture_type=option_set_mapping'));
```
josephjclark, line 942: This example is interesting. If the key has dot notation, should it be a path? This would be easy to implement with lodash.set. But I also don't know the use-case

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,35 @@ export function assert(expression, errorMessage) {
...
+ * @param {string} key - The key used to store the return value in the state object
```
josephjclark, line 943: Note for me: Perhaps as well as the string, we can accept an object which maps stuff. So `"key"` and `{ data: "key" }` would be equivalent.

This would bypass the problems of a) only `state.data` gets mapped and `state.data` is kind of arbitrary, and b) what if I want to map multiple state keys, like http data AND response?

Please don't implement this yet, it might be too complicated

josephjclark (commented):
Some comments but I'd like to sleep on this and speak to the team before committing it to the public!

josephjclark:
A nice real use-case for this:
```
collections.get('cce-data-dhis2', { key: `*:*:${$.syncedAt}*` }).then(state => {
  state.cceData = state.data;
  return state;
});
```
Becomes:
```
as('cceData', collections.get('cce-data-dhis2', { key: `*:*:${$.syncedAt}*` }));
```

josephjclark:
I suppose if I was to get creative I might say:
```
collections.get('cce-data-dhis2', { key: `*:*:${$.syncedAt}*` }).as('cceData)
```
But that requires compiler or runtime magic. Probably too much of it.

josephjclark:
I would like to pick this up again please - let's go with `as()` because I think it's a nice pattern. And if no-one uses it then no problem, we'll remove it later.

packages/common/src/Adaptor.js
josephjclark: Let's go ahead and use lodash.set to support paths in `as`. I don't really know why, but this is more of a WHY NOT situation

packages/common/src/Adaptor.js
josephjclark: Yes, too complicated. Maybe we'll add it later if there's a demand.

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,35 @@ export function assert(expression, errorMessage) {
...
+ */
+export function as (key, op) {
+  return async state => {
+    const data = state.data; // save incoming state before it is mutated by the operation
+
+    const [resolvedKey] = newExpandReferences(state, key);
+    try {
+      const result = await op(state);
+
+      const [resolvedResult] = newExpandReferences(state, result);
```
josephjclark, line 955: No need to expand the result :thinking:

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,35 @@ export function assert(expression, errorMessage) {
...
+
+      state[resolvedKey] = resolvedResult;
```
josephjclark, line 957: This needs to be `state[resolvedKey] = result.data`;

We're not trying to nest the new state object under a new key - we're just trying to alias `state.data` to `state.whatever`

josephjclark:
Whoever picks this up needs to:
1) Create a new branch from main
2) Merge this PR into that branch
3) Pick up the comments and finish the work

josephjclark:
Closing in favour of #1252

PR #923, opened 2025-01-17
openmrs: add a generic http request

### LINKED ISSUE #889: Phase 1: Add generic HTTP request
See parent issue #887 for details.

hunterachieng:
## Summary

Implement a generic `http` request to allow both `fhir` and `non-fhir` requests to be made

Fixes #889 

## Details

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/odd-bobcats-sell.md
+++ b/.changeset/odd-bobcats-sell.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-openmrs': minor
+---
+
+Implement namespaced http generic helpers
--- a/.changeset/odd-bobcats-sell.md
+++ b/.changeset/odd-bobcats-sell.md
@@ -2,4 +2,4 @@
 '@openfn/language-openmrs': minor
 ---
 
-Implement namespaced http generic helpers
+Implement namespaced http.request() function. The function makes a call against the `instanceUrl` and the path provided, while allowing manipulation to the API call as needed.
--- a/.changeset/odd-bobcats-sell.md
+++ b/.changeset/odd-bobcats-sell.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-openmrs': minor
----
-
-Implement namespaced http.request() function. The function makes a call against the `instanceUrl` and the path provided, while allowing manipulation to the API call as needed.
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -17,7 +17,7 @@ export const prepareNextState = (state, response, callback) => {
 
 export async function request(state, method, path, options = {}) {
   const {
-    baseUrl = 'https://demo.openmrs.org/openmrs',
+    baseUrl = state.configuration.instanceUrl,
     query = {},
     data = {},
     headers = { 'content-type': 'application/json' },
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -7,6 +7,7 @@ import {
 
 /**
  * Options object
+ * @global
  * @typedef {Object} OpenMRSOptions
  * @property {object} query - An object of query parameters to be encoded into the URL
  * @property {object} headers - An object of all request headers
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -27,7 +27,7 @@ export const prepareNextState = (state, response, callback) => {
 
 export async function request(state, method, path, options = {}) {
   const {
-    baseUrl = '',
+    baseUrl = 'https://demo.openmrs.org/openmrs',
     query = {},
     data = {},
     headers = { 'content-type': 'application/json' },
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -5,6 +5,16 @@ import {
   logResponse,
 } from '@openfn/language-common/util';
 
+/**
+ * Options object
+ * @typedef {Object} OpenMRSOptions
+ * @property {object} query - An object of query parameters to be encoded into the URL
+ * @property {object} headers - An object of all request headers
+ * @property {object} body - The request body (as JSON)
+ * @property {string} baseUrl - The base url for the request
+ * @property {string} [parseAs='json'] - The response format to parse (e.g., 'json', 'text', or 'stream')
+ */
+
 export const prepareNextState = (state, response, callback) => {
   const { body, ...responseWithoutBody } = response;
   const nextState = {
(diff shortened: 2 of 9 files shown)


packages/openmrs/src/Utils.js
```diff
@@ -15,28 +15,39 @@ export const prepareNextState = (state, response, callback) => {
...
 };
 
-export async function request(state, method, path, data, params) {
-  const { instanceUrl, username, password } = state.configuration;
-  const headers = makeBasicAuthHeader(username, password);
+export async function request(state, method, path, options = {}) {
 
-  const options = {
+  const {
+    baseUrl = '',
```
josephjclark, line 21: If we're going to default the baseURL, we should default it to something like the main openMRS platform (assuming there is such a thing).

Alternatively, and probably better, is we throw an error if baseURL is not set. Because we shouldn't allow users to call out to non openMRS systems from this adaptor

packages/openmrs/src/http.js
```diff
@@ -0,0 +1,43 @@
+import { expandReferences } from '@openfn/language-common/util';
+import * as util from './Utils';
+
+/**
+ * State object
+ * @typedef {Object} OpenMRSOptions
```
josephjclark, line 6: Is this a duplicate typedef? I don't actually know if this works, but can we re-use typedefs across files? 

Ideally `util.js` would define the set of options, and we reference the type in the docs here, rather than duplicating it.

packages/openmrs/src/http.js
```diff
@@ -0,0 +1,43 @@
...
+ * @property {object} query - An object of query parameters to be encoded into the URL
+ * @property {object} headers - An object of all request headers
+ * @property {object} body - The request body (as JSON)
+ * @property {string} baseUrl - The base url for the request
+ */
+
+/**
+ * Make a HTTP request to any OpenMRS endpoint
+ * @example
+ * request("GET","http://msf-ocg-openmrs3-dev.westeurope.cloudapp.azure.com/openmrs/ws/rest/v1/patient/d3f7e1a8-0114-4de6-914b-41a11fc8a1a8", {
```
josephjclark, line 16: :exclamation: Important! We do not want to allow absolute URLs here. Users should set state.configuration.instanceUrl

I actually think we should throw if an absolute URL is passed

Am I right in thinking that the common adaptor's request will throw if the url is absolute and baseURL has a value?

packages/openmrs/src/Utils.js
hunterachieng: @josephjclark  we have two different urls and these can change anytime. Should we check for both of them?

packages/openmrs/src/http.js
hunterachieng: @josephjclark  I did discover a way to share `typedef` across files

packages/openmrs/src/http.js
hunterachieng: Common does not throw an error, but rather `commcare` flags with a `404` if both the `instanceUrl` and `path` are absolute urls

josephjclark:
QA note: check the built docs locally and make sure everything is OK

packages/openmrs/src/http.js
josephjclark: Wow great work!

packages/openmrs/src/Utils.js
josephjclark: Can you say more about this? I don't think I understand

hunterachieng: @josephjclark  in openmrs, one can give either `https://demo.openmrs.org/openmrs`, or `http://msf-ocg-openmrs3-dev.westeurope.cloudapp.azure.com/openmrs`.

josephjclark:
@hunterachieng dang, that's a shame.

Don't spend much time on this. I was actually hoping the typedef would be available "globally" and you wouldn't need  to do anything to be able to use it. Or do some simple namespacing.

So if you can't resolve it in less than half an hour, please just copy and paste the typedef

packages/openmrs/src/Utils.js
```diff
@@ -7,6 +7,7 @@ import {
 
 /**
  * Options object
+ * @global
```
josephjclark, line 10: We should probably remove this if we're not keeping it

josephjclark:
@hunterachieng can you please generate a Minor changeset for this feature?

packages/openmrs/src/http.js
```diff
@@ -0,0 +1,46 @@
+import { expandReferences } from '@openfn/language-common/util';
+import * as util from './Utils';
+
+/**
+ * Options object
+ * @typedef {Object} OpenMRSOptions
+ * @property {object} query - An object of query parameters to be encoded into the URL
+ * @property {object} headers - An object of all request headers
+ * @property {object} body - The request body (as JSON)
+ * @property {string} baseUrl - The base url for the request
```
josephjclark, line 10: We should not document baseUrl here. Technically we should not support it.

The `http.request` function is ALWAYS relative to the baseUrl on config. Users should not be setting this.

hunterachieng:
> @hunterachieng can you please generate a Minor changeset for this feature?

Done

packages/openmrs/src/http.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Make a HTTP request to any OpenMRS endpoint
+ * @example
+ * request("GET",
+ *   "/ws/rest/v1/patient/d3f7e1a8-0114-4de6-914b-41a11fc8a1a8", {
+ *    query:{ 
+ *       limit: 1, 
+ *       offset: 20 
+ *    },
+ *     baseUrl: "http://msf-ocg-openmrs3-dev.westeurope.cloudapp.azure.com/openmrs"
```
josephjclark, line 23: Again, the baseURL should be on config, not passed into the option.

To be clear, the private  underlying helper function `util.request should have a baseUrl. But the user-facing `http.request()` operation should not.

packages/openmrs/src/http.js
```diff
@@ -0,0 +1,46 @@
...
+ * });
+ * @function
+ * @public
+ * @param {string} method - HTTP method to use
+ * @param {string} path - Path to resource
+ * @param {OpenMRSOptions}  [options={}] - An object containing either query, headers, and body for the request
```
josephjclark, line 29: Please remove the word "either", because it implies that only one of query, headers or body can be set

packages/openmrs/test/index.test.js
```diff
@@ -63,6 +64,71 @@ describe('execute', () => {
...
+        {
+          results: [{ display: 'Sarah 2' }],
+        },
+        { ...jsonHeaders }
+      );
+
+    const state = { configuration };
+    const { data } = await http.request('GET', '/ws/rest/v1/patient', {
+      query: { q: 'Sarah', limit: 1 },
+      baseUrl: 'https://fn.openmrs.org',
```
josephjclark, line 125: In these tests, baseUrl should be `instanceUrl` on `state.configuration`

josephjclark (changes requested):
Just a few adjustments please :)

.changeset/odd-bobcats-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-openmrs': minor
+---
+
+Implement namespaced http generic helpers
```
josephjclark, line 5: Can we be more accurate here? What we've actually done is add  a new `http.request()` operation. I would not call that a `helper`.

It is probably worth ALSO adding a helpful note to explain the difference between `fetch()` and `get()` (the difference being that `fetch` requests against the instanceURL directly and does not append anything to the path).

josephjclark (approved):

PR #929, opened 2025-01-21
Fix encode, decode and uuid docs examples

josephjclark:
## Summary

Namespaces are missing in these examples!

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


packages/common/src/util/base64.js
```diff
@@ -40,12 +40,12 @@ export const encode = (data, options = { parseJson: true }) => {
  * @param {object} options - Options.
```
mtuchi, line 40: Can we add what options we support, Something like this i guess 
```
* @param {boolean} [options.parseJson=true] - Whether to parse the ended string as JSON.
```

packages/common/src/util/base64.js
```diff
@@ -9,13 +9,13 @@ import _ from 'lodash';
...
```
mtuchi, line 9:
```
* @param {boolean} [options.parseJson=true] - Whether to parse the encoded string as JSON.
```

mtuchi (commented):
The docs looks good @josephjclark 
Just a tiny suggestion, can we document the supported option. 
Something like this 
```js
* @param {boolean} [options.parseJson=true] - Whether to parse the decoded string as JSON.
```

If your happy with the language i can push the changes on this branch

packages/common/src/util/base64.js
josephjclark: Agreed - thanks!

josephjclark:
I am patching common but not updating other adaptors. I think this will be OK

josephjclark:
hmm. That's caused a weird package lock diff because common is unlinked. Not sure I understand this.

Since this is a docs only change, I'll revert and remove the version bump.

PR #930, opened 2025-01-21
Gmail adaptor - Added reasonable result limits

### LINKED ISSUE #928: gmail: add a limit option to `getContentsFromMessages`
In the gmail adapators `getContentsFromMessages`, we need to add an option to limit the amount of emails downloaded. This probably also implies we need an `offset` or `start` key too.

Basically this is a pagination API.

We need to check how to handle this with the google client, and then add a couple of keys to the options like:
- `limit`: set the maximum amount of emails that will be downloaded
- `offset`: ignore emails up to this offset, and start downloading from here

decarteret:
## Summary

Added options.maxResults field to limit excessive data transfer.

Fixes #928

## Details

With options.query being optional, it's possible to (inadvertently) & unnecessarily process and download 10ks of messages. Now we have a default limit of 50 and a hard limit of 1000. Updated README.md with supporting documentation and an example.

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/fifty-mails-melt.md
+++ b/.changeset/fifty-mails-melt.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-gmail': minor
+---
+
+Added options.maxResults field to limit excessive data transfer.
--- a/.changeset/fifty-mails-melt.md
+++ b/.changeset/fifty-mails-melt.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-gmail': minor
----
-
-Added options.maxResults field to limit excessive data transfer.
--- a/packages/gmail/CHANGELOG.md
+++ b/packages/gmail/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @openfn/language-gmail
 
+## 1.1.0
+
+### Minor Changes
+
+- 8203e90: Added options.maxResults field to limit excessive data transfer.
+
 ## 1.0.1
 
 ### Patch Changes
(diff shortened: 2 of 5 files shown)


josephjclark:
So if I understand correctly from the readme and code, the maxResults thing is just the max _processed_ items? So if I do a query and download 1000 items, then do the same query again and pass my processedIds back, I'll retrieve the _next_ 1000 items. That's pretty neat.

My questions are:
1)  why the hard limit? Shouldn't users be able to download a million messages if they want? Maybe they're small messages. Also we stream stuff down so we're relatively memory efficient
2) 50 as a default limit feels really low. What brought us to this figure?

josephjclark:
> Would the job fail gracefully with some type of "out of memory" exception?

Yep! The runtime handles memory limits reasonably well, subject to payment plans and stuff.  The resulting state might not be returned though, so processed Ids may be lost. I think there's actually an outstanding bug report for that - if a job fails, we should return the last known state.

For a 3gb download the two things to watch out for would be time AND memory.

As we're streaming, if you're doing some data transformation/aggregation, the memory overhead for all those emails may not actually be too bad. Like if you're just incrementing values on a prior dataset, memory will be fine. Timeout might catch you out.

So I really don't think we should set a hard limit. Maybe set a default max of 1000, and let users pass `Infinity` if they want to disable it (at their own risk)

josephjclark (approved):
Thanks!

PR #939, opened 2025-01-24
Add `asBase64` option in dhis2 get() function

### LINKED ISSUE #884: dhis2: allow to download an image binary as a base64 string
We have a requirement for the Ghana national ID project to upload an image as a base64 string.

The image in question will come from DHIS2 and be returned as a binary steam.

We need to extend the dhis2 adaptor so that we can convert a binary stream into a base64 string.

I think this means an option on the `get` helper, something like `asBase64`, which would take whatever response we get from dhis2 and convert it into a base64 string.

Note that this will return the whole string, not a stream.

## A complication on large images and payloads

 In order to pass the image onto the next step (ie, to be uploaded in to the NID generator), we have to write the string to the state object and send it back lightning as the output for the step. It could be a big image and could blow memory limits for worker-lightning payloads.

Two possible solutions to this are:
a) work out some way to pass a stream directly between steps (meaning the stream does not get sent back to lightning and the step is not repeatable)
b) download the image from dhis2 straight into the National ID step using a vanilla HTTP helper (in which case this is not a dhis2 issue)

In both approaches, basically we pass the unprocessed binary stream from dhis2 straight into the NID step, then base64 encode on the fly and pass the encoded stream straight into the request (a stream pipeline  -I think this is possible but I don't actually know a lot about the base64 encoding process - can we encode on the fly in chunks? Surely yes?)

Saying all that I _think_ that serializing the image to a string on state is feasible and appropriate

## Credentials
Play DHIS2 instance: https://lmis.integration.dhis2.org/sandbox/
Web User: demo Password: Stock123!

## Testing Guidance
If you log in via the web UI and go to this `trackedEntityInstance`, you will find that it has an image (see TEI attribute `Equipment Image`) that you can test downloading: https://lmis.integration.dhis2.org/sandbox/dhis-web-tracker-capture/index.html#/dashboard?tei=qHVDKszQmdx&program=GqijeO4NM3E&ou=FV43JisquSm

You can try to get this^ record by sending a GET request for TEI `qHVDKszQmdx`. In the response, search for the Attribute ID for `Equipement image`: `BqaEWTBG3RB`

mtuchi:
## Summary

Add `asBase64` option in `get()` function which will convert the `responseType` option to `arraybuffer` and when we get the result we convert the result into a base64 encoded string 

Fixes #884 

## Details

- Update `axios.interceptors.response`  to ignore `arraybuffer` response type
- Add `asBase64` option in `get()`
- Update `jsdocs` for `get()` options argument 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/dhis2/src/Adaptor.js
+++ b/packages/dhis2/src/Adaptor.js
@@ -1,5 +1,5 @@
 import axios from 'axios';
-import { execute as commonExecute } from '@openfn/language-common';
+import { execute as commonExecute, util } from '@openfn/language-common';
 import { expandReferences } from '@openfn/language-common/util';
 import _ from 'lodash';
 const { indexOf } = _;
@@ -11,6 +11,7 @@ import {
   selectId,
   shouldUseNewTracker,
   ensureArray,
+  encode,
 } from './Utils';
 import { request } from './Client';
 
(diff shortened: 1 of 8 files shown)


packages/dhis2/src/Adaptor.js
```diff
@@ -497,25 +503,29 @@ export function update(
...
+      params: {
+        ...resolvedQuery,
+        ...params,
+      },
+      responseType: asBase64 ? 'arraybuffer' : 'json',
       ...requestConfig,
     }).then(result => {
       console.log(`Retrieved ${resolvedResourceType}`);
-      return handleResponse(result, state, callback);
+      const response = asBase64 ? { data: encode(result.data) } : result;
```
josephjclark, line 527: This sent me on a bit of a rabbit hole. I think it will be really healthy to rebsae dhis2 on unidic and make it much more standardised

PR #942, opened 2025-01-26
remove SMPP adaptor

taylordowns2000:
Hi all. I think that this adaptor is for working with the old OpenFn v1 SMPP Client. There was this crazy Erlang SMPP client implementation we did back in the day to support some work in Guatemala. It's no longer around. Is this a generic SMPP adaptor that would work with lots of different SMPP APIs? Or is it something that was specific to an OpenFn implementation, making our SMPP client accessible via REST API?

(I think the latter.)


mtuchi:
@josephjclark whats the process for removing adaptors ?
Do we deprecate the adaptor on npm and remove it from monorepo ?

josephjclark:
@mtuchi yeah exactly that. Delete it from the repo and deprecate on npm

josephjclark (approved):

PR #951, opened 2025-01-28
BDR and NIA adjustments

taylordowns2000:
## Summary

This PR introduces a mock-validator which is used by `ghana-bdr` and `ghana-nia` to verify that the request bodies sent to a specific endpoint match the requirements for those systems.

I think that the improved mocks for these adaptors are not particularly interesting for you guys. The thing that I'd like your thoughts on is whether or not the mock-validator is in the right place. I can see myself using this type of simple helper for building mock servers a lot.

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


josephjclark:
@taylordowns2000  not in tools please! There is no runtime code in tools.

I agree that a general validator seems useful, although I'd probably recommend using json schema (we have a function for that in common, see for reference). But for now, I'd just bake it into the adaptor that needs it and we'll work out how to generalise later.

josephjclark:
@taylordowns2000 we need to fix or skip that failing test before merging

Don't  feel you need my approval before updating these adaptors btw (unless you're doing something unusual, like digging into `tools/`!

tools/mock-validator.js
mtuchi, line 1: Move this to mocks.js file in their respective adaptors

PR #963, opened 2025-01-30
Refactored the OpenMRS test suite

PiusKariuki:
## Summary

Refactor the OpenMRS test suite

Fixes #

## Details

Removing duplicates and storing interceptors in beforeEach blocks 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/openmrs/test/index.test.js
+++ b/packages/openmrs/test/index.test.js
@@ -28,6 +28,12 @@ const configuration = {
   instanceUrl: 'https://fn.openmrs.org',
 };
 
+const state = {
+  configuration,
+  patient: testData.patient,
+  encounter: testData.encounter
+};
+
 describe('execute', () => {
   it('executes each operation in sequence', done => {
     let state = { configuration };
@@ -74,7 +80,6 @@ describe('http', () => {
         method: 'GET',
       })
       .reply(200, { results: [{ display: 'Sarah 1' }] }, { ...jsonHeaders });
-    const state = { configuration };
 
     const { data } = await http.request('GET', '/ws/rest/v1/patient', {
       query: { q: 'Sarah 1' },
@@ -118,7 +123,6 @@ describe('http', () => {
         { ...jsonHeaders }
       );
 
-    const state = { configuration };
     const { data } = await http.request('GET', '/ws/rest/v1/patient', {
       query: { q: 'Sarah', limit: 1 },
     })(state);
@@ -161,8 +165,6 @@ describe('http.get', () => {
       );
   });
 
-  const state = { configuration };
-
   it('should make http request with the "GET" verb', async () => {
     const response = await http.get('/ws/rest/v1/patient')(state);
     expect(response.response.method).to.eql('GET');
@@ -204,8 +206,6 @@ describe('http.post', () =>{
       .reply(404, { ...jsonHeaders });
   });
 
-  const state = { configuration };
-
   it('should make http request with the "POST" verb', async () => {
     const response = await http.post('/ws/rest/v1/patient', testData.newPatient)(state);
     expect(response.response.method).to.eql('POST');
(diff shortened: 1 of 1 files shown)


packages/openmrs/test/index.test.js
```diff
@@ -66,65 +66,43 @@ describe('execute', () => {
 });
 
 describe('http', () => {
-  it('should GET with a query', async () => {
+  beforeEach(()=> {
```
josephjclark, line 69: hmm, I think there are two problems with this structure :thinking: 

1) The interceptor is quite far removed from the test. Like, physically. When I'm looking at the test, it's hard to understand how the mock endpoint will behave.
2) Each of these interceptors only triggers once. So they're literally a 1;1 binding to the test that uses them. And again, they're quite far removed. You could use `.persist()` to make the interceptor permanent (until we tear it down). But that's only helpful if we have multiple tests using the same endpoint.

So I'm really sorry but I'm not convinced that this is a positive change.

If I was to radically think about how to re-do these unit tests (and I'm really not sure I want to invest time in this), I would probably build a mock server and add an API to add data to it. So rather than mocking out very specific endpoints, I could just add or remove data to the server to get the right behaviour out of it. But I do think that's way too much work for this adaptor for now.

packages/openmrs/test/index.test.js
```diff
@@ -66,65 +66,43 @@ describe('execute', () => {
...
-      query: { q: 'Sarah 1' },
-    })(state);
-
-    expect(data.results[0].display).to.eql('Sarah 1');
-  });
+        method: 'DELETE'
+    })
+      .reply(200, { results: 'Successful' }, {...jsonHeaders});
 
-  it('should auto-fetch patients with a limit', async () => {
```
josephjclark, line 86: Have we dropped this test entirely? Why? It's testing some quite important behaviour

packages/openmrs/test/index.test.js
```diff
@@ -321,115 +291,49 @@ describe('fhir', () => {
...
-        },
-        { ...jsonHeaders }
-      );
-
-    const state = { configuration };
-    const { body } = await request(state, 'GET', '/ws/rest/v1/patient', {
-      query: { q: 'Sarah', limit: 1 },
-      baseUrl: state.configuration.instanceUrl,
-    });
+      .reply(200, { results: testData.patientResults }, { ...jsonHeaders });
```
josephjclark, line 312: I do approve of re-using the same test data - that makes the tests much more concise and readable

packages/openmrs/test/index.test.js
```diff
@@ -321,115 +291,49 @@ describe('fhir', () => {
...
-      })
-      .reply(
-        200,
-        {
-          results: [{ display: 'Sarah 2' }],
-        },
-        { ...jsonHeaders }
-      );
+  // Define state only once
+  const state = { configuration }
```
josephjclark, line 321: Yep - we could probably do this once at the top of the file couldn't we? Because actually, despite what this comment says, by my count we define state 14 times :sweat_smile:

packages/openmrs/test/index.test.js
```diff
@@ -321,115 +291,49 @@ describe('fhir', () => {
...
-    expect(body.results[0].display).to.eql('Sarah 1');
-    expect(body.results[1].display).to.eql('Sarah 2');
+  afterEach(() => {
+    // Remove all interceptors after each test
+    testServer.interceptors = [];
   });
 
-  it('should not auto-fetch if the user sets startIdex', async () => {
-    testServer
-      .intercept({
```
josephjclark, line 389: I'm concerned we've lost some important test logic here too

josephjclark (changes requested):
Hi @PiusKariuki, thank you for this.

I'm really sorry but I'm not sold on  some of the changes. As a rule I would prefer tests to be verbose but readable rather than super efficient. I think some of these refactors make the tests harder to read and understand and extend. I also think we've lost some important test logic?

I don't want to spend a lot of time on this going forward. If you can quickly and easily keep some of the smaller changes - the state and data re-use for example - then do it. Otherwise I think I have to close this down. No more than an hour of dev from this point please.

PiusKariuki:
Hi @josephjclark I totally understand this let me wrap this up quickly. I will revert  the commit and just keep the state and data re-use logic.

Issue #955, opened 2025-01-29
fhir4: things I'm breaking


josephjclark:
I am rebuilding the fhir adaptor generator to specialise in fhir4.

This is making me break regular fhir generation from an IG. For a couple of reasons:
- The base fhir spec has a different format to the IG. I don't _really_ understand this - surely fhir is fhir? but the Ig and the base spec are packaged very differently
- A base fhir adaptor has a few different assumptions about where types come from

As a result, I'm introducing the following problems which will all need fixing before we can generate from an IG:
- signatures now support a simple format `b.patient(props)` which uses a default profile, but also you can specify a profile `b.patient'inbound-patient', props)`. 
- The typings I've done in these builders assume only a single profile. For multiple profile support, I need to do some fancier type stuff. Basically I need to union all the variants and also lookup the matching type from the string. I've done this in the IG already, I just need to rebuild that stuff into the generator
- more I'm sure??

Also, a quick reminder, we need to do further work once all this i done:
- When generating from an IG, we need a "base" adaptor (defaulting to r4). We should:
  - import data types from the base
  - In the builders, call the base resource buillder first, and then only apply diffs from the differential in the spec.

josephjclark:
I need to spin out mappings and extensions to different issues.

It's taking a long time to get fhir working in typescript. I'm close now and want to get that released.

data values would be incredibly useful - for example, to provide completions on name usage or gender. All the data is available to do this.

Even more useful is shortcuts for complex values -I don't have a good example to hand but eg the simple string `M` can expand to `{ value: 'male', code: 'blahblah' }`. Again, we have all the data fhir to do this.

Even better would be generic value mappers. So I could do something in job code to auto-map vales. Like we provide a function once to map incoming name objects to FHIR name objects. Now I can pass an upstream name object straight into the builder and let it automap. IG guides can't do this because they only provide one side of the mapping. But I'm sure we should be able to allow jobs to define mapping rules (or helper functions).

The other area is extensions, which need a bit more work. I would like builders to use a key like `religion` and use that to auto-map the incoming value into the correct extension structure. Job code then just has to provide the value (and maybe we can take advantage of value maps and custom mapping to allow this to be properly converted)

josephjclark:
One thing that doesn't work so well is the typedproperties like `deceased[x]`, which can expand to `deceasedBoolean` and `deceasedDateTime`. the builder works great, but docs don't really work

PR #964, opened 2025-01-30
Fhir 4 typescript

josephjclark:
FHIR r4 adaptor with auto-generated typescript.

Replaces #909

This is a fully typescript implementation, so we get really good type definitions out of it, which gives us great code-assist support.

I am not using the npm `fhir` package for typings. It just doesn't work very well for us, and introduces a problem bundling for Monaco. Also it exposes `_` variants of each property, for extensions. Which really pollutes the code assist.

Code assist is now provided in test files (which makes writing unit tests easier AND more lightning-like) as well as in Lightning. I've tested it locally, works great.

Major changes:
- Update jsdocs2md to support typescript docs. I think this was necessary but doesn't introduce any changes to existing adaptors
- This would be our first typescript-first adaptor. Cool.
- I have condensed the generated code to be more terse and permissive. Whatever the user passes in is spread into the result. So you can do whatever you like
- The builder signature now overloaded. You can do `b.patient('Patient', { ...})` or `b.patient({ ... })`. Profile names aren't actually needed or useful here, but I want to keep API consistency with the IG adaptors. It works well.

Still do to before release:
- Add basic operations with fhir verbs - http, read, create, update, bundle
- I think I'll absorb the `datatypes` stuff into builders, so you can do `b.concept`, `b.identifier`.
- ~~Properly extract meta and text properties from the generator, and let uses provide extension hooks for this in the mappings file~~ DONE
- Manually write out a few tests for key types. This helps validates improvements
- Add a good docs overview and some examples to showcase best practice
- Fix docs for composite keys like `deceased[x]`.
- Should we be a lot more restrictive in the builders we support? We include almost everything but maybe we should only include the medical stuff

Also note that this totally breaks the implementation guide generator. I'm tracking that in #955

Spin out issues to do later:

- Parse valuesets to allow auto-complete for basic types (like human name enums)
- Use valuesets to map simple string values to system/coded concepts
- Better support for extensions
- Allow job code to set mapping values. So that I can pass any arbitrary object to, eg, patient.identifier, and the mapping function will handle the mapping. Essentially we allow custom support for value mapping. This should make mapping to fhir MUCH easier

## AI Disclaimer

I used a little Claude wisdom to help work out how to handle some of the typings.


josephjclark:
I'm SO CLOSE to having the typescript build working. But hitting two big issues this afternoon:
1)  The generated d.ts files are all bundled into the global scope. so the types in lightning are a bit misleading. I hope I can just add some namespace to fix this :crossed_fingers: 
2) jsdocs2md are refusing to parse the ts. This is a bit weird because typescript file are all supported, I've been through the guide and set it up. I should be able to crack this but I'm a little bit stumped....

josephjclark:
I'm a little bit disappointed by docs, but I think this is a problem for later.

Here's how generated docs look: 
image

Issues:
- It's a really, really big page and quite slow to load
- There are no usage examples
- there's a LOT of repetition. Stuff like id, meta, text each written to each of the, uh, hundreds of generated types
- Types like HumanName aren't very useful. I should probably include those typedefs. Code assist should compensate but that doesn't help CLI users
- it's not at all clear when things can be arrays. Quite a lot of things can be, but we don't say so

image


Solutions?
- Put builders in a different page in the docs. Doesn't really help the size, but maybe stops the operations drowning
- Add some common typedefs and offload keys into them, so stuff like meta, id doesn't appear in every table
- Document allowed properties a lot more tersely. Probably doens't help.

josephjclark:
Spun out a nice solution for docs in https://github.com/OpenFn/adaptors/issues/987 . I don't think it's a big fix and I'll probably take a swing real soon.

josephjclark:
I've just been doing my final flight checks, getting ready for release and running some extra tests.

And I've hit the weirdest issue. It's taken a while to whittle it down to barebones.

If I make `index.d.ts` say this:
```js
type Identifier = {
  foo: 'string';
  bar: 'string';
};
declare const identifier: (id: Identifier) => string;

export { identifier };
```

Everything is fine, I get code assist in monaco for the identifier.

if I change the signature to this:
```
declare const identifier: (id: string | Identifier) => string;
```

I get no type assist at all.

It's like adding the string union blows it up. A union like `'foo' | 'bar'` works great.

Works just fine in the monaco playground for 0.43 (the version we're running)

I've disabled the magic code completion stuff. Still breaks.

josephjclark:
I think I'm gonna have to cut out these `string | Identifier` shorthand types and call this a Lightning bug. Very frustrating. We have a parallel work stream in this area so maybe it can get looked at there.

josephjclark:
Ok, incredibly, what we have to do is add this in the es5 language definition in Monaco:
```
interface Array<T> {
    copyWithin(target: number, start: number, end?: number): this;
    [n: number]: T;
}
```

I don't understand it at all - but the type tracking totally breaks down when typings for `Array.copyWithin()` aren't present.

Raising a PR against Lightning now.

josephjclark:
Right, everything is working, the docs build is fine, docs are updated. I think I'm gonna release this as a 0.1.0 :crossed_fingers:

josephjclark:
Oh, note that we're waiting for a PR on the lightning side before code assist works properly in the platform. Sad but true.

PR #976, opened 2025-02-05
Googlesheets expose the client directly

### LINKED ISSUE #575: `googlesheets`: expose the client directly
Googlesheets provides a huge API and our adaptor is not likely to ever map all of it.

But in various use-cases, users may want to use the API directly.

In the spirit of exposing `get` and `post` helpers for HTTP APIs, I propose we expose the googlesheets client directly:
```
useSheets((state, sheetsClient) => {
    // do whatever you want with the client
    const response = await sheetsClient.values.batchUpdate({
        spreadsheetId: 'abc',
        resource:  { data: [{ range, values }]
     });

  // return a new state object
   return { ...state, data: response.data }
})
```

We could call the function any of these:
* `useSheets()`
* `sheets()`
* `getClient()`
* `client()`

Actually, we could even override `fn()` - that's kinda neat, because that's really all we're doing. That could establish a nice pattern to other adaptors.

```
fn((state, sheets) => {
    // do whatever you want with the client
  
  // return a new state object
   return { ...state, data: response.data }
})
```

## Spec

Here's how the function would work:

* State and the googlesheets client are passed into the function
* The function uses regular JS, async await, whatever you like
* The function must return state - just like `fn` - and that state will be passed downstream

## Security

Are there security concerns with exposing the client directly? Like can a user read credentials back out of it?

PiusKariuki:
## Summary

Expose `googlesheets` client for more flexible job writing

Fixes #575

## Details

Override the common `fn` function that takes the `state` and the `sheetsClient` as arguments in the callback

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/googlesheets/src/Adaptor.js
+++ b/packages/googlesheets/src/Adaptor.js
@@ -252,8 +252,8 @@ export function getValues(spreadsheetId, range, callback = s => s) {
  * @returns {Operation}
  */
 export function fn(func) {
-  return (state, sheetsClient) => {
-    return func(state, sheetsClient)
+  return state => {
+    return func(state, client)
   }
 }
 
--- a/packages/googlesheets/src/Adaptor.js
+++ b/packages/googlesheets/src/Adaptor.js
@@ -18,6 +18,7 @@ function createConnection(state) {
   auth.credentials = { access_token: accessToken };
 
   client = google.sheets({ version: 'v4', auth });
+  console.log(' >> created client');
   return state;
 }
 
@@ -253,7 +254,7 @@ export function getValues(spreadsheetId, range, callback = s => s) {
  */
 export function fn(func) {
   return state => {
-    return func(state, client?.sheets);
+    return func(state, client?.spreadsheets);
   };
 }
 
--- a/packages/googlesheets/src/Adaptor.js
+++ b/packages/googlesheets/src/Adaptor.js
@@ -233,6 +233,30 @@ export function getValues(spreadsheetId, range, callback = s => s) {
   };
 }
 
+
+/**
+ * Exposes the googlesheets client for more flexible job writing
+ * @function
+ * @public
+ * @example
+ * fn((state, sheets) => {
+ *  // make any request with the client
+ *   const response  = await sheetsClient.values.batchUpdate({
+ *     spreadsheetId: 'abc',
+ *     resource: {data: [{ range, values }]}
+ *   });
+ *   // return a new state object
+ *   return {...state, data: response.data }
+ * }
+ * @param {Function} func is the function
+ * @returns {Operation}
+ */
+export function fn(func) {
+  return (state, sheetsClient) => {
+    return func(state, sheetsClient)
+  }
+}
+
 export {
   alterState,
   combine,
@@ -242,7 +266,6 @@ export {
   each,
   field,
   fields,
-  fn,
   fnIf,
   http,
   lastReferenceValue,
--- a/packages/googlesheets/src/Adaptor.js
+++ b/packages/googlesheets/src/Adaptor.js
@@ -233,28 +233,28 @@ export function getValues(spreadsheetId, range, callback = s => s) {
   };
 }
 
-
 /**
- * Exposes the googlesheets client for more flexible job writing
+ * Execute an arbitrary function against state. The Google Sheets adaptor
+ * also exposes a {@link https://googleapis.dev/nodejs/googleapis/latest/sheets/classes/Resource$Spreadsheets.html | googlesheets client} instance.
  * @function
  * @public
- * @example
+ * @example <caption>Use the Sheets client directly</caption>
  * fn((state, sheets) => {
- *  // make any request with the client
- *   const response  = await sheetsClient.values.batchUpdate({
+ *   // make any request with the sheets client
+ *   const response  = await sheets.values.batchUpdate({
  *     spreadsheetId: 'abc',
- *     resource: {data: [{ range, values }]}
+ *     resource: {data: [{ range: state.range, values: state.values }]}
  *   });
  *   // return a new state object
  *   return {...state, data: response.data }
  * }
- * @param {Function} func is the function
+ * @param {Function} func The function to execute. Must return state.
  * @returns {Operation}
  */
 export function fn(func) {
   return state => {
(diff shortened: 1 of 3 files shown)


packages/googlesheets/src/Adaptor.js
```diff
@@ -233,6 +233,30 @@ export function getValues(spreadsheetId, range, callback = s => s) {
...
+ *     resource: {data: [{ range, values }]}
+ *   });
+ *   // return a new state object
+ *   return {...state, data: response.data }
+ * }
+ * @param {Function} func is the function
+ * @returns {Operation}
+ */
+export function fn(func) {
+  return (state, sheetsClient) => {
```
josephjclark, line 255: This doesn't look right to me :eyes: 

The sheetsClient is bound the local scope called `client`. It's created up on line 20.

We want to pass that thing into the user's function, along with state. So more like this:
```
export function fn(func) {
  return (state) => {
    return func(state, client)
  }
}
```

packages/googlesheets/src/Adaptor.js
```diff
@@ -233,6 +233,30 @@ export function getValues(spreadsheetId, range, callback = s => s) {
   };
 }
 
+
+/**
+ * Exposes the googlesheets client for more flexible job writing
```
josephjclark, line 238: I hadn't realised our docs for `fn()` were so bad :see_no_evil: 

When we're done I'll do a pass on these docs (then raise an issue to fix it in common)

packages/googlesheets/test/index.js
```diff
@@ -68,3 +68,48 @@ describe('batchUpdateValues', () =>{
     expect(result).to.eql(state);
   });
 });
+
+describe('fn function', () => {
```
josephjclark, line 72: these mock tests don't look right either :thinking: 

You'll need to call `execute` to trigger the creation of the sheets client, and then call `fn(callback)`. Your callback needs to ensure that a sheets client is passe as the second argument (by the adaptor, not by your test code)

packages/googlesheets/src/Adaptor.js
```diff
@@ -233,6 +233,30 @@ export function getValues(spreadsheetId, range, callback = s => s) {
...
+ * @function
+ * @public
+ * @example
+ * fn((state, sheets) => {
```
josephjclark, line 242: This example is actually correct! I just don't think the implementation honours it correctly.

josephjclark (changes requested):
Thank you for looking at this @PiusKariuki! I don't think the implementation looks quite right though

packages/googlesheets/src/Adaptor.js
PiusKariuki: Yeah I was sure I has missed something. This clarifies it. Thanks @josephjclark !. I'll make a commit to rectify

josephjclark:
So I've tweaked the docs and, on reflection, made a small adjustment to the implementation (pass `client.spreadsheets` because that's the bit we want, and the docs for `client` itself are utterly useless!)

But I've hit a security problem - something I was worried about. If you console.log the client then the access token becomes visible:

image

This is a security issue. A malicious actor who can use but not see the access token could gain access to it. Maybe a bit of a long shot  but I'm uncomfortable with this.

I'll have to sleep on it!

PiusKariuki:
Oh yeah that's no good. Best we think on it see if we can find a workaround.

packages/googlesheets/src/Adaptor.js
```diff
@@ -18,6 +18,7 @@ function createConnection(state) {
   auth.credentials = { access_token: accessToken };
 
   client = google.sheets({ version: 'v4', auth });
+  console.log(' >> created client');
```
josephjclark, line 21: Oh this will need removing, that's my sloppy debug code

josephjclark:
@PiusKariuki can you investigate this a bit for me? Please don't give it more than 3 hours. I was hoping this would be an easy thing really.

You'll need be able to run a simple googlesheets job locally with the CLI, for which you'll need an oauth `access_token`. There are some good [docs for this on the gmail adaptor](https://docs.openfn.org/adaptors/packages/gmail-readme#acquiring-an-access-token) but I've never actually done it myself. Mtuchi may be able to help too.

Anyway you'll need to get a simple job working that can read from a google sheet of your creation.

Then you need to try and wrap the sheets client in a way that can't be accessed from job code. What I hope we can do is block off access externally, in job code, but still allow the client to read the credential to do its own thing.

You might be able to do this by wrapping the sheets client with a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) and denying access to `context._options` from outside. The wrapper would present context._options as undefined but hopefully inside the proxy the client can still get at the auth details.

PiusKariuki:
@josephjclark these are some great ideas to start off of👏. I'll explore them and hopefully we get our solution.

josephjclark:
Just a heads-up that we are still thinking about this :)

PiusKariuki:
Yeah I've been trying to see if I can find a way around the security issue too.

PiusKariuki:
@josephjclark wWhat if we had a handler function that took arguments  and then used them to carry out the request. See signature below

`handler(method, arguments, callback)`

Examples:
1.  `handler("get", {spreadsheetId: 'abc', range: 'Sheet1!A1:D10', someCallback})`
2. `handler("get", {spreadsheetId: 'abc', resource: {
    data: [{range: 'Sheet1!A1:E1, values: [
    ['From expression', '$15', '2', '3/15/2016'],
    ['Really now!', '$100', '1', '3/20/2016']
]}]
}, someCallback})`

josephjclark:
Hey @PiusKariuki that's a really interesting idea. It's a bit of a messy API but it would provide an escape hatch and would solve the problem :thinking: 

On a similar line I suppose we could make our own sort of manual proxy and wrap the client API. But that sounds hard to do and to maintain.

Do you have a branch with your proxy experiment in it? Can I take a look?

PiusKariuki:
I might have the changes stashed somewhere. I'll retrieve them then I'll send the link to the branch .

josephjclark:
@PiusKariuki Don't worry if it's lost - just a thought :)

PR #982, opened 2025-02-05
kobotoolbox: implement automatic pagination

### LINKED ISSUE #943: Kobotoolbox: New API signatures
For background see #603

# New API

Optimised for getting form submisisons
```
getForms() // no options
geSubmissions(formId, options) // options is { query, limit, start }
getDeployment(formId) // no options
```

Notes:
- getSubmissions supports `{start, limit, query}`, see `https://eu.kobotoolbox.org/api/v2/assets/<uuid>/data/`
- make sure to include `?format=json` in the query URLs. The accept header might work too?
- getSubmissions calls /assets/<formId>/data. Technically you can pass any asset id and it'll get the data.
- getForms  MUST add the asset_type filter: `?q=asset_type:survey`.  
- rename params to options
- remove callbacks
- See #603 for docs comments - lets make sure the docs are nice and clean

# Old API

For the record the current adaptor is not so different after all:
```
getForms(params, callback) // gets all asset
getSubmissions(params, callback) gets submissions for a given asset (form) - ie asset/id/data
getDeploymentInfo(params, callback) gets deployment info, ie asset/id/submission
```
# Pagination

pagination is really hard in openfn if the adpator  doesn't support it. We provide zero looping capability. 

So we should automatically handle pagination to handle all data, unless `offset` or `limit` is provided (in which case we assume users will handle their own pagination)

Image

hunterachieng:
## Summary

Implement automatic pagination in `kobotoolbox`

Fixes #943 

## Details

`kobotoolbox` API supports pagination with `limit` and `start`. We are implementing an automatic pagination to support these items and return all data where applicaple.

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/kobotoolbox/src/Adaptor.js
+++ b/packages/kobotoolbox/src/Adaptor.js
@@ -59,11 +59,12 @@ export function getForms() {
  * Get submissions for a specific form
  * @example <caption>Get all submissions for a specific form</caption>
  * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
- * @example <caption>Get submissions with exactly 10 items. Equivalent to `<baseUrl>/assets/?offset=0&limit=10`</caption>
+ * @example <caption>Get submissions with exactly 10 items. Equivalent to `<baseURL>/api/v2/assets/?offset=0&limit=10`</caption>
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { limit: 10 });
  * @function
  * @public
  * @param {string} formId - Form Id to get the specific submissions
- * @param {object} [options={}] - Optional query params, headers, limit, and start, for the request
+ * @param {object} [options={}] - Optional query params; limit and start for the request
  * @returns {Operation}
  */
 export function getSubmissions(formId, options = {}) {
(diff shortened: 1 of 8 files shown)


josephjclark:
@hunter is the title and description of this PR up to date? I think there's a lot more going on here than just pagination?

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -41,78 +41,83 @@ export function execute(...operations) {
...
- *   console.log(state.data);
- *   return state;
- * });
+ * getDeploymentInfo('aXecHjmbATuF6iGFmvBLBX');
  * @function
  * @public
- * @param {object} params - Form Id and data to make the fetch or filter
- * @param {function} callback - (Optional) Callback function to execute after fetching form deployment information
+ * @param {string} formId - Form Id to get the deployment information
+ * @param {RequestOptions} [options={}] - Optional query params and headers for the request
```
josephjclark, line 105: I don't think we need any options on deployments? The spec says there should be no options, and looking a kobo docs I still agree that that's correct

packages/kobotoolbox/test/index.js
```diff
@@ -1,14 +1,16 @@
 import chai from 'chai';
 const { expect } = chai;
 import { enableMockClient } from '@openfn/language-common/util';
+import { request } from '../src/Utils';
 
-import nock from 'nock';
+import { http } from '../src';
```
josephjclark, line 6: Can you remove `nock` now and then we can close #984?

From `packages/kobotoolbox` I think it;'s just `pnpm remove nock`, and then check in the diff

josephjclark (commented):
Almost there - just some docs really!

josephjclark:
@hunterachieng I've pushed a changeset and a couple of docs things. But I still don't see `@state` annotations for `state.data` - could you please add them?

hunterachieng:
@josephjclark  I have added the annotations

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -3,6 +3,14 @@ import { expandReferences } from '@openfn/language-common/util';
 
 import * as util from './Utils';
 
+/**
+ * State object
+ * @typedef {Object} KoboToolboxState
```
josephjclark, line 8: This doesn't look right - `getForms` shouldn't return a `response` object? Are we doing this? If so it should be removed. And the data isn't really a response body.

I'm expecting eg `getForms` to say something like `@state data - an array of form objects`

hunterachieng: Fixed

josephjclark: Great, thanks!

PR #989, opened 2025-02-10
kobotoolbox: switch baseURL to baseUrl

hunterachieng:
## Summary

Switch from using `baseURL` to `baseUrl` in `kobotoolbox`

Fixes #

## Details

All adaptors use `baseUrl` as opposed to `kobotoolbox` that uses `baseURL`. We have switched to `baseUrl` but still allow users to send requests with `baseURL` but log a warning.

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/kobotoolbox/src/Utils.js
+++ b/packages/kobotoolbox/src/Utils.js
@@ -15,7 +15,15 @@ export const prepareNextState = (state, response) => {
 };
 
 export async function request(state, method, path, opts) {
-  const { baseURL, apiVersion, username, password } = state.configuration;
+  const { baseUrl, apiVersion, username, password } = state.configuration;
+
+  let baseURL = baseUrl || state.configuration.baseURL;
+
+  if (!baseUrl) {
+    console.warn(
+      'No baseUrl found in state.configuration. baseURL will be used instead, but this will be deprecated in the future.'
+    );
+  }
 
   const {
     data = {},
--- a/packages/kobotoolbox/ast.json
+++ b/packages/kobotoolbox/ast.json
@@ -4,7 +4,7 @@
       "name": "getForms",
       "params": [],
       "docs": {
-        "description": "Make a request to fetch all survey forms accessible to the user's API token. The url is `/assets/?asset_type=survey`.",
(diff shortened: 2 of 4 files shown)


packages/kobotoolbox/src/Utils.js
```diff
@@ -15,7 +15,15 @@ export const prepareNextState = (state, response) => {
 };
 
 export async function request(state, method, path, opts) {
-  const { baseURL, apiVersion, username, password } = state.configuration;
+  const { baseUrl, apiVersion, username, password } = state.configuration;
+
+  let baseURL = baseUrl || state.configuration.baseURL;
+
+  if (!baseUrl) {
```
josephjclark, line 22: This won't work because  if baseURL is set, the baseUrl var will have a value and this block won't be entered. Did you test it manually?

I think you should be more explicit and have a whole if statement around state.configuration.baseURL, and do the warning and assignment inside

hunterachieng: I did test it, but let me change

hunterachieng: Fixed

packages/kobotoolbox/src/Utils.js
```diff
@@ -15,7 +15,18 @@ export const prepareNextState = (state, response) => {
 };
 
 export async function request(state, method, path, opts) {
-  const { baseURL, apiVersion, username, password } = state.configuration;
+  const { baseUrl, apiVersion, username, password } = state.configuration;
+
+  let baseURL;
+
+  if (state.configuration.baseURL) {
```
josephjclark, line 22: Ok but this way, if baseURL and baseUrl are set, baseURL will win AND users will get an incorrect warning.

I find it kinda weird too that the code uses the wrong styling of `baseURL`. It would feel cleaner if code and config were consistent and baseURL was the outlier.

So:
`baseUrl` needs to be the main variable
`if (!baseURL && state.configuration.baseURL) { /* warn and assign */ }

josephjclark: Please don't rush your solution! It is very expensive to keep revisiting little fixes like this

josephjclark (approved):
Thank you!

PR #994, opened 2025-02-13
Openmrs rework api phase 3

### LINKED ISSUE #891: Phase 3: Add rest API helpers
See parent issue for details.

PiusKariuki:
## Summary

Clean up support for the OpenMRS REST API

Fixes #891 

## Details

The `get`, `upsert`, `update` and `delete` functions have the general siganture: `verb(path, data?, options?)`

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/openmrs/src/Adaptor.js
```diff
@@ -60,31 +60,33 @@ export function getPatient(uuid, callback = s => s) {
...
+ *    q: "Patient",
  *   limit: 1,
+ * }
  * });
  * @function
  * @public
  * @param {string} path - Path to resource (excluding /ws/rest/v1/)
- * @param {object} query - parameters for the request
- * @param {function} [callback] - Optional callback to handle the response
+ * @param {object} options - parameters for the request
```
josephjclark, line 71: Can we provide a link to the supported parameters here? If you look at our [commcare](https://docs.openfn.org/adaptors/packages/commcare-docs#bulk) docs, we link to the destination  docs to explain what's available.

packages/openmrs/src/Adaptor.js
```diff
@@ -247,9 +249,8 @@ export function getEncounters(query, callback = s => s) {
  * Create a record
  * @public
  * @function
- * @param {string} resourceType - Type of resource to create. E.g. `person`, `patient`, `encounter`, ...
- * @param {OpenMRSData} data - Object which defines data that will be used to create a given instance of resource. To create a single instance of a resource, `data` must be a javascript object, and to create multiple instances of a resources, `data` must be an array of javascript objects.
```
josephjclark, line 251: ooh, just noticed this. Do arrays of data still work to create multiple resources at once?

It looks like we're using the API the same way so presumably arrays are still supported

packages/openmrs/src/Adaptor.js
```diff
@@ -307,29 +308,30 @@ export function getEncounters(query, callback = s => s) {
...
   return async state => {
-    const [resolvedResource, resolvedData] = expandReferences(
+    const [resolvedPath, resolvedData] = expandReferences(
       state,
-      resourceType,
+      path,
       data
     );
-    console.log('Preparing to create', resolvedResource);
+    const resourceName = resolvedPath[0] === '/' ? resolvedPath.split('/')[1] : resolvedPath.split('/')[0];
```
josephjclark, line 318: If we're only using `resourceName` for logging then I would just log it verbatim.

I forget,  does the `request()` helper log the URL and method like our other adaptors do? In which case I'm not sure that this log line is actually useful :thinking: 

This crops up a few times

packages/openmrs/test/index.test.js
```diff
@@ -437,6 +439,31 @@ describe('get', () => {
...
   });
+  it('should be robust to leading and trailing slashes', async () => {
+    testServer
+      .intercept({
+        path: '/ws/rest/v1/encounter/123',
+        method: 'GET',
+      })
+      .reply(200, { uuid: '123' }, { ...jsonHeaders });
+
+    const { data } = await execute(get('/encounter/123/'))(state);
```
josephjclark, line 450: I would like to see this same test repeated for:
```
get('/encounter/123/')
get('encounter/123/')
get('encounter/123')
```

That gives us great coverage. I also like that its independent of the implementation - the test should be the same if we're using `path.join()` or `cleanPath`

josephjclark (changes requested):
Thank you @PiusKariuki! Looks great, the API design seems to be holding up, and the implementation is nice and clean.

Just a few comments, mostly around docs and path parsing

packages/openmrs/src/Utils.js
```diff
@@ -87,3 +87,10 @@ export async function request(state, method, path, options = {}) {
 
   return allResponses;
 }
+
+
+export function cleanPath(path) {
```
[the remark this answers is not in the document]
PiusKariuki: Hi @josephjclark I substituted `node:path` for the custom function but I noticed that it fails to remove trailing slashes at the end of a path which leads to mismatches.

Also if a user makes the mistake of  leaving a trailing slash in a query with params it doesn't remove that slash either.

Example: `get('patient/', {query: {q: 'Sarah'}})` decodes to `'/ws/rest/v1/patient/?q=Sarah'`

packages/openmrs/src/Adaptor.js
PiusKariuki: The `request` does not log the  URL or method. I have logged the path as it is.

packages/openmrs/test/index.test.js
PiusKariuki: Great idea. Again the first two examples will fail to match the intercepted URL since the trailing slash is persisted

packages/openmrs/src/Adaptor.js
josephjclark: Can we test this? I'd like to get clarity.

Looking at the  diff I don't think the actual underling behaviour has changed? if the user passes an array of data, the adaptor will update an array of data, and it'll still work?

Do you have a test environment for this stuff?

packages/openmrs/src/Utils.js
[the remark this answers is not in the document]
josephjclark: huh, well spotted. And thanks for flagging - I appreciate that!

This is annoying and might affect our other adaptors. Luckily I think a trailing slash probably quite a rare case.

Let me look into this a bit more tomorrow. I'm still a bit uncomfortable solving this sort of thing with regexes :thinking:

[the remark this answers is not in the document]
josephjclark: Looks like you've got a good set of tests though!

packages/openmrs/src/Adaptor.js
PiusKariuki: I can just use the OpenMRS dev server for this. I'll write a job and see if arrays are working.

PiusKariuki: Hi @josephjclark I created this job to test this behaviour. See  below ....

```javascript
 fn(state=> {
  state.configuration = {...state.configuration, username: 'admin', password: 'Admin123', instanceUrl: 'https://dev3.openmrs.org/openmrs'};
  return state;
})

create('encountertype', [
  {
    "name": 'Tooth filing v2',
    "description": "Attach encounters related to hospital dischargers"
  },
  {
    "name": "Routine dental exam",
    "description": "Attach encounters related to hospital dischargers"
  }
]);
```


And I am getting this error back

> "message": "[JSON parse error: Cannot deserialize value of type `org.openmrs.module.webservices.rest.SimpleObject` from Array value (token `JsonToken.START_ARRAY`); nested exception is com


The util function is passing the body as an array instead of sending multiple requests of individual items in the array.

josephjclark: Agreed. So even if we were to add a feature that iterated over every item in `data`, there would be not advantage over just doing `each($.data, create($.data)`.

Then this is perfect as it is. Thanks for the help!

josephjclark (approved):
Thank you @PiusKariuki ! :pray:

PR #998, opened 2025-02-17
http: change body signature in functions

### LINKED ISSUE #922: http: It's too hard to POST JSON
It took me like 5 minutes to post a JSON object to a server with HTTP post.

It shouldn't be so hard!

Where I went wrong is:

- The signature is `post(path, params, callback). 
- Params should be called `options` , because 'we not defining URL query params
- The body goes in params.body. Why not `post(path, body, options)` ?
- I have to specify a content type application.json. Why doesn't the adaptor do this for me?

To be fair we do document the content type thing:

Image

I don't remember why we don't auto-parse an object 🤔 Feels like the natural thing to do. We should log to say we're doing it, and handle errors well.

I suggest:

- Rename params to options
- Either auto-detect content type, or just default to json, unless the content-type header is set
- Update the signature when we do our next major version bump

hunterachieng:
## Summary

Change signature on all `http` functions to allow body data sent separately from the `options`

Fixes #922 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/http/src/Adaptor.js
+++ b/packages/http/src/Adaptor.js
@@ -6,7 +6,7 @@ import { request as sendRequest, xmlParser } from './util';
  * @typedef {Object} RequestOptions
  * @public
  * @property {object} errors - Map of errorCodes -> error messages, ie, `{ 404: 'Resource not found;' }`. Pass `false` to suppress errors for this code.
- * @property {boolean} form - True/False if the body data is multipart HTML form (as FormData).
+ * @property {boolean} form - Set to `true` if the body data is multipart HTML form (as FormData).
  * @property {object|string} body - body data to append to the request. JSON will be converted to a string (but a content-type header will not be attached to the request).This is only applicable to the request function
  * @property {object} query - An object of query parameters to be encoded into the URL.
  * @property {object} headers - An object of headers to append to the request.
--- a/packages/http/src/Adaptor.js
+++ b/packages/http/src/Adaptor.js
@@ -7,6 +7,7 @@ import { request as sendRequest, xmlParser } from './util';
  * @public
  * @property {object} errors - Map of errorCodes -> error messages, ie, `{ 404: 'Resource not found;' }`. Pass `false` to suppress errors for this code.
  * @property {boolean} form - True/False if the body data is multipart HTML form (as FormData).
+ * @property {object|string} body - body data to append to the request. JSON will be converted to a string (but a content-type header will not be attached to the request).This is only applicable to the request function
  * @property {object} query - An object of query parameters to be encoded into the URL.
  * @property {object} headers - An object of headers to append to the request.
  * @property {string} parseAs - Parse the response body as json, text or stream. By default will use the response headers.
@@ -51,7 +52,6 @@ export function execute(...operations) {
  * request(
  *   'GET',
  *   '/myEndpoint',
- *   {},
  *    {
  *      query: {foo: 'bar', a: 1},
  *      headers: {'content-type': 'application/json'},
@@ -60,14 +60,13 @@ export function execute(...operations) {
  * @function
  * @param {string} method - The HTTP method to use.
  * @param {string} path - Path to resource. Can be an absolute URL if baseURL is NOT set on `state.configuration`.
- * @param {object} data - Body data to append to the request. JSON will be converted to a string.
- * @param {RequestOptions} options - Query, Headers and Authentication parameters
+ * @param {RequestOptions} options - Body, Query, Headers and Authentication parameters
  * @param {function} callback - (Optional) Callback function
  * @state {HttpState}
  * @returns {Operation}
  */
-export function request(method, path, data, options, callback) {
-  return sendRequest(method, path, { body: data, ...options }, callback);
+export function request(method, path, options, callback) {
+  return sendRequest(method, path, options, callback);
 }
 
 /**
@@ -80,7 +79,7 @@ export function request(method, path, data, options, callback) {
  * })
  * @function
  * @param {string} path - Path to resource. Can be an absolute URL if baseURL is NOT set on `state.configuration`.
- * @param {RequestOptions} Options - Body, Query, Headers and Authentication parameters
+ * @param {RequestOptions} options - Body, Query, Headers and Authentication parameters
  * @param {function} callback - (Optional) Callback function
  * @state {HttpState}
  * @returns {Operation}
@@ -92,9 +91,9 @@ export function get(path, options, callback) {
 /**
  * Make a POST request. If `configuration.baseUrl` is set, paths must be relative.
  * @public
- * @example
+ * @example <caption>Create a resource with data</caption>
  *  post('/myEndpoint',
- *  {'foo': 'bar'},
+ *  $.data
  * {
  *    headers: {'content-type': 'application/json'},
  *  })
@@ -115,7 +114,8 @@ export function post(path, data, options, callback) {
  * Make a PUT request. If `configuration.baseUrl` is set, paths must be relative.
  * @public
  * @example
- *  put('/myEndpoint',  {'foo': 'bar'},{
+ *  put('/myEndpoint', {'foo': 'bar'},
+ * {
  *    headers: {'content-type': 'application/json'},
  *  })
  * @function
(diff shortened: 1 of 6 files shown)


josephjclark:
@hunterachieng If this is finished and ready to merge, can you please take the PR off draft?

hunterachieng:
> @hunterachieng If this is finished and ready to merge, can you please take the PR off draft?

@josephjclark  I still have to update the tests if the implementation is approved

packages/http/src/Adaptor.js
```diff
@@ -5,9 +5,8 @@ import { request as sendRequest, xmlParser } from './util';
  * Options provided to the HTTP request
  * @typedef {Object} RequestOptions
  * @public
- * @property {object|string} body - body data to append to the request. JSON will be converted to a string (but a content-type header will not be attached to the request).
  * @property {object} errors - Map of errorCodes -> error messages, ie, `{ 404: 'Resource not found;' }`. Pass `false` to suppress errors for this code.
- * @property {object} form - Pass a JSON object to be serialised into a multipart HTML form (as FormData) in the body.
+ * @property {boolean} form - True/False if the body data is multipart HTML form (as FormData).
```
josephjclark, line 9: I would just say `True/False if the body data is multipart HTML form (as FormData).`

packages/http/src/Adaptor.js
```diff
@@ -60,13 +60,14 @@ export function execute(...operations) {
...
- * @param {RequestOptions} params - Query, Headers and Authentication parameters
+ * @param {object} data - Body data to append to the request. JSON will be converted to a string.
+ * @param {RequestOptions} options - Query, Headers and Authentication parameters
  * @param {function} callback - (Optional) Callback function
  * @state {HttpState}
  * @returns {Operation}
  */
-export function request(method, path, params, callback) {
-  return sendRequest(method, path, params, callback);
+export function request(method, path, data, options, callback) {
```
josephjclark, line 69: hmm, a bit surprised by this change :thinking: Can you say more about why this function changed?

Is it because of the `RequestOptions` typedef? If so, can you think of any alternative solutions?

packages/http/src/Adaptor.js
```diff
@@ -79,71 +80,73 @@ export function request(method, path, params, callback) {
  * })
  * @function
  * @param {string} path - Path to resource. Can be an absolute URL if baseURL is NOT set on `state.configuration`.
- * @param {RequestOptions} params - Query, Headers and Authentication parameters
+ * @param {RequestOptions} Options - Body, Query, Headers and Authentication parameters
```
josephjclark, line 83: Ooh good catch on this parameter name :+1:

packages/http/src/Adaptor.js
```diff
@@ -79,71 +80,73 @@ export function request(method, path, params, callback) {
...
 }
 
 /**
  * Make a POST request. If `configuration.baseUrl` is set, paths must be relative.
  * @public
  * @example
- *  post('/myEndpoint', {
- *    body: {'foo': 'bar'},
+ *  post('/myEndpoint',
+ *  {'foo': 'bar'},
```
josephjclark, line 97: Maybe `$.data` is a more useful example? With a helpful caption?

packages/http/src/Adaptor.js
```diff
@@ -79,71 +80,73 @@ export function request(method, path, params, callback) {
...
+  return sendRequest('POST', path, { body: data, ...options }, callback);
 }
 
 /**
  * Make a PUT request. If `configuration.baseUrl` is set, paths must be relative.
  * @public
  * @example
- *  put('/myEndpoint', {
- *    body: {'foo': 'bar'},
+ *  put('/myEndpoint',  {'foo': 'bar'},{
```
josephjclark, line 118: There's some untidy formatting here - worth fixing

packages/http/src/Adaptor.js
```diff
@@ -180,18 +183,18 @@ export function del(path, params, callback) {
  *   ($) => {
  *     return $("table[class=your_table]").parsetable(true, true, true);
  *   },
- *   (next) => ({ ...next, results: next.data.body })
+ *   (next) => ({ ...next, results: next.data.data })
  * );
  * @function
- * @param {String} body - data string to be parsed
+ * @param {String} data - Body string to be parsed
```
josephjclark, line 189: Another good naming fix - well done!

packages/http/src/util.js
```diff
@@ -99,7 +99,10 @@ export function request(method, path, params, callback = s => s) {
 
     const options = {
       ...resolvedParams,
-      headers,
+      headers: {
```
josephjclark, line 102: Worried about this fix. Isn't this going to break form uploads, because we're setting the header to `JSON` when actually we're sending `multi-part` form?

josephjclark (commented):
@hunterachieng I've left some comments. I'm worried about the `request` change, as well as form handling

packages/http/src/Adaptor.js
hunterachieng: I changed it to match post, but I can let it look the way it was before with only changing `params` to `options`

josephjclark: Yes please

This is a general function and could be used for anything

`get` and `post` are specialists - they're designed to do one thing very well.

The signatures must reflect this :)

packages/http/src/util.js
```diff
@@ -99,7 +99,10 @@ export function request(method, path, params, callback = s => s) {
...
+        ...(resolvedParams.json ? { 'Content-Type': 'application/json' } : {}),
```
josephjclark, line 103: Sorry but I still don't understand this diff :grimacing: 

What problem are we trying to solve here? We're adding a behavioural change but it seems a) unrelated to the main PR and b) related to a property that's actually deprecated

josephjclark:
@hunterachieng you can carry on and add/fix unit tests.

You don't always need my review to start writing tests. Only if you're super unsure of the implementation.

The value of tests is that that help you to understand, explain and verify the behaviour of the code you're implementing. Not sure if you're doing the right thing? Write a unit test to explore the behaviour! Has a test started failing? Look closely at _why_ and decide whether you expect this failure or not as a result of your code changes.

packages/http/src/util.js
hunterachieng: Since we do not want to add `Content-Type: application/json` when there is formData, I added the check to ensure we only use it when `form:false`. I have changed to use `resolvedParams.form` instead

josephjclark: Maybe the right thing to do here is:

- add an option called `contentType`, which defaults to `json`. 
- If  `contentType` is `json`, `xml` or `string`, we expand that to the proper MIME type and set the content type header
- If `contentType` is falsy, we do not do any automatic header stuff
- Any other value gets written directly to the content-type header, we use that value instead of our own automation. In other words `options.headers` should spread over whatever default headers we set.
- this behaviour would be implemented in `util.request`, but needs documenting in `RequestOptions` and the changeset
- Finally! If `form` is set to true, surely we need to set the content type to multipart form? I don't think we  do this properly in production right now and I'm worried about it

Oh, wait, even better! Sorry I know this is a bigger change, but if `content-type` is set to `form`, then we do the header and the `form` option. So we replace `form: true` with `contentType: 'form'`

How does that sound?

It means that by default we send JSON, but it's very easy for users to override this.

There are quite complex variations here so I'd like a really good set of unit tests which validate the behaviour.

If we do this, we'll make it a new issue and we can merge this as-is. So _this_ PR is about changing the signatures, and we'll open a new PR to properly control and default `contentType`

hunterachieng: @josephjclark What I had implemented yesterday was breaking the tests, so I switched to defaulting it. 
Nevertheless, I do agree with using `contentType` instead of `form` and implementing the different `content-type's`

josephjclark: That's a great validation of tests :) But changing a default is a really big deal, and it needed to be made clearer.

Can you please copy my comment above, paste it into a new issue, and edit it down to a clean spec? Thank you!

hunterachieng: Doing that now

hunterachieng: Here is the issue link [https://github.com/OpenFn/adaptors/issues/1005](url)

josephjclark: Thank you!

josephjclark (approved):

PR #1006, opened 2025-02-20
Phase 6 improve OpenMRS adaptor docs

### LINKED ISSUE #492: Phase 6: Improve documentation
OpenMRS docs don't currently give you a lot of help with query APIs:

image

A couple of comments on this:

- "Sarah"  in the example should be in a string. The example is invalid Javascript.
- No query arguments are documented. What shape is the`query` object? What do you want from me?
- What does the `q` query string actually mean? According to the openMRS docs it means "display name of patient" - so we should probably reflect that somewhere.

The wider issue is that each endpoint has a bunch of different query parameters, where `q` seems pretty common (although maybe with different semantics), and `limit` is usually accepted I think but even OpenMRS don't document it properly.

The query object provided by our adaptor maps 1:1 with the query parameters in the OpenMRS rest API.

So:  I think each operation should have a better, more correct high level explanation, and include a link to the query parameters documented on openMRS. I do not think we should document each parameter on our own docs.

Maybe something like this mock up?

image

## API improvement

I wonder if an improvement on the  API would be something like:

```
searchPatient(name, options, callback)
```

Where name is the query string (maps to`q`, and we document that` and, options is an optional list of other parameters. I wonder if that scales nicely across the adaptor

PiusKariuki:
## Summary

OpenMRS docs don't currently give you a lot of help with query APIs

Fixes #492 

## Details

- add `@state` annotations to every function (including the http and fhir namespace).

- make sure the get() function makes sense and has examples of getting a list, searching a list by name, getting a single resource, and getting a subresource.

- Add a dedicated typedef for OpenMRS request params

- Remove `throw` declaration in `upsert` docbloc

- Update `upsert` examples signatures

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/eleven-spiders-play.md
+++ b/.changeset/eleven-spiders-play.md
@@ -1,0 +1,9 @@
+---
+'@openfn/language-openmrs': minor
+---
+
+- Added `state` annotations for all requests.
+- Improved the examples on the `get` function. 
+- Added a typedef for OpenMRS options that only includes query params.
+- Removed the `throw` declaration on the upsert docbloc.
+- Updated the`upsert`example signatures.
--- a/packages/openmrs/ast.json
+++ b/packages/openmrs/ast.json
@@ -7,22 +7,32 @@
         "options"
       ],
       "docs": {
-        "description": "Make a get request to any OpenMRS REST endpoint.",
+        "description": "Fetch resources from OpenMRS. Use this to fetch a single resource,\nor to search a list. Query parameters will be appended to the request URL,\nrefer to {@link https://rest.openmrs.org/ OpenMRS Docs} for details.",
         "tags": [
           {
             "title": "example",
-            "description": "get(\"/patient/abc\")",
-            "caption": "Get a patient by UUID"
+            "description": "get(\"patient\")",
+            "caption": "List all patients"
           },
           {
             "title": "example",
-            "description": "get(\"/patient\", {query: {q: 'Jon', limit: 1}})",
-            "caption": "Get a patient with query and limit"
+            "description": "get(\"patient\", { q: \"brian\", limit: 1 })",
+            "caption": "List patients by name with a limit(<a href=\"https://rest.openmrs.org/#search-patients\">see OpenMRS API</a>)"
           },
           {
             "title": "example",
-            "description": "get(\"/patient/abc/allergy/xyz\")",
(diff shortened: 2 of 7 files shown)


josephjclark:
@PiusKariuki I can't tell from the openMRS docs - what is actually returned from a POST, PUT or DELETE? Do we get anything back in state.data?

josephjclark:
Also Pius our existing docs for update say:
> Updating an object requires to send `all required fields` or the `full body`

But if I look at the openmrs docs, it says:

> Update a person. This method only modifies properties specified in the request

So I think our OG docs are just plain wrong?

I'm going to push a commit with corrections, but it  would be great if you could verify :)

PiusKariuki:
Yeah there's a message after a POST or PUT request. I make a commit to show this.

josephjclark:
Thanks @PiusKariuki - if you see this please don't commit yet! I'm just making a few adjustments

josephjclark:
@PiusKariuki this is really excellent work, thank you! I've gone through and thought carefully about some things. The docs are so good and clean now that I can really scrutinize the API. And it looks fabulous!

One thing I notice is that the get and delete options should just be query objects, so I've refactored `get(path, { query: { q: '.' }  })  to `get(path, { q: '.'  })`

Since we're merging into the epic I've done ahead and done that here, and updated some tests to that effect.

PiusKariuki:
@josephjclark thank you! Great idea with the options too.

josephjclark:
Ok @PiusKariuki  just one small thing now, can you update the state.data docs for create and update? Checkout what I did for `get` as reference. I just want to avoid HTTP semantics in the docs for those functions if we can.

PiusKariuki:
Sure thing I'll get on this shortly.

josephjclark:
@PiusKariuki if it's accurate then it's perfect :)

PiusKariuki:
I can confirm it's accurate. I even tested the behaviour on postman first.

josephjclark (approved):

josephjclark:
@PiusKariuki thank you!

We're nearly there now. This is such a lovely clean API, I'm really pleased with it

(I just hope our users agree!)

PiusKariuki:
@josephjclark Happy to help!

PR #1016, opened 2025-02-24
odoo: default `create` options to `{}` and remove logs

### LINKED ISSUE #1010: odoo: default `create` options to `{}` and remove logs
When creating a new resource, it is optional to send an externalId. To make this clear to the users, we have to update our examples and default the `create` options to `{}`.

We also need to remove all the `console.logs()` in this function

hunterachieng:
## Summary

Default the `create()` function in `odoo` to  `{ }`. Update the logs to a more descriptive sentence such as `Creating a res.partner resource...` 

Fixes #1010

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/cool-actors-argue.md
+++ b/.changeset/cool-actors-argue.md
@@ -2,5 +2,6 @@
 '@openfn/language-odoo': patch
 ---
 
-Default `create()` to `{}` and update logs to a descriptive message:
-`Creating a res.partner resource...`
+Default `create()` options to `{}` to ensure that the options argument is optional.
+
+
--- a/.changeset/cool-actors-argue.md
+++ b/.changeset/cool-actors-argue.md
@@ -1,0 +1,6 @@
+---
+'@openfn/language-odoo': patch
+---
+
+Default `create()` to `{}` and update logs to a descriptive message:
+`Creating a res.partner resource...`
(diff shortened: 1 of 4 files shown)


.changeset/cool-actors-argue.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-odoo': patch
+---
+
+Default `create()` to `{}` and update logs to a descriptive message:
```
josephjclark, line 5: Since this is two different fixes, it should be listed as two different bullet points (_technically_ it should actually be two different changeset files)

It's not clear what "Default `create()` to `{}`". But I'd focus on the user problem. What does this fix mean for users? How does it affect them? Well, it prevents an error being thrown when `options` is not set for create. Or it ensures that the `options` argument of `create` is optional. Something like that.

I don't think you need to explain the logging chance in much detail. Again, what does it matter to users? Well, maybe it's important to say that we no longer log user data. Maybe it's enough to say that we've simplified (or tidied up) logging across the adaptor. I think the only bit that users really care about is that their data isn't getting logged

hunterachieng: Fixed

josephjclark (approved):

PR #1022, opened 2025-02-28
http: control and default contentType

### LINKED ISSUE #1005: http: control and default contentType
The HTTP adaptor processes different types of data such as `xml`, `json`, and `form`. 
With these different types, we need to allow users to send which type of data they are sending, but we also need to have a default one set.
What needs to be done include:

- Adding an option called contentType, which defaults to json.
- If contentType is json, xml or string, we expand that to the proper MIME type and set the content type header.
- If contentType is falsy, we do not automate the header.
- If any other value gets written directly to the content-type header, we use that value instead of our own automation. Ensure `options.headers` spreads over the default headers set.
- Replace form: true with contentType: 'form'

Implement the behaviour in `util.request`, and document in RequestOptions and the changeset.

hunterachieng:
## Summary

Replace `form` with `contentType` in `http` to allow automatic content-header configuration. 

Fixes #1005 

## Details

The `http` adaptor processes different types of data such as `xml`, `json`, and `form`.
With these different types, we need to allow users to send which type of data they are sending, but we also need to have a default one set, hence using an options `contentType` that replaces `forms`

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/http/src/util.js
+++ b/packages/http/src/util.js
@@ -72,16 +72,26 @@ export function request(method, path, params, callback = s => s) {
 
     let { body, contentType = 'json', headers = {} } = resolvedParams;
 
-    const hasContentType = Object.keys(headers).some(
+    const contentTypeKey = Object.keys(headers).find(
       key => key.toLowerCase() === 'content-type'
     );
+
+    const hasContentType = !!contentTypeKey;
+    const contentTypeValue = hasContentType ? headers[contentTypeKey] : '';
 
     if (hasContentType) {
       headers = { ...headers };
     } else if (contentType === 'form') {
       body = encodeFormBody(body);
     } else {
-      headers = { ...headers, 'Content-Type': CONTENT_TYPES[contentType] };
+      headers = {
+        ...headers,
+        'Content-Type': CONTENT_TYPES[contentType] || 'application/json',
+      };
+    }
+
+    if (contentTypeValue.toLowerCase().includes('multipart/form-data')) {
+      body = encodeFormBody(body);
     }
 
     const baseUrl = state.configuration?.baseUrl;
--- a/packages/http/src/util.js
+++ b/packages/http/src/util.js
@@ -72,14 +72,17 @@ export function request(method, path, params, callback = s => s) {
 
     let { body, contentType = 'json', headers = {} } = resolvedParams;
 
-    const hasContentType = Object.keys(headers).some(
+    const hasContentType = Object.keys(headers).find(
       key => key.toLowerCase() === 'content-type'
     );
 
-    if (hasContentType) {
+    if (contentType === 'form') {
+      delete headers[hasContentType];
       headers = { ...headers };
-    } else if (contentType === 'form') {
+
       body = encodeFormBody(body);
+    } else if (hasContentType) {
+      headers = { ...headers };
     } else {
       headers = {
         ...headers,
--- a/packages/http/src/util.js
+++ b/packages/http/src/util.js
@@ -72,19 +72,11 @@ export function request(method, path, params, callback = s => s) {
 
     let { body, contentType = 'json', headers = {} } = resolvedParams;
 
-    const contentTypeKey = Object.keys(headers).find(
+    const hasContentType = Object.keys(headers).some(
       key => key.toLowerCase() === 'content-type'
     );
 
-    const hasContentType = contentTypeKey && headers[contentTypeKey];
-
-    const contentTypeValue = hasContentType
-      ? headers[contentTypeKey].toLowerCase()
-      : '';
-
-    const shouldEncodeForm =
-      (!hasContentType && contentType === 'form') ||
-      contentTypeValue.includes('multipart/form-data');
+    const shouldEncodeForm = !hasContentType && contentType === 'form';
 
     if (shouldEncodeForm) {
       body = encodeFormBody(body);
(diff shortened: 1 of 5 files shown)


packages/http/src/util.js
```diff
@@ -64,10 +70,14 @@ export function request(method, path, params, callback = s => s) {
       params
     );
 
-    let { body, headers = { 'Content-Type': 'application/json' } } = resolvedParams;
+    let { body, contentType = 'json', headers = {} } = resolvedParams;
 
-    if (resolvedParams.form) {
+    if (resolvedParams.headers) {
```
josephjclark, line 75: I don't think this logic is right - but you should be able to prove that to me in your tests

But If the user passes in headers, the content-type will never be set

hunterachieng: @josephjclark  I have changed the logic

packages/http/test/index.js
```diff
@@ -286,8 +286,10 @@ describe('get()', () => {
         headers: { 'x-openfn': 'testing' },
       })
     )(state);
-
-    expect(data).to.eql({ 'x-openfn': 'testing' });
+    expect(data).to.eql({
```
josephjclark, line 289: This test change reveals a very interesting, very subtle, behaviour change!

Before, if the user passed a header object, we send that object straight to undici. We don't otherwise mess with it.

Now, as a result of these changes, we always default the content-type to JSON.

I think this change is an improvement! But it might break the behaviour of an existing workflow, and it is not mentioned in the changeset notes. Changing a default like this is usually considered a breaking change (ie, requires a major version bump)

Every time a test break starts failing, an alarm needs to go off in your head :rotating_light: :rotating_light: :rotating_light:  Why did this test fail? Did I change a behaviour? Do I expect this change? Was it anticipated in the spec? Who do I need to talk to about it?

josephjclark (changes requested):
Thank you Hunter

I'd like to look more closely at the tests. I think I see a bug but I'd like you to catch it with a unit test (or prove to  me that it's not a bug!)

There is only one test on the contentType key.

But there are lots of subtle behavioural changes we need to look at.

What happens if I pass contentType AND headers? What happens if I pass contentType AND headers with a content-type header already set? What happens if I pass an invalid/unsupported content type? What if I set contentType: form and also set the content type header to multipart form? Or accidentally set the content type to json?

I don't expect us to handle every single edge case: but I do want to see tests that exercise the code properly and prove that we've thought things through.

packages/http/src/util.js
```diff
@@ -76,24 +76,27 @@ export function request(method, path, params, callback = s => s) {
       key => key.toLowerCase() === 'content-type'
     );
 
-    const hasContentType = !!contentTypeKey;
-    const contentTypeValue = hasContentType ? headers[contentTypeKey] : '';
+    const hasContentType = contentTypeKey && headers[contentTypeKey];
```
josephjclark, line 79: But `contentTypeKey` is only exists in the user's headers? So this test is not needed

packages/http/src/util.js
```diff
@@ -76,24 +76,27 @@ export function request(method, path, params, callback = s => s) {
...
 
-    if (hasContentType) {
-      headers = { ...headers };
-    } else if (contentType === 'form') {
+    const contentTypeValue = hasContentType
```
josephjclark, line 81: This ternary is a little strange. I'm always suspicious of ternary expressions where the `false` branch is a falsy value. They are very rarely needed. It makes me think something isn't quite right about the solution.

josephjclark (commented):
Hi @hunterachieng 

The complexity of this code still bothers me :thinking:  My instinct is that there is just too much conditional stuff for what is really as simple rule set. I've left some comments calling out bits that particularly bother me.

I've been thinking about it for a little while (probably too long to be honest!) and what I've come up with is this:

What are the rules?
1. If the user says `contentType: form`, we need to encode form data
2. Otherwise if the user sets their own content type header, we'll trust them and just use that
3. Otherwise if the user gives us a contentType that we understand, we'll set the header
4. Otherwise we'll default the content to JSON because that should satisfy like 95% of use-cases

These rules, I think, slightly simplify/change some of our thinking about the form handling. But as a result,I think the code gets much much simpler.

I've drafted what I think is a model code solution to implement this and it's _tiny_. Would you like to have a go and we'll compare notes? See how short and simple you can get the code.

Please do spend some time on this and really think it through. I know I'm pushing hard on it. I take full responsibility for messing up estimates on the sprint (CC @martalovescoffee)

packages/http/test/index.js
```diff
@@ -572,37 +572,6 @@ describe('contentType', () => {
     expect(JSON.parse(response.data).id).to.eql(1);
     expect(req.headers['Content-Type']).to.equal('application/json');
   });
-
-  it('should handle multipart/form-data if sent in the headers ', async () => {
```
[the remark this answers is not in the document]
hunterachieng: We already have a test that looks for this. This one was mainly checking if the body will be encoded if we send `multipart/form-data` in the content headers.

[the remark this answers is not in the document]
josephjclark: Ok, if it's totally redundant that's fine. Do you mean `should prefer content-type header to contentType option` ?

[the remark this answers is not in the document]
hunterachieng: Yes. That test was checking for that

PR #1023, opened 2025-02-28
Openmrs pagination

### LINKED ISSUE #1003: OpenMRS pagination
We've hit on a couple of problems with OpenMRS pagination.

In 4.3.0, if you do searchPatient({ q: 'foo', limit: '100' } (does limit really have to be a string??) , it'll auto-page all data in pages of 100. It will not download 100 search results and return.

In other words, limit is actually working as pageSize, not limit, which is super misleading.

Against the new 5.0 branch, what we need to ensure is this:

- The REST API should basically auto-page by default to download the desired number of records. It should provide `limit` `pageSize` and `offset` options. These do **not** map directly to url query parameters. `pageSize` controls how many are downloaded per page. `limit` controls how many in total are downloaded. The API needs to take care to not download more than the limit. I _think_ by default we should download everything? Note that just 400 patient records seems to blow up the worker (although we're working on that)
- The `http.*` namespace should **not** auto-page by default. It should do whatever you ask of it.If you pass `query.limit`, it'll just send that to the server and give you a page of data
- _later_ we can add either a `autopage: true` option to `http.get`? Or maybe we support `options.limit` and `options.pageSize`, and for `http.get` we'll honour them for you?. But if we disable autopage by default for `http`, we can add options to enable it as minor patches later.

PiusKariuki:
## Summary

Fix pagination and include a pageSize option.

Fixes #1003

## Details

The `limit` option would act as `pageSize`  when it wasn't accompanied by a `startIndex`  parameter.  

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -65,11 +65,9 @@ export async function request(state, method, path, options = {}) {
   let queryParams = opts?.query;
   const requestOptions = queryParams ? { ...opts, query: queryParams } : opts;
 
-  if (shouldPaginate({ underGlobalLimit: true, hasPageSize: query.pageSize })) {
-    return paginate(method, url, requestOptions);
-  } else {
-    return fetchAndLog(method, url, requestOptions);
-  }
+  return (shouldPaginate({ underGlobalLimit: true, hasPageSize: query.pageSize }))
+    ? paginate(method, url, requestOptions)
+    : fetchAndLog(method, url, requestOptions);
 }
 
 export async function paginate(method, url, opts) {
@@ -95,7 +93,7 @@ export async function paginate(method, url, opts) {
     const response = await fetchAndLog(method, url, paginatedRequestOptions);
 
     // Stop itereation if there are no more objects
-    if(!response.body){
+    if (!response.body) {
       break;
     }
 
--- a/packages/openmrs/src/Utils.js
+++ b/packages/openmrs/src/Utils.js
@@ -15,6 +15,16 @@ export const prepareNextState = (state, response, callback = s => s) => {
   return callback(nextState);
 };
 
+export async function fetchAndLog(method, url, opts) {
+  try {
+    const response = await commonRequest(method, url, opts);
+    logResponse(response);
+    return response;
+  } catch (err) {
+    return err;
+  }
+}
+
 export async function request(state, method, path, options = {}) {
   const {
     baseUrl = state.configuration.instanceUrl,
@@ -38,6 +48,7 @@ export async function request(state, method, path, options = {}) {
   }
 
   const { username, password } = state.configuration;
+
   const authHeaders = makeBasicAuthHeader(username, password);
 
   const opts = {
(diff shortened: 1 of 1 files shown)


PiusKariuki:
@josephjclark what do you think of this? The `get` function can now call only this function and it will work seamlessly for both paginated and non-paginated scenarios. I am making just this one commit for your review to make sure we align before I go on to write tests and invoke it.

josephjclark:
Hey @PiusKariuki, thank you for this!

A couple of things leap out at me:

- Even as a first pass, I'm concerned about the duplication here. We should be taking the do/while out of `request`, and `requestWithpagination` basically then becomes the do/while wrapper around `request`
- `requestWithPagination` presumably only supports `GET`, so we wouldn't need `method` in the signature
- The handling of `limit`, `pageSize` and `startIndex` is kinda tricky. Ok, I think `options` accept have those three values (as in `options.limit`, not `options.query.limit`), and the function should then set the query in the appropriate way.The user basically should not be setting those values on the query themselves. If they do, we should either ignore or override, depending on various factors.
- The `shouldAllowPagination` function looks very dense and complex and makes me a bit uncomfortable. If I as a dev want to understand when and why this function paginates, this looks like the funcction I want to use... but I also don't think it tells the whole story. And not in a clear way.
- If we fetch a page, and we're under the requested limit, it looks like this function will just go off and ask for another page. So if limit is `101` and page size is `100`, we'll actually download 200 items. I think you need to handle the `query.limit` param you send a lot more carefully. There's quite a good pattern for this in the collections adaptor if you want some inspiration.

PiusKariuki:
Thank you for this @josephjclark . I think with this I have an idea of the kind fo solution you need. I will rework this in a way that is no so duplicative.

PiusKariuki:
@josephjclark what do you think of this approach.

packages/openmrs/src/Utils.js
```diff
@@ -51,43 +62,84 @@ export async function request(state, method, path, options = {}) {
...
+  // global limit variable that is a source of truth incase there is a pageSize variable
+  const globalLimit = opts.query.limit;
+
+  // Declare a variable that takes in all the options. We can then modify this variable to fetch the next page
+  let paginatedRequestOptions = { ...opts }
+
+
+  // if page size variable exists use that as limit. Delete pageSize from params since OpenMRS does recognize it.
+  if (opts.query.pageSize) {
+    paginatedRequestOptions.query.limit = opts.query.pageSize;
```
josephjclark, line 88: This needs to be pagesize OR (total items - current items), whichever is smaller

packages/openmrs/src/Utils.js
```diff
@@ -51,43 +62,84 @@ export async function request(state, method, path, options = {}) {
   };
 
   const url = `${baseUrl}${path}`;
+  let queryParams = opts?.query;
+  const requestOptions = queryParams ? { ...opts, query: queryParams } : opts;
+
+  return (shouldPaginate({ underGlobalLimit: true, hasPageSize: query.pageSize }))
```
josephjclark, line 68: We should paginate if:
1) it's a GET
2) The number of downloaded items is less than the number of requested items
3) The user is not controlling pagination (ie, the user passed `startIndex`)

packages/openmrs/src/Utils.js
```diff
@@ -15,6 +15,16 @@ export const prepareNextState = (state, response, callback = s => s) => {
   return callback(nextState);
 };
 
+export async function fetchAndLog(method, url, opts) {
+  try {
+    const response = await commonRequest(method, url, opts);
+    logResponse(response);
+    return response;
+  } catch (err) {
+    return err;
```
josephjclark, line 24: I think we should let this throw?

josephjclark:
Next steps:

@PiusKariuki to make changes per these comments and add some unit tests

@josephjclark to update the API and take on the other issues on the epic

josephjclark:
Hi @PiusKariuki, how did you get on with this on Friday? Would you like to hand it over and I'll finish it?

Can you help me get a test sandbox set up?

PiusKariuki:
Hi @josephjclark. So sorry, I was swamped last week barely got anything done. If you have the time then you can take it up please instead of having me block everyone.😄

PiusKariuki:
@josephjclark I can help out with the sandbox setup. I know some running instances that would be great for just read operations too if you just want to test those.

josephjclark:
Great @PiusKariuki,thanks. Can you DM me sandbox credentials on Slack? I'll put some time aside this week to get all this sorted

josephjclark:
Closing in favour of #1093

PR #1042, opened 2025-03-07
odoo: Add `downloadNewRecord` to create

### LINKED ISSUE #1008: odoo: Add `downloadNewRecord` to create
When a new instance is created in `odoo` , only the `id` of the created entity is returned and not the whole payload plus the id.

We need to add a `downloadNewRecord` option in the `create()` and default it to `false`. 

If `downloadNewRecord` is truthy, we create a new entity and get all the data for the created entity and return that plus the id to the user

Add tests and update the documentation.

hunterachieng:
## Summary

Add a `downloadNewRecord: true` option during `create` to enable downloading of the
whole record in the response.

Fixes #1008

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


josephjclark:
Thank you @hunterachieng  - I'll review this next week

packages/odoo/src/Adaptor.js
```diff
@@ -60,33 +60,44 @@ async function login(state) {
...
  * @public
  * @example <caption> Create a partner record with an external Id</caption>
  * create("res.partner", { name: "Kool Keith" }, {externalId: 23});
+ * @example <caption> Create a partner record and download the whole record in the response</caption>
+ * create("res.partner", { name: "Kool Keith" }, {downloadNewRecord: true});
  * @function
  * @param {string} model - The specific record model i.e. "res.partner"
  * @param {object} data - The data to be created in JSON.
- * @param {object} options - Options to send to the request. Includes an optional external ID for the record.
+ * @param {object} options - Options to send to the request. Includes an optional external ID for the record and a downloadNewRecord option defaulted to `false` incase a user intends to receive the whole created resource in the response.
```
josephjclark, line 74: Can we create a TypeDef for CreateOptions? This documentation is very dense and won't scale

packages/odoo/src/Adaptor.js
```diff
@@ -60,33 +60,44 @@ async function login(state) {
...
-    const response = await odooConn.create(
+    let response = await odooConn.create(
       resolvedModel,
       resolvedData,
       resolvedOptions?.externalId
     );
 
+    if (resolvedOptions.downloadNewRecord) {
+      console.log(`Fetching ${resolvedModel} resource with id ${response}...`);
+      response = await odooConn.read(resolvedModel, [response], []);
```
josephjclark, line 98: Oh, that's interesting :thinking: 

I would say that we should always return the `create` response, not the subsequent `read` response. Because the read should be sort of invisible. But I'm open to argument on that.

Also: we should only do this if the creation was successful. Presumably if it failed then `create` will throw? I note that we don't have any tests around this. Maybe we can add a test for if, say, credentials are invalid or if the record already exists?

packages/odoo/src/Adaptor.js
```diff
@@ -60,33 +60,44 @@ async function login(state) {
...
 /**
- * Create a record in Odoo
+ * Create a record in Odoo.
+ * You can pass an external ID to the options object to create a record with a specific ID.
+ * You can also pass a downloadNewRecord option to download the whole created resource in the response.
  * @public
  * @example <caption> Create a partner record with an external Id</caption>
  * create("res.partner", { name: "Kool Keith" }, {externalId: 23});
+ * @example <caption> Create a partner record and download the whole record in the response</caption>
+ * create("res.partner", { name: "Kool Keith" }, {downloadNewRecord: true});
```
josephjclark, line 70: Formatting is kind of ugly here - can we add better spacing?

packages/odoo/src/Adaptor.js
```diff
@@ -60,33 +60,44 @@ async function login(state) {
   return state;
 }
 /**
- * Create a record in Odoo
+ * Create a record in Odoo.
```
josephjclark, line 63: I note that we have no `@state` docs on create (or the other apis). Can we add them now? state.data is interesting here because it changes depending on `downloadNewRecord`

josephjclark (changes requested):
One thing that strikes me here is that in the examples, the value is in `state.data`, which means we need a second operation to use it.

But I prefer examples to be a single operation.

Maybe we can add something to the Examples docs, in the main docs repo? We can have a multi-line example there which creates a record from $.data, then take takes the new object on state.data, logs it, and aliases it on state for the next step

packages/odoo/test/Adaptor.test.js
```diff
@@ -97,6 +97,29 @@ describe('create', () => {
...
+      },
+    };
+    setMockClient(mock);
+
+    await create(
+      'res.partner',
+      { name: 'Jane Doe' },
+      { externalId: 23 }
+    )(state).catch(error => {
+      expect(error).to.be.an('error');
```
josephjclark, line 119: Maybe here we can just assert that the error message includes the string "UniqueViolation"? Just so that we know we're catching the right sort of error.

But generally nice work on this :+1:

hunterachieng: Thank you. Let me add that

packages/odoo/src/Adaptor.js
josephjclark: Ah, I think I misunderstood the code in my previous comment here.

But I'm also being mislead by the code, so I don't feel _too_ bad about it :sweat_smile: 

I think:
- forget what I said about "invisible"
- We should restore the log line
- We should fix variable names for clarity (`response` is super confusing - this isn't a HTTP response, which is normally what we mean)

```
 try {
      const newRecordId = await odooConn.create(
        resolvedModel,
        resolvedData,
        resolvedOptions?.externalId
      );

      if (resolvedOptions.downloadNewRecord) {
         const newRecord = await odooConn.read(resolvedModel, [newRecordId], []);
         return composeNextState(state, response);
      } else {
         return composeNextState(state, newRecordId);
     }
    } catch (e) {
      console.error(`Error creating ${resolvedModel} resource: ${e}`);
      throw e;
    }
```

josephjclark:
@hunterachieng Thank you, this good to go. Can you please do the release yourself?

1. Run `pnpm changeset vesion`
2. Check that the changelog for odoo is correct and you're happy with it
3. Run `pnpm install && pnpm build` just to make sure everything is up to date (you don't have to build the whole repo, you can just  build odoo)
4. Commit all changes and push
5. When CI passes, double-check the diff to make sure there's nothing weird), and then Squash and Merge the PR

josephjclark (approved):

Issue #975, opened 2025-02-05
dhis2: remove axios


josephjclark:

aleksa-krolls:
See #974

josephjclark:
Now that axios is maintained again, is it worth replacing it?

Maybe not.

Arguments for removing axios:

- consistency across the code base
- consistency across user APIS (the http namespace in dhis2 and commcare is exactly the same)
- consistency across automated logging
- consistency in error reporting and handling

Arguments against removing it:

- It'll take a lot of time for low valie
- We may introduce subtle new bugs

I think I'm gonna say: let's keep axios for now and focus on the higher value things (like organising the API and fixing the tracker/create APIs). We'll keep the http namespace as consistent as we can, within reason.

Then, if anyone makes a complaint around consistency, error handling, logging or API design, we'll jump in and replace axios

josephjclark:
counterpoint: maybe we can keep axios but we absolutely _have_ to fix this:

![Image](https://github.com/user-attachments/assets/05f7e72c-2eb4-457c-ba7e-e39238afb2a9)

PR #1046, opened 2025-03-10
dhis2: Implement `http` namespace

### LINKED ISSUE #978: dhis2: move http helpers into `http.` namespace
- Move `get`, `patch`, `post` into a new http namespace, like [openMRS](https://docs.openfn.org/adaptors/packages/openmrs-docs)
- Add a `request` helper too
- Remove callbacks from all APIs
- Make sure documentation is correct with good examples

hunterachieng:
## Summary

Move all the generic functions into a `http` namespace, remove callbacks and add a generic `request` function.

Fixes #978 

## Details

All the generic functions such as `post`, `get`, and `update` are moved to their own name-spaced file to separate them from the other specific functions, and a generic `request` function was added as well.

All callbacks are being removed since the same behaviour can be achieved with promise chaining

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/dhis2/src/Adaptor.js
+++ b/packages/dhis2/src/Adaptor.js
@@ -13,6 +13,7 @@ import {
   ensureArray,
 } from './Utils';
 import { request } from './Client';
+import * as http from './http';
 
 /**
  * Execute a sequence of operations.
@@ -135,7 +136,6 @@ axios.interceptors.response.use(
  * @magic resourceType $.children.resourceTypes[*]
  * @param {Dhis2Data} data - Object which defines data that will be used to create a given instance of resource. To create a single instance of a resource, `data` must be a javascript object, and to create multiple instances of a resources, `data` must be an array of javascript objects.
  * @param {Object} [options] - Optional `options` to define URL parameters via params (E.g. `filter`, `dimension` and other import parameters), request config (E.g. `auth`) and the DHIS2 apiVersion.
- * @param {function} [callback] - Optional callback to handle the response
  * @returns {Operation}
  * @example <caption>Create a program</caption>
  * create('programs', {
@@ -252,7 +252,7 @@ axios.interceptors.response.use(
  *   ],
  * });
  */
-export function create(resourceType, data, options = {}, callback = s => s) {
+export function create(resourceType, data, options = {}) {
   return state => {
     console.log(`Preparing create operation...`);
 
@@ -288,7 +288,7 @@ export function create(resourceType, data, options = {}, callback = s => s) {
       const { location } = result.headers;
       if (location) console.log(`Record available @ ${location}`);
 
-      return handleResponse(result, state, callback);
+      return handleResponse(result, state);
     });
   };
 }
@@ -302,7 +302,6 @@ export function create(resourceType, data, options = {}, callback = s => s) {
  * @param {string} path - The `id` or `path` to the `object` to be updated. E.g. `FTRrcoaog83` or `FTRrcoaog83/{collection-name}/{object-id}`
  * @param {Object} data - Data to update. It requires to send `all required fields` or the `full body`. If you want `partial updates`, use `patch` operation.
  * @param {Object} [options] - Optional `options` to define URL parameters via params (E.g. `filter`, `dimension` and other import parameters), request config (E.g. `auth`) and the DHIS2 apiVersion.
- * @param {function} [callback]  - Optional callback to handle the response
  * @returns {Operation}
  * @example <caption>a program</caption>
  * update('programs', 'qAZJCrNJK8H', {
@@ -423,13 +422,7 @@ export function create(resourceType, data, options = {}, callback = s => s) {
  *   incidentDate: '2013-10-17',
  * });
  */
-export function update(
-  resourceType,
-  path,
-  data,
-  options = {},
-  callback = s => s
-) {
+export function update(resourceType, path, data, options = {}) {
   return state => {
     console.log(`Preparing update operation...`);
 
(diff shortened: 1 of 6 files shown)


.changeset/silly-bulldogs-complain.md
```diff
@@ -0,0 +1,24 @@
+---
+'@openfn/language-dhis2': minor
+---
+
+- Implement a namespaced `http` for all generic functions.
+- Add a generic request function in the `http` file.
+- Remove all callbacks
```
josephjclark, line 7: This was a surprise - I think you should have flagged this in the PR title

You did mention it in the description, so that's good. I missed it, which is my fault, but still. I'd like to see a big change like this reflected in the PR title. Because if I one day want to look through history and work out where callbacks were removed, the title would help me out a lot.

A good rule of thumb: major changes should be reflected in PR titles :)

.changeset/silly-bulldogs-complain.md
```diff
@@ -0,0 +1,24 @@
+---
+'@openfn/language-dhis2': minor
```
josephjclark, line 2: this is a major changeset because we've introduced breaking changes (`post()` no longer exists!)

You can just replace `minor` with `major` in the file here

packages/dhis2/src/Adaptor.js
```diff
@@ -634,31 +513,33 @@ export function upsert(
...
-          return create(resourceType, data, options)(state);
-        } else {
-          // Pick out the first (and only) resource in the array and grab its
-          // ID to be used in the subsequent `update` by the path determined
-          // by the `selectId(...)` function.
-          const path = resources[0][selectId(resourceType)];
-          return update(resourceType, path, data, options)(state);
-        }
-      });
+      promise = http
```
josephjclark, line 516: ooh this is really annoying

Because we're using the `http.get()` operation directly, we have to feed state back into the call here.

it's an antipattern really. We certainly don't want to encourage it, and this is a big, brand-defining adaptor.

We don't _have_ to do this in this PR, but what's the cost in using like `util.request` instead of `http.request`? Of using a function rather than an operation?

Also: if `asbase64` is true, I think this stuff might be broken? Because response.data is a string and all these tests will fail :grimacing:

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,228 @@
+import { expandReferences, encode } from '@openfn/language-common/util';
+import { generateUrl, handleResponse } from './Utils';
+import * as client from './Client';
+
+/**
+ * Options object
+ * @typedef {Object} RequestOptions
```
josephjclark, line 7: Worried about these docs but they're not technically part of this PR. Let me think about it. Referenced here: https://github.com/OpenFn/adaptors/issues/975#issuecomment-2711368315

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,228 @@
...
+ * @function
+ * @param {string} resourceType - The type of resource to be updated. E.g. `dataElements`, `organisationUnits`, etc.
+ * @param {string} path - The `id` or `path` to the `object` to be updated. E.g. `FTRrcoaog83` or `FTRrcoaog83/{collection-name}/{object-id}`
+ * @param {Object} data - Data to update. Include only the fields you want to update. E.g. `{name: "New Name"}`
+ * @param {Object} [options] - Optional configuration, including params for the update ({preheatCache: true, strategy: 'UPDATE', mergeMode: 'REPLACE'}). Defaults to `{operationName: 'patch', apiVersion: state.configuration.apiVersion, responseType: 'json'}`
+ * @returns {Operation}
+ * @example <caption>a dataElement</caption>
+ * http.patch('dataElements', 'FTRrcoaog83', { name: 'New Name' });
+ */
+// TODO: @Elias, can this be deleted in favor of update? How does DHIS2 handle PATCH vs PUT?
```
josephjclark, line 140: @mtuchi can you help resolve this? Should we just remove patch?

josephjclark (commented):
I am having a hard time seperating "move http methods out into a clean namespace" from "fix all the problems with DHIS2 http APIs"

Docs are worrying me, base64 is worrying me, and I wonder if handleResponse needs to move into here.

I need to break here today but I'll think about this stuff more tomorrow.

It doesn't technically matter which PR fixes these problems - but it does matter what's in the next release. Dhis2 7 really needs to have a clean, well documented API. And do get to there we'll need more work.

I'll come back with a fresh brain.

packages/dhis2/src/http.js
mtuchi: Hiya @josephjclark this is what i was able to gather regarding `PATCH` method in DHIS2

The [DHIS2 API](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/metadata.html#webapi_partial_updates) supports the PATCH method for various metadata endpoints. 

For example, you can use **PATCH** to:
- Update properties of metadata objects
- Add or remove items from collections
- Change domain types and value types
- Remove specific organization units from groups

When using the PATCH method with any of these endpoints, you must use the content type `application/json-patch+json` in your request headers as per the JSON patch specification. See [DHIS2 Documentation - Partial Updates](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/metadata.html#webapi_partial_updates) [DHIS2 Documentation - Bulk Sharing](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/sharing.html#webapi_bulk_sharing)

Meanwhile, you can use **PUT** on: 
- Metadata endpoints - For updating various metadata objects like data elements, organization units, etc. [DHIS2 Documentation - Metadata CRUD](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/metadata.html#webapi_metadata_crud)
- Data Store endpoints - For updating key-value pairs [DHIS2 Documentation - Data Store](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/data-store.html#webapi_data_store_update_values)
- Maintenance endpoints - For performing system maintenance operations [DHIS2 Documentation - Maintenance](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/maintenance.html#webapi_maintenance)
- Visualization endpoints - For updating visualizations [DHIS2 Documentation - Visualizations](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/visualizations.html#webapi_visualization_add_update_remove_visualizations)


That being said, I have never used the API endpoints in this discovery but i guess it doesn't hurt to maintain both `patch` and `put` or if we decide to remove them we should have a `http.request` function which will still allow me to specify `patch` or `put` when there is a need for that

mtuchi: I am happy with that, and since the link to DHIS2 docs is subjected to either metadata apis or sharing apis maybe we add an example with JSON Patch format.  For example we should convert this example 👇🏽 to openfn job code

```
PATCH /api/dataElements/{id}

[
  {"op": "add", "path": "/name", "value": "New Name"},
  {"op": "add", "path": "/valueType", "value": "INTEGER"}
]
```

josephjclark:
@hunterachieng after you've addressed my comments, you'll want to rebase this on top of #1059 (or maybe merge is easier). It might be a hard rebase so it might be easier to start again? You'll need to take a look and decide.

hunterachieng:
@josephjclark  I have re-worked this. Do you mind taking a look?

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,238 @@
+import { expandReferences, encode } from '@openfn/language-common/util';
+import * as util from './Utils';
+
+/**
+ * State object
+ * @typedef {Object} Dhis2State
```
josephjclark, line 6: Can we make state `@private`, so that it doesn't appear on the docsite?

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,238 @@
...
+ * Options object
+ * @typedef {Object} RequestOptions
+ * @property {object} query - An object of query parameters to be encoded into the URL
+ * @property {object} headers - An object of all request headers
+ * @property {string} [parseAs='json'] - The response format to parse (e.g., 'json', 'text', 'stream', or 'base64'. Defaults to `json`
+ * @property {string} [apiVersion=42] - The apiVersion of the request. Defaults to 42.
+ */
+
+/**
+ * Get data. Generic helper method for getting data of any kind from DHIS2.
```
josephjclark, line 22: Not a "generic" helper. Maybe a HTTP helper.

How about this:

> HTTP helper method for fetching data from DHIS2. All paths are relative to baseUrl + version in your configuration

That pattern should be applied to all helpers

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,238 @@
...
+ * });
+ * @example <caption>Get an image from a trackedEntityInstance.</caption>
+ * http.get('trackedEntityInstances/qHVDKszQmdx/BqaEWTBG3RB/image', {
+ *   headers:{
+ *       Accept: 'image/*'
+ *   },
+ *   parseAs: 'base64',
+ * });
+ */
+export function get(resourceType, options = {}) {
```
josephjclark, line 59: `resourceType` needs renaming to `path`

josephjclark (commented):
Ah, now the problem with this is is that there's no basic `get()` function to retrieve resources. This is partly my fault for not thinking this through.

What I'm looking for is something like OpenMRS, where we have a basic HTTP namespace and a fancier main rest namespace which doesn't use HTTP semantics. This is actually the focus of OpenMRS 5 (which isn't ready for release yet but see the PR)

The HTTP namespace is pretty good. I like that it's lower level - no clever tracker stuff. That's great. I've left a few comments but it's broadly right.

Let's finish the http namespace, but then we need to think a bit more about the main namespace. What should `get()` look like? Maybe we should just restore the old `get()` for now?

josephjclark:
Sorry @hunterachieng I posted this review a bit early - I'll update it in the morning :)

packages/dhis2/src/http.js
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Get data. Generic helper method for getting data of any kind from DHIS2.
+ * - This can be used to get `DataValueSets`,`events`,`trackers`,`etc.`
+ * @public
+ * @function
+ * @param {string} resourceType - The type of resource to get(use its `plural` name). E.g. `dataElements`, `tracker/trackedEntities`,`organisationUnits`, etc.
+ * @param {RequestOptions} [options] - An optional object containing query, parseAs,and headers for the request
+ * @state {Dhis2State}
+ * @returns {Operation}
+ * @example <caption>Get all data values for the 'pBOMPrpg1QX' dataset</caption>
```
josephjclark, line 30: We don't need so many examples in http.get. Maybe just one? But we do need to think about the main `get()` and have more examples there

packages/dhis2/src/Adaptor.js
```diff
@@ -253,7 +251,7 @@ export function create(resourceType, data, options = {}, callback = s => s) {
     const { location } = response.headers;
     if (location) console.log(`Record available @ ${location}`);
 
-    return handleResponse(response, state, callback);
+    return handleResponse(response, state);
```
josephjclark, line 254: I _think_ - and I'll come back to you on this - that in the main namespace here, we don't include `response` in the result. We just return to `state.data`.

The http methods should surface the response, but the resource helpers should not

packages/dhis2/src/http.js
hunterachieng: @josephjclark  some of the examples focus on fetching the image, and another for the tracker endpoints. Should we have 3 to demonstrate the different usecases?

josephjclark: 3 sounds fine, so long as they're materially different. Use your judgement!

packages/dhis2/src/Adaptor.js
hunterachieng: @josephjclark  What should we do here? I have fixed the other comments

josephjclark:
Sorry @hunterachieng I see you've been waiting for me on this. I'll try and get this unblocked tomorrow.

josephjclark:
Hi @hunterachieng , can we please restore the `get` function and remove the `reponse` from the main `get()`, `create()`, `update()`, `upsert()` etc functions.

basically the http namespace should include `response`, but the other functions should not.

Thanks!

josephjclark:
@hunterachieng something really weird has happened on this diff? The history is all messed up. I think you need to revert that merge.

hunterachieng:
@josephjclark  I have implemented the changes

packages/dhis2/src/Adaptor.js
```diff
@@ -253,6 +248,76 @@ export function create(resourceType, data, options = {}, callback = s => s) {
...
+ * });
+ * @example <caption>Get an image from a trackedEntityInstance.</caption>
+ * get('trackedEntityInstances/qHVDKszQmdx/BqaEWTBG3RB/image', {
+ *   headers:{
+ *       Accept: 'image/*'
+ *   },
+ *   parseAs: 'base64',
+ * });
+ */
+export function get(resourceType, options = {}, callback = s => s) {
```
josephjclark, line 294: Should be no callback here right? Please also check docs and tests

josephjclark: But there should already be get tests right? The ask is to fix (or restore) them, rather than writing more.

hunterachieng: I did restore them. I can add an assertion that `state.response` is `undefined`

josephjclark:
@hunterachieng can you please take the http tests and split them out into a new file `http.test.js`? The test file is very big now and it gets a bit confusing to know which tests you're looking at.

.changeset/silly-bulldogs-complain.md
```diff
@@ -0,0 +1,24 @@
...
+
+```
+get('programs', { orgUnit: 'TSyzvBiovKh', fields: '*' });
+
+```
+
+Do this instead: 
+
+```
+http.get('programs', { orgUnit: 'TSyzvBiovKh', fields: '*' });
```
josephjclark, line 21: I don't think this example will work? Shouldn't this be either
```
http.get('programs', { query: { orgUnit: 'TSyzvBiovKh', fields: '*' } });
```

or:

```
get('programs', { query: { orgUnit: 'TSyzvBiovKh', fields: '*' } });
```

hunterachieng: Fixed

josephjclark (approved):
Great, thank you @hunterachieng 

I'm still confused about the main namespace `get()`. It just doesn't sit well with `http.get()`.

 I'm going to have to work out what to do and raise an new issue for next sprint.

PR #1056, opened 2025-03-13
OpenMRS: support error maps (common: support errors:false)

### LINKED ISSUE #1055: OpenMRS: support error map on http functions
In OpenMRS 4.3,  the http. namespace functions all call out to the common request helper.

Any requests that return an error <=400 will trigger an exception. This cannot be controlled by the user: it's hard-coded behaviour.

The underlying request function supports an error map via the `errors` option. See [http docs](https://docs.openfn.org/adaptors/packages/http-docs#requestoptions)

I want this same behaviour in OpenMRS: I want to do `http.get('concept/1234`, { errors: false })` to suppress all error handling. Then in a callback function, I can look at the response code and do what I want with it.

josephjclark:
## Summary

This PR allows `errors` to be passed to all http functions, which lets users prevent throw-on-error for some or all error codes.

It also adds behaviour to common.http to pass `errors:false` and disable all error checking.

Fixes #1055

Annoyingly there are a lot of formatting diffs in this PR :(

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/openmrs/src/http.js
+++ b/packages/openmrs/src/http.js
@@ -7,6 +7,7 @@ import * as util from './Utils';
  * @property {object} query - An object of query parameters to be encoded into the URL
  * @property {object} headers - An object of all request headers
  * @property {object} body - The request body (as JSON)
+ * @property {object|boolean} errors - Pass `false` to not throw on errors. Pass a map of errorCodes: error messages, ie, `{ 404: 'Resource not found' }`, or `false` to suppress errors for a specific code.
  * @property {string} [parseAs='json'] - The response format to parse (e.g., 'json', 'text', or 'stream')
  */
 
@@ -15,9 +16,9 @@ import * as util from './Utils';
  * @example <caption>GET request with a URL query</caption>
  * http.request("GET",
  *   "/ws/rest/v1/patient/d3f7e1a8-0114-4de6-914b-41a11fc8a1a8", {
- *    query:{ 
- *       limit: 1, 
- *       offset: 20 
+ *    query:{
+ *       limit: 1,
+ *       offset: 20
  *    },
  * });
  * @function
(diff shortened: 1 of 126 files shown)


packages/openmrs/src/http.js
```diff
@@ -56,22 +57,37 @@ export function request(method, path, options = {}) {
...
+ * @example <caption>Don't throw if OpenMRS returns a 404 error code</caption>
+ * http.get(
+ *  "/ws/rest/v1/patient",
+ *  {
+ *    errors: { 404: false }
+ *  }
+ * )
  * @param {string} path - path to resource
  * @param {RequestOptions} [options={}] - An object containing query params and headers for the request
  * @returns {operation}
```
mtuchi, line 70: I think we need to documentation for final state for each `http.namespace` function 

```
/**
 * State object
 * @typedef {Object} HttpState
 * @property data - the parsed response body
 * @property response - the response from the HTTP server, including headers, statusCode, body, etc
 * @property references - an array of all previous data objects used in the Job
 **/
```

packages/openmrs/src/http.js
```diff
@@ -126,14 +151,23 @@ export function post(path, data, options = {}) {
...
+    );
 
     return util.prepareNextState(state, response);
-  }
+  };
 }
 
-export { _delete as delete };
\ No newline at end of file
+export { _delete as delete };
```
mtuchi, line 173: Just a question, Not important to change this now. But why not `export function delete()`  why do we have to keep `_delete` ? It's not used anywhere right ?

mtuchi (commented):
Hiya @josephjclark left some few comments. Let me know if you need help in making those changes. 

And just a realization note. When i saw changes in common adaptor it reminded me it's important to write unit test in adaptors that uses common because for instance. This PR will update all adaptors that uses `common: workspace*`, So if the implementation of that adaptor uses `util.http` from common, unit test should be able to capture any regression introduced with this new release. Of-course it only make sense if common had a major release but still it's a reminder on how important unit tests are 

cc @hunterachieng

packages/openmrs/src/http.js
```diff
@@ -7,6 +7,7 @@ import * as util from './Utils';
  * @property {object} query - An object of query parameters to be encoded into the URL
  * @property {object} headers - An object of all request headers
  * @property {object} body - The request body (as JSON)
+ * @property {object|boolean} errors - Pass `false` to not throw on errors. Pass a map of errorCodes: error messages, ie, `{ 404: 'Resource not found' }`, or `false` to suppress errors for a specific code.
```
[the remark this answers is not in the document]
josephjclark: I didn't like the common description. Also we're trying to say 3 slightly complicated things in a short sentence:
1) You can pass `errors: false` to disable all error handling
2) You can pass`errors: { 404: 'message' }` to set a custom error message (or throw for a code <400)
3) You can pass `errors: { 404: false }` to not throw for a particular code

It's just really hard to document.

What we should actually do is say `configure error handling rules, see DOCS` and provide a link to a longer explanation. I'll raise an issue for this but not address it in this PR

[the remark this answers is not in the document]
josephjclark: https://github.com/OpenFn/docs/issues/648

packages/openmrs/src/http.js
josephjclark: This is done in openMRS v5 (coming soon!). We won't duplicate the work here.

packages/openmrs/src/http.js
josephjclark: `delete` is a keyword, so you can't declare `function delete()` without syntax errors

mtuchi: Noted 👍🏽

packages/openmrs/src/http.js
mtuchi: Got you 👍🏽

PR #1059, opened 2025-03-14
dhis2: remove axios and fix documentation

### LINKED ISSUE #975: dhis2: remove axios

hunterachieng:
## Summary

Remove `axios` from `dhis2` and update the documentation, support `base64` and update tests. 

Fixes #975 

## Details
- Remove `axios` from `dhis2`.
- Remove interceptor.
- Allow users to pass in their own `headers` and `parseAs`.
- Support `base64` format in the `get` function

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/dhis2/test/integration.js
```diff
@@ -252,7 +252,7 @@ describe('Integration tests', () => {
 
     it('should get dataValueSets matching the query specified', async () => {
       const finalState = await execute(
-        get('dataValueSets', {
+        http.get('dataValueSets', {
```
josephjclark, line 255: Why is this http.get? There's no http namespace on this branch?

packages/dhis2/src/Utils.js
```diff
@@ -14,9 +19,9 @@ export const CONTENT_TYPES = {
   xls: 'application/vnd.ms-excel',
 };
 
-export function buildUrl(urlString, hostUrl, apiVersion) {
+export function buildUrl(urlString, apiVersion) {
```
josephjclark, line 22: We have `buildUrl` and we have `generateUrlPath`.

Do we _really_ need both? I'd love to consolidate these.

I think `buildUrl` is only called from inside `geneateUrlPath`, so we can probably just move the functionality inside?

packages/dhis2/src/Utils.js
```diff
@@ -53,8 +62,13 @@ export function ensureArray(data, key) {
   return Array.isArray(data) ? { [key]: data } : { [key]: [data] };
 }
 
-export function generateUrl(configuration, options, resourceType, path = null) {
-  let { hostUrl, apiVersion } = configuration;
+export function generateUrlPath(
```
josephjclark, line 65: I think all this is _actually_ doing is maybe adding a version prefix to the path.

Like, it takes a path and just sticks a version number in front of it.

So maybe a better function name would be `prefixVersionToPath()` ?

packages/dhis2/src/Adaptor.js
```diff
@@ -500,31 +452,41 @@ export function update(
...
  * get('tracker/relationships', {
- *   trackedEntity:['F8yKM85NbxW'],
+ *   query: { trackedEntity:['F8yKM85NbxW'] }
+ * });
+ * @example <caption>Get an image from a trackedEntityInstance.</caption>
+ * get('trackedEntityInstances/qHVDKszQmdx/BqaEWTBG3RB/image', {
+ *   headers:{
+ *       Accept: 'image/*'
+ *   },
+ *   parseAs: 'text',
```
josephjclark, line 462: Didn't we say we'd support `parseAs: 'base64'` ?

That means we can drop the `asBase64` key completely

packages/dhis2/src/Utils.js
```diff
@@ -40,9 +45,13 @@ export function selectId(resourceType) {
 
 // Write a unit test for this one
```
josephjclark, line 46: Let's remove this comment.

I don't really think `handleResponse` should have its own unit test at this point.

packages/dhis2/src/Utils.js
```diff
@@ -66,8 +80,51 @@ export function generateUrl(configuration, options, resourceType, path = null) {
...
+  const { hostUrl } = configuration;
+  const { method, path, options = {}, data = {} } = requestData;
+
+  const {
+    headers = { 'content-type': 'application/json' },
+    query = {},
+    parseAs = 'json',
+  } = options;
+
+  console.log(`Sending ${method} request to ${path}`);
```
josephjclark, line 113: Do we want this log? because common HTTP should log the response for us (whether it failed or not). So this feels like spam.

packages/dhis2/src/Adaptor.js
```diff
@@ -5,14 +5,23 @@ import _ from 'lodash';
...
   shouldUseNewTracker,
   ensureArray,
+  generateUrlPath,
+  request,
 } from './Utils';
-import { request } from './Client';
+
+/**
+ * Options object
+ * @typedef {Object} RequestOptions
```
josephjclark, line 19: This should include `apiVersion` as well, no? So that we can override the API version for a specific request.

Are there any other options we need to document? Stuff used by dhsi2 but not by the common helper?

packages/dhis2/src/Adaptor.js
```diff
@@ -273,16 +223,23 @@ export function create(resourceType, data, options = {}, callback = s => s) {
...
+          resolvedOptions,
+          resolvedResourceType
+        ),
+        options: resolvedOptions,
         data: resolvedData,
-        ...requestConfig,
       });
     }
 
     return promise.then(result => {
```
josephjclark, line 237: We should refactor `result` to `response`

packages/dhis2/src/Adaptor.js
```diff
@@ -273,16 +223,23 @@ export function create(resourceType, data, options = {}, callback = s => s) {
...
-      const details = `with response ${JSON.stringify(result.data, null, 2)}`;
+      const details = `with response ${JSON.stringify(
+        result?.body?.message,
```
josephjclark, line 239: common .request _always_ returns us a response object, so we don't need the question mark here. Also note that if `result` really is undefined, this function will throw a bit later down on `const { location } = result.headers;`

packages/dhis2/src/Adaptor.js
```diff
@@ -273,16 +223,23 @@ export function create(resourceType, data, options = {}, callback = s => s) {
...
+        null,
+        2
+      )}`;
       console.log(`Created ${resolvedResourceType} ${details}`);
```
josephjclark, line 243: Do we really want to log this btw? Maybe we should just log the id?

Can you check the issues to see if anyone has complained about dhis2 logging too much data, and maybe ask Aleksa and Mtuchi about this?

My instinct is that we should **not**  be force-logging data

josephjclark: It might technically be another issue to fix this, but because I've seen it now I want to ask the question here first

packages/dhis2/src/Adaptor.js
```diff
@@ -634,24 +605,54 @@ export function upsert(
       // NOTE: that these parameters are all expanded by the `get`, `create`, and
       // `update` functions used inside this composed "upsert" function.
       console.log(`Preparing upsert via 'get' then 'create' OR 'update'...`);
-      promise = get(
-        resourceType,
-        query,
-        options
-      )(state).then(resp => {
-        const resources = resp.data[resourceType];
+      promise = request(configuration, {
```
josephjclark, line 608: Can we refactor this, and other similar structures, to use async/await? While we're here?

packages/dhis2/src/Adaptor.js
```diff
@@ -634,24 +605,54 @@ export function upsert(
...
+          ...resolvedOptions,
+          query: {
+            ...resolvedQuery,
+          },
+        },
+      }).then(resp => {
+        const resources = resp.body[resourceType];
         if (resources.length > 1) {
           throw new RangeError(
             `Cannot upsert on Non-unique attribute. The operation found more than one records for your request.`
```
josephjclark, line 625: This is such a poor error message.

Maybe we can do use `util.throwError()` to provide a simpler, cleaner explanation of the error and provide a fix.

This should probably be spun out into a new issue

.changeset/brown-singers-sit.md
```diff
@@ -0,0 +1,26 @@
+---
+'@openfn/language-dhis2': major
```
josephjclark, line 2: We've also removed the discover function

Can you please check this PR really carefully and make sure this list is complete?

.changeset/brown-singers-sit.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+- Remove axios from `dhis2`.
+- Remove interceptor.
+- Allow users to pass in their own `headers` and `parseAs`.
+- Support `base64` format in the `get` function
+
+### Migration Guide
+
+- The `get` function has major changes in the parameters accepted.
```
josephjclark, line 12: I will come back to this migration guide right at the end, but just to help my memory later, can you rephase to `the get function signature has changed. In particular, note that query parameters must be passed to request.options`

Don't worry about the language, I'll rework it later. It just helps me remember. Thanks!

josephjclark (changes requested):
Pretty good Hunter - just a few things I'd like you to take another look at

I want to merge this PR relatively quickly and I'll do a deeper review and QA on the final branch once we've done the basic changes.

packages/dhis2/src/Adaptor.js
hunterachieng: I have added it. We are already checking for it in the `prefixVersionToPath()` function

josephjclark:
Just ran the integration tests and half of them fail. I suspect that's a problem with the integration tests, not with the code.

josephjclark (approved):
Great, merging this down to the epic. Thanks @hunterachieng

Issue #687, opened 2024-07-22
Refactor `salesforce` adaptor to use client Instead of `state.connection`


mtuchi:
### Description:
Currently, our Salesforce adapter uses `state.connection` to establish the session and removes `state.connection` at the end of execution. To improve the structure and maintainability of the code, we want to refactor the adaptor to use a client setup similar to what we have done with our Hive and GoogleSheets adaptor. This involves the following changes:

1. Remove `state.connection`.
2. Introduce a client or session variable (e.g., `let client = null;` or `let session = null;`).
3. Set up the client using a new function `createConnection()`.
4. Destroy the client using a new function `removeConnection()`.

### Tasks:
1. **Remove `state.connection`:**
   - Identify all instances where `state.connection` is used.
   - Refactor the code to remove `state.connection`.

2. **Introduce client variable:**
   - Define a new client or session variable (e.g., `let client = null;`).

3. **Update `createConnection()` function:**
   - Implement a function `createConnection()` that initializes the client.
   - Ensure the function handles all necessary configurations and authentication.

4. **Create `removeConnection()` function:**
   - Implement a function `removeConnection()` that properly closes and cleans up the client session.
   - Ensure the function handles any necessary cleanup and error handling.

5. **Refactor code to use client:**
   - Update all relevant parts of the adapter to use the new client variable.
   - Replace instances of `state.connection` with the appropriate client methods and properties.

6. **Testing:**
   - Thoroughly test the refactored adaptor to ensure it functions correctly.
   - Ensure all existing functionality is preserved and no regressions are introduced.

### References:
- Example from PostgreSQL adapter:
  ```javascript
  let client = null;

  function createConnection() {
    // Setup client
  }

  function removeConnection() {
    // Destroy client
  }
  ```

### Related Issues/PRs:
- #662

josephjclark:
The unit tests currently rely on `state.connection` to provide a mock connection handler.

What you'll need to do is provide something like a `setMockClient(client)` private function in the adaptor. A bit like how we handle mock connections in common http.

Then the tests will call `setMockClient` each time and provide mock implementations to satisfy the tests

This is sort of what I want to do in the mockable tests epic, but I agree that we need to fix salesforce before we pick that epic up.

mtuchi:
Made good progress on this, but integration test are failing with an error saying `Only absolute URLs are supported`.  I will investigate further tomorrow

PR #1063, opened 2025-03-18
Update salesforce to use `connection` client 

mtuchi:
## Summary

Replace `state.connection` with a global `connection` client and update unit tests

**Fixes**
-  #687
- #1067

## Details
- Update `configuration-schema` to default `loginUrl` to `https://login.salesforce.com` and `apiVersion` to `50.0`. Because of defaultConnectionConfig in V3. See [V3 connection.ts:L109](https://github.com/jsforce/jsforce/blob/3.6.6/src/connection.ts#L109), [V3 connection.ts:L111](https://github.com/jsforce/jsforce/blob/3.6.6/src/connection.ts#L111)
- Removed `util.createBasicConnection()`.
- Removed `util.createTokenConnection`
- Add `connect` function for connecting to salesforce using basic or access_token
- Add `setMockConnection()` function, which make it easy to setup fake connection
- Update unit test to use `setMockConnection()` for setting up fake connection 
- Update `execute()` function to `createConnection` and removed `removeConnection` function
- Update all Adaptor function to use the global `connection` instead of `state.connection`
- Remove `lodash` and the use of `flatten` in execute function

## AI Usage

know!):

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


mtuchi:
Hiya @josephjclark , this one is ready for review. I can't express enough how grateful i am with the unit tests. They helped me a lot on getting this done 🙌🏽

packages/salesforce/src/Adaptor.js
```diff
@@ -104,11 +128,10 @@ export function execute(...operations) {
   };
 
   return state => {
-    util.createConnection(state);
-    connection = util.getConnection();
     return commonExecute(
       util.loadAnyAscii,
-      ...operations
+      createConnection,
+      ...flatten(operations)
```
josephjclark, line 130: Why did we introduce flatten here? it wasn't here before?

packages/salesforce/src/util.js
```diff
@@ -1,15 +1,13 @@
...
-let connection = null;
-
-async function basicAuth(state) {
+export const basicAuth = async configuration => {
   const { loginUrl, username, password, securityToken, apiVersion } =
-    state.configuration;
+    configuration;
 
-  connection = new Connection({ loginUrl, version: apiVersion });
+  const conn = new Connection({ loginUrl, version: apiVersion });
```
josephjclark: I'm not really sure about this design

First off, `basicAuth` doesn't sound like a function that'll create a  Salesforce connection.

Secondly, I was surprised to see new Connection in utils at all.

I would simply have a `connect(configuration)` function which looks at config, creates the right sort of connection. and throws with a clear message if invalid configuration is provided.

The `connect` function could live in Adaptor or Utils, both are fine.

josephjclark: It's kind of telling that if you look at `createConnection` in adpator.js,  the only lines of code that actually do anything are calling into `util.js`. So I think these functions all need to be in the same file.

packages/salesforce/src/util.js
```diff
@@ -18,64 +16,33 @@ async function basicAuth(state) {
...
-export function removeConnection(state) {
-  connection = null;
-  return state;
-}
+  return conn;
+};
 
-export function getConnection() {
+export const checkConnection = async connection => {
+  // console.log(typeof (await connection));
```
josephjclark: This log line needs removing

packages/salesforce/src/util.js
```diff
@@ -18,64 +16,33 @@ async function basicAuth(state) {
...
- */
-export function removeConnection(state) {
-  connection = null;
-  return state;
-}
+  return conn;
+};
 
-export function getConnection() {
+export const checkConnection = async connection => {
```
josephjclark, line 41: I think this logic should all be part of the new `connect()` function I've recommended. It's only called from one place right now and it doesn't really DO very much? So I think I'd like to see it assimilated elsewhere

packages/salesforce/test/integration.js
```diff
@@ -13,7 +13,7 @@ import {
 const state = { configuration };
 
 describe('Integration tests', () => {
-  describe.skip('bulk', () => {
+  describe('bulk', () => {
```
josephjclark: Are these tests supposed to be restored in this PR? 

They don't run in CI right, just locally? So it's probably OK to enable them. I'm just a bit surprised at the diff (and conscious that you're looking at tests in another PR)

packages/salesforce/src/Adaptor.js
mtuchi: @josephjclark, flatten was there before. I just moved `createConnection` from util.js to Adaptor.js and removed `removeConnection` function

packages/salesforce/src/util.js
mtuchi: Updated `checkConnection` to only be used internally by `basicAuth` and `tokenAuth` functions

packages/salesforce/test/integration.js
mtuchi: Yes the don't run in CI, i enabled them to make sure test is passing. They should be enabled

packages/salesforce/src/util.js
mtuchi: This is outdated, it was remove long ago

packages/salesforce/src/util.js
mtuchi: I have moved all function for handling connection in Adaptor.js

packages/salesforce/src/Adaptor.js
josephjclark: Ah then I'm getting confused.

Ok, I've looked over this code. You can remove `flatten` completely and close #1067. Let's just do this here, it's a simple change and it's getting merged into an epic anyway

mtuchi: @josephjclark i have removed the use of `flatten` and deleted `lodash` package

josephjclark: Great, thanks!

packages/salesforce/src/Adaptor.js
```diff
@@ -82,6 +81,73 @@ import flatten from 'lodash/flatten';
...
+ * @param {State} state - Runtime state.
+ * @returns {State}
+ */
+const connect = async state => {
+  if (connection) {
+    return state;
+  }
+  const { configuration } = state;
+  configuration.access_token
+    ? await tokenAuth(configuration)
```
josephjclark, line 142: hmm, sorry, i was just about to merge but this ternary just caught my eye and it's needling me. I think if I let this go I'll regret it forever.

IMO this is a "naughty ternary" because it doesn't return anything. This should really be an if statement.

And something about the structure and naming of this is bothering me :thinking: 

I would either:
1) Move _all_ the code inside this connect function, so that you have a big connect function. I don't really understand why `basicAuth` and `tokenAuth` need to be in their own function. We only call them from one place and presumably we're not unit testing them. If you move them inside if/else branches in connect, you can also move `checkConnection` inside (which is what I was getting at before). And I think that simplifies all the code.
2) Rename `basicAuth` to `connectWithBasicAuth` (because this function _does_ something, it really needs a verb in the name) and `tokenAuth` to `connectWithTokenAuth`, then have those functions accept state and return state. Now connect can end with `return config.access_token? connectWithBasicAuth(state) : connectWithTokenAuth(state)`, which at least feels like a cleaner ternary and a nicer way to end the function.

I haven't seen both approaches written down but my instinct says option 1.

What do you think?

packages/salesforce/src/Adaptor.js
mtuchi: @josephjclark just a heads up, #1070 is rebased against `687-use-client` branch. So that PR should be merged before this one to avoid conflict resolution

packages/salesforce/src/Adaptor.js
mtuchi: Options 1 definitely, it makes the whole implementation of `connect` clean, let me work on that

mtuchi: @josephjclark just a heads up, #1070 is rebased against `687-use-client` branch. So that PR should be merged before this one to avoid conflict resolution

josephjclark (approved):
Thank you @mtuchi , I think this is much cleaner :raised_hands:

PR #1091, opened 2025-03-21
Release: dhis2 7.0

### LINKED ISSUE #919: Epic: dhis2 7.0
Epic issue to track work on the next major dhis2 version

## Notes

- Remove callbacks
- move attrs to utils? `util.attr(attribute, value)` is pretty annoying though. Should we have a dhis2 builder API? `b.attr()`?
- `create` should not include HTTP semantics
- `create` should have fewer examples?
- `destroy` is not "a generic helper function"

josephjclark:
Epic branch for next dhis2 release

Closes #919


packages/common/src/util/http.js
```diff
@@ -245,6 +245,10 @@ async function readResponseBody(response, parseAs) {
       return undefined;
     }
 
+    if (response.statusCode === 204) {
```
josephjclark, line 248: I have a feeling we need to remove this (and the associated changeset).  `main` has had a bunch of changes in this area and this branch should be stable without this diff

josephjclark:
When the new http namespace is merged into here I want to take a really close look at the `get()` function

We changed the `get()` signature when removing axios (and didn't document the breaking change). Perhaps it made sense at the time but I think we made it a more http-like function. But now that we've got a http namespace, the ORIGINAL pre-7.0 `get` function looks much better to me

I'm not quite sure where I stand on it. I need to do a deep compare and have a little think about it.

packages/dhis2/src/tracker.js
```diff
@@ -0,0 +1,136 @@
...
+      path: util.prefixVersionToPath(
+        state.configuration,
+        {
+          ...resolvedOptions,
+          resolvedStrategy,
+        },
+        'tracker'
+      ),
+      options: {
+        apiVersion,
```
josephjclark, line 74: @hunterachieng I don't think `apiVersion` should be passed into the request options like this? It's not used, and so it's really confusing to pass it here.

PR #1092, opened 2025-03-21
`salesforce` update query function

### LINKED ISSUE #1030: `salesforce` refactor `query` function
**Description**
The `query` function auto fetching implementation will no longer be needed on jsforcev3, Once we upgrade the adaptor to use the latest version of jsforcev3, See #1028, We should update the query function to use `conn.query` [See docs here](https://jsforce.github.io/document/#query) also we should add support for queryOptions which will make it easy to auto fetch records when we hit the maximum fetch limit. [See queryOptions here](https://jsforce.github.io/jsforce/types/query.QueryOptions.html)

mtuchi:
## Summary

Refactor query function in salesforce adaptor

Fixes #1030

## Details

- Update query function to use thew new jsforce v3 connection. The new `connection.query()` support query options for fetching more data we reached query limit hence removing the current implementation of `autoFetch`
- Update query options [documentation link]( https://jsforce.github.io/jsforce/types/query.QueryOptions.html) 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/salesforce/src/Adaptor.js
+++ b/packages/salesforce/src/Adaptor.js
@@ -75,12 +75,6 @@ import * as util from './util';
  * @property {integer} [pollInterval=3000] - Polling interval in milliseconds.
  * */
 
-/**
- * @typedef {Object} QueryOptions
- * @public
- * @property {boolean} [autoFetch=false] - When true, automatically fetches next batch of records if available.
- * */
-
 let connection = null;
 
 /**
@@ -634,8 +628,8 @@ export function post(path, data, options = {}) {
  * @example <caption>Query patients by Health ID using a lazy state reference</caption>
  * query(`SELECT Id FROM Patient__c WHERE Health_ID__c = '${$.data.healthId}'`);
  * @function
- * @param {(string|function)} query - A SOQL query string or a function that returns a query string. Must be less than 4000 characters in WHERE clause
- * @param {QueryOptions} [options] - Optional configuration for the query operation
+ * @param {string} query - A SOQL query string. Must be less than 4000 characters in WHERE clause
+ * @param {object} [options] - Optional query options, {@link https://jsforce.github.io/jsforce/types/query.QueryOptions.html query options}
  * @state {SalesforceState}
  * @property data - Array of result objects of the form <code>\{ done, totalSize, records \}</code>
  * @returns {Operation}
@@ -648,63 +642,44 @@ export function query(query, options = {}) {
       options
     );
     console.log(`Executing query: ${resolvedQuery}`);
-    const autoFetch = resolvedOptions.autoFetch || resolvedOptions.autofetch;
+    const { autoFetch = true, maxFetch = 10000 } = resolvedOptions;
 
     if (autoFetch) {
-      console.log('autoFetch is enabled: all records will be downloaded');
+      console.log(
+        `autoFetch is enabled: A maximum of ${maxFetch} records will be downloaded`
+      );
     }
 
-    const result = {
-      done: true,
-      totalSize: 0,
-      records: [],
-    };
-
-    const processRecords = async res => {
-      const { done, totalSize, records, nextRecordsUrl } = res;
-
-      result.done = done;
-      result.totalSize = totalSize;
-      result.records.push(...records);
-
-      if (!done && !autoFetch && nextRecordsUrl) {
-        result.nextRecordsUrl = nextRecordsUrl;
-      }
-      if (!done && autoFetch) {
-        console.log('Fetched records so far:', result.records.length);
-        console.log('Fetching next records...');
-
-        try {
-          const newResult = await connection.request({ url: nextRecordsUrl });
-          await processRecords(newResult);
-        } catch (err) {
+    let result = {};
+    const records = [];
+    await new Promise((resolve, reject) => {
+      result = connection
+        .query(resolvedQuery)
+        .on('record', record => {
+          records.push(record);
+        })
+        .on('end', () => {
+          console.log('total in database : ' + result.totalSize);
+          console.log('total fetched : ' + result.totalFetched);
+          resolve();
+        })
+        .on('error', err => {
           const { message, errorCode } = err;
           console.error(`Error ${errorCode}: ${message}`);
-          throw err;
-        }
-      }
+          throwError('QUERY_ERROR', { message, errorCode });
+          reject();
+        })
+        .run(resolvedOptions);
+    });
+
+    const allResults = {
+      done: result.totalFetched === result.totalSize,
+      records,
+      totalSize: result.totalSize,
+      totalFetched: result.totalFetched,
     };
 
-    try {
-      const qResult = await connection.query(resolvedQuery);
-      if (qResult.totalSize > 0) {
-        console.log('Total records:', qResult.totalSize);
-        await processRecords(qResult);
-        console.log('Done ✔ retrieved records:', result.records.length);
-      } else {
-        console.log('No records found.');
-      }
-    } catch (err) {
-      const { message, errorCode } = err;
-      console.log(`Error ${errorCode}: ${message}`);
-      throw err;
-    }
-
-    console.log(
-      'Results retrieved and pushed to position [0] of the references array.'
-    );
-
-    return composeNextState(state, result);
+    return composeNextState(state, allResults);
   };
 }
 
--- a/packages/salesforce/test/Adaptor.test.js
+++ b/packages/salesforce/test/Adaptor.test.js
@@ -99,7 +99,7 @@ describe('Adaptor', () => {
     });
   });
 
-  describe('query', () => {
+  describe.skip('query', () => {
     it('should properly pass the query string', async () => {
       const fakeConnection = {
         query: function () {
--- a/packages/salesforce/ast.json
+++ b/packages/salesforce/ast.json
@@ -682,30 +682,21 @@
           },
           {
             "title": "param",
-            "description": "A SOQL query string or a function that returns a query string. Must be less than 4000 characters in WHERE clause",
+            "description": "A SOQL query string. Must be less than 4000 characters in WHERE clause",
             "type": {
-              "type": "UnionType",
-              "elements": [
-                {
-                  "type": "NameExpression",
-                  "name": "string"
-                },
-                {
-                  "type": "NameExpression",
-                  "name": "function"
-                }
-              ]
+              "type": "NameExpression",
+              "name": "string"
             },
             "name": "query"
           },
           {
             "title": "param",
-            "description": "Optional configuration for the query operation",
+            "description": "Optional query options, {@link https://jsforce.github.io/jsforce/types/query.QueryOptions.html query options}",
             "type": {
               "type": "OptionalType",
               "expression": {
                 "type": "NameExpression",
-                "name": "QueryOptions"
+                "name": "object"
               }
             },
             "name": "options"
--- a/packages/salesforce/test/integration.js
+++ b/packages/salesforce/test/integration.js
@@ -13,6 +13,25 @@ import {
 const state = { configuration };
 
 describe('Integration tests', () => {
+  describe('query', () => {
+    it('should return 1000 records', async () => {
+      const { data } = await execute(
+        query('SELECT Id, Name FROM Account LIMIT 1000')
+      )(state);
+
+      expect(data.done).to.eq(true);
+      expect(data.totalSize).to.eq(1000);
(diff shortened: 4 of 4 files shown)


mtuchi:
I will wait for #1070 to be merged so that in can fix unit test for `query` function

mtuchi:
This is almost done, i just need to re-think the query options documentation. Because by default `autoFetch: false` and `maxFetch: 10000` and jsdoc doesn't have good documentation. They have note section that  says `When maxFetch option is not set, the default value (10,000) is applied.` [See docs here](https://jsforce.github.io/document/#using-soql).

I would like to point users to this [query opts docs](https://jsforce.github.io/jsforce/types/query.QueryOptions.html) to see supported options, but it doesn't say the default values for each options

mtuchi:
Hiya @hunterachieng please have a look at changes when you got time

packages/salesforce/test/Adaptor.test.js
```diff
@@ -87,172 +87,172 @@ describe('Adaptor', () => {
         .catch(done);
     });
   });
-
   describe('query', () => {
+    // Common test responses and mocks can be moved to the top
```
hunterachieng, line 91: Can we remove this comment?

.changeset/sweet-ties-juggle.md
```diff
@@ -0,0 +1,7 @@
+---
+'@openfn/language-salesforce': minor
+---
+
+Modernize `query()` implementation using jsforce v3, adding support for
+additional query options [ `maxFetch`, `headers`, `responseTarget` and
```
hunterachieng, line 6: In your code, I see we are only using `maxFetch` and `autofetch`. Can one still pass the other options as mentioned here?

mtuchi: Yes, you can pass the other options

packages/salesforce/src/Adaptor.js
```diff
@@ -528,77 +522,40 @@ export function insert(sObjectName, records) {
...
-      query,
-      options
-    );
+    const [
+      resolvedQuery,
+      resolvedOptions = { autoFetch: false, maxFetch: 10000 },
+    ] = expandReferences(state, query, options);
     console.log(`Executing query: ${resolvedQuery}`);
-    const autoFetch = resolvedOptions.autoFetch || resolvedOptions.autofetch;
+    const { autoFetch, maxFetch } = resolvedOptions;
```
josephjclark, line 538: Is autoFetch actually doing anything now?

Presumably if you pass maxFetch: 10m, it'll just download 10m records regardless of whether autofetch is enabled? This, by the way, is the correct pagination behaviour. Ask for a million records and you should be given a million records.

packages/salesforce/src/Adaptor.js
```diff
@@ -528,77 +522,40 @@ export function insert(sObjectName, records) {
...
-    console.log(
-      'Results retrieved and pushed to position [0] of the references array.'
+    const { records, totalSize, done } = await connection.query(
+      resolvedQuery,
+      resolvedOptions
     );
+    const totalFetched = records.length;
 
-    return composeNextState(state, result);
+    if (records.length > 0) {
```
josephjclark, line 552: If there are no records, we say nothing? Does that seem correct?

packages/salesforce/src/Adaptor.js
```diff
@@ -528,77 +522,40 @@ export function insert(sObjectName, records) {
...
+      console.log('total in database : ' + totalSize);
+      console.log('total fetched : ' + totalFetched);
+    }
+    const allResults = { done, records, totalSize, totalFetched };
```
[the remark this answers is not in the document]
josephjclark: I'm not neccssarily asking for a change here, just sounding out the question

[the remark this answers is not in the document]
mtuchi: If `autoFetch: false` and we have 8000 records in the database
```
done: false,
totalFetched: 2000, 
records: [{}] // 2000 objects
totalSize: 8000,
nextRecordsUrl: 'url-for-next-batch'
```

If `autoFetch: true`
```
done: true,
totalFetched: 8000, 
records: [] // 8000 records
totalSize: 8000,
```

I can remove `totalFetched` and leave the log only that says `total fetched : records.length`. I thought it's useful because `query` has a limitation of `2000 records` if `autFetch: false` and `10000 max records` if `autoFetch: true`  and `maxFetch` value is not specified

packages/salesforce/src/Adaptor.js
mtuchi: Okay that make sense, but i wonder if we need a better way of querying for more results. Because making the query by default autoPage make sense but you still have limitation of 10000 max records by default, if you need more you just need to adjust `maxFetch` which still is not a guarantee if it will return all results

mtuchi: I think `autoFetch` is still useful for when user wants to get the default maximum number of results and get the `nextRecordsUrl` incase they need to use it in their job code

packages/salesforce/src/Adaptor.js
```diff
@@ -528,77 +522,40 @@ export function insert(sObjectName, records) {
...
-      'Results retrieved and pushed to position [0] of the references array.'
+    const { records, totalSize, done } = await connection.query(
+      resolvedQuery,
+      resolvedOptions
     );
+    const totalFetched = records.length;
 
-    return composeNextState(state, result);
+    if (records.length > 0) {
+      console.log('total in database : ' + totalSize);
```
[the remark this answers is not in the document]
mtuchi: If it only say `totalFetched` and it doesn't say how many are left then i don't have anything in the logs that suggest i haven't fetch all records.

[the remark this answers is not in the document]
mtuchi: `query` has a limit of `2000 records` if `autoFetch: false` and `10000 records` if `autoFetch: true`.

packages/salesforce/src/Adaptor.js
josephjclark: But autofetch isn't doing anything right now. So if you explicitly want _all_ records, what do you do?

Also we haven't documented the 10k default? It's probably worth us duplicating documentation on options we think are going to be super common. Like `maxFetch` right now (which in other adaptors we call `limit`, so maybe we should think about that) is basically essential to large queries

packages/salesforce/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Ok. Let me go back and remind myself about this data structure - I seem to remember talking about it before.

My instinct is that we should just have an API that says "get me all records" or "get me this specific set of records", and nothing else. And in that API you'd just have a flat `state.data` results array.

I don't know if we've talked about this, but if we were to expose the jsforce object directly, we'd have a lot of license to have more opinionated APIs for 90% of use-case, and point users to the client for the edge-cases that aren't covered by by this.

I suppose that depends on how common it is to use `nextRecordsUrl`

josephjclark:
@mtuchi how about this:

We make the signature `query(q, { limit })`. We always autoFetch, and we explicitly default the limit to 10k. Pass `limit: false` or `limit:Infinity` to download everything.

No other jsforce options are supported.

We return a result array to `state.data`.

We return the `{ done, totalSize, records, nextRecordsUrl }` object returned by jsforce to `state.response`

Users wanting to implement their own pagination need to do `http.get($.response.nextRecordsUrl)`. I don't really want them to do this (because it's too hard in openfn). They should be using a cursor to query for the correct range of data, and processing that entire range in the step.

This is not a jsforce query wrapper. This is a salesforce query operation.

This would be a breaking change and should be released in the next major (not 6), along with the bulk API changes.

mtuchi:
Okay got you @josephjclark, I will work on your suggestion and fix the conflicts on the next sprint

mtuchi:
Hiya @josephjclark can we change `limit` option to `maxFetch` [1:1 map to jsforce] or to `max` [Openfn only opts]
Because For the case of `query()` you could write a query specifying the limit for example 
```js
query('select Name from Account LIMIT 10')
```
So having the `limit` option feels like it will introduce confusion

josephjclark:
@mtuchi That's cool - let's call it `max` (I feel no obligation to map to jsforce)

Agreed on removing `records` from state.response!

mtuchi:
> @mtuchi That's cool - let's call it `max` (I feel no obligation to map to jsforce)
> 
> Agreed on removing `records` from state.response!

Hiya @josephjclark done ✅

.changeset/sweet-ties-juggle.md
```diff
@@ -0,0 +1,47 @@
...
+
+```js
+// Old way - using autoFetch
+query('select name from account', { autoFetch: true });
+// Result: state.data = { done, totalSize, records: [] }
+```
+
+**After**
+
+To fetch all records now, use the `max: false` option:
```
josephjclark, line 39: Can we document this as `max: Infinity`? I think that's easier to comprehend

.changeset/sweet-ties-juggle.md
```diff
@@ -0,0 +1,47 @@
...
+
+- Modernize `query()` implementation using jsforce v3
+- Remove `autoFetch` option from `query()` function
+- Add `max` option to `query()` function
+- Change query result structure:
+  - `state.data` now contains only the array of records
+  - Query metadata (`done`, `totalSize`, `nextRecordsUrl`) moved to
+    `state.response`
+
+### Migration Guide
```
josephjclark, line 13: Fantastic docs, thank you :pray:

packages/salesforce/src/Adaptor.js
```diff
@@ -515,90 +509,54 @@ export function insert(sObjectName, records) {
...
  * Note that in an event of a query error, error logs will be printed but the operation will not throw the error.
  *
  * The Salesforce query API is subject to rate limits, {@link https://sforce.co/3W9zyaQ learn more here}.
- *
  * @public
  * @example <caption>Run a query and download all matching records</caption>
- * query('SELECT Id FROM Patient__c', { autoFetch: true });
+ * query('SELECT Id FROM Patient__c', { max: false });
+ * @example <caption>Run a query and limit records</caption>
+ * query('SELECT Id From Account Limit 10');
```
josephjclark, line 519: Oh hey this is a great point, I hadn't considered that users can limit in the query itself.

This makes me think that we should actually default max to infinity, which would avoid a conflict on limit and max

Salesforce users used to running big queries are likely to limit as standard, and don't need our API getting in the way. They probably want control from the query, right?

josephjclark (commented):
Looks great but your excellent examples  have raised a question in my mind! Can we tackle that before deciding next steps?

packages/salesforce/src/Adaptor.js
mtuchi: If we default `max: Infinity` we are partially doing `bulkQuery()` job. Because the only time you will need to use bulkQuery will be when you're query string exceed 4000 characters. otherwise just use `query()`.

There is also a difference between using `max: 100` to `select id from account limit 100`. 

With `max: 100` you will get 100 records + metadata for the next records. 
But  will `limit 100` in a query` you will only get `100 records`

mtuchi: I like the idea of defaulting `max: Infinity` but i am concerned if user have 1million records. Do we really want them to use `query()` function ? Shouldn't we have a default limit even if it's high Eg: 30k records ??

mtuchi: Hiya @josephjclark, I have replaced `max` with `limit` and updated docs, test and add warning if query contain `LIMIT` or `limit` clause. 

Regarding the default value. I think it should be 10k which matches the internal default maxFetch, I think 5k is very small number of records

.changeset/sweet-ties-juggle.md
```diff
@@ -0,0 +1,47 @@
+---
+'@openfn/language-salesforce': major
+---
+
+- Modernize `query()` implementation using jsforce v3
+- Remove `autoFetch` option from `query()` function
+- Add `max` option to `query()` function
```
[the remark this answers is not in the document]
mtuchi: Ooh yes, sorry my bad. I have update the changeset now

josephjclark:
Thanks @mtuchi - I think we'll release this alongside the bulk stuff, in the same major version. I'll hold off the final review for now and try and get your other stuff merged :)

josephjclark:
Tempted to remove common.map form here and release as one breaking change. It'll be a bit cleaner :thinking:

PR #1093, opened 2025-03-22
OpenMRS Pagination 2

josephjclark:
## Summary

This PR extends @PiusKariuki's branch to get pagination working.

I've actually really struggled with this because I've seen a lot of small problems and I'm really short on time today.

So this PR is left in a partial and mostly broken state.

There are some things I like:
- I've refactored unit tests to make them easier to manage
- I've created a mock openmrs server which allows neater unit tests
- And so there's some good unit tests on pagination logic now
- I think the actual pagination logic is quite good here

What I think we need to do from here:
- I've changed the signature of `request` and so everything that calls it is broken :(
- I'm not actually convinced by the return signature of `request` right now. But request need to include the response and the response body seperately, AND it needs to handle the case of multiple responses if paginating. It might be OK as it is, but I need another look with a fresh brain

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


packages/openmrs/src/Utils.js
```diff
@@ -18,20 +17,45 @@ export const prepareNextState = (state, response, callback = s => s) => {
...
+    response.url = urlWithQuery;
     return response;
   } catch (err) {
     return err;
   }
 }
 
+// TODO get() should pass autopage true and options
+// TODO http.*() functions should all pass autopage false (at least for now)
+// TODO this currently returns { result, response }. Is that right? It'll break everything!!
```
josephjclark, line 41: I think this is correct, yes, and so now we have to go and fix everything.

- we use `result` because response bodies may be aggregated, if using pagination.
- `response` right now might be an object or array, if paginate is enabled

Would it be cleaner to have `request()` and `paginatedGet()` as two distinct APIs, with two distinct return types?

That would mean that the main rest `get()` explictly calls paginated get, and basically every other API just calls get. The http helpers will not autopage for you. This is a nice clear seperation of concerns and makes the behaviour and implementation way simpler.

Ok, I think I like that. But then I am very tired.

packages/openmrs/src/Utils.js
```diff
@@ -18,20 +17,45 @@ export const prepareNextState = (state, response, callback = s => s) => {
...
 export async function request(state, method, path, options = {}) {
+  const { username, password, instanceUrl: baseUrl } = state.configuration;
+
   const {
-    baseUrl = state.configuration.instanceUrl,
+    // baseUrl = instanceUrl, // TODO should not be able to override baseUrl on config
```
josephjclark, line 46: I've left this commented out as a reminder: we need to check the implications of removing this line

packages/openmrs/src/Utils.js
```diff
@@ -59,85 +81,76 @@ export async function request(state, method, path, options = {}) {
...
+  }
 
-  let allResponses;
-  let fetchedCount = 0;
+  const response = await fetchAndLog(method, url, opts);
+  const { body, ...responseWithoutBody } = response;
+  return { response: responseWithoutBody, data: body };
+}
 
+// TODO: work out page size (if startIndex AND limit are set, this is the page size I think)
```
josephjclark, line 109: I think we can ignore this comment

josephjclark:
Writing a in integration test and I've discovered a pretty major problem: `delete()` is not a valid operation name and won't compile properly. I thought we could get around this  but we cannot. 

Maybe an issue for kit? use compiler magic to alias the import to `$delete`? hmm

PR #1098, opened 2025-03-25
New adaptor: Google Drive

aloundoye:
## Summary
A new adaptor for Google Drive

Feat #908

## Details

A Google drive adaptor with `get` `update` and `create` helpers
## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to
know!):

- [ ] Code generation (copilot but not intellisense)
- [ ] Learning or fact checking
- [ ] Strategy / design
- [ ] Optimisation / refactoring
- [ ] Translation / spellchecking / doc gen
- [ ] Other
- [ x] I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Review Checklist

Before merging, the reviewer should check the following items:

- [ x] Does the PR do what it claims to do?
- [ ] If this is a new adaptor, added the adaptor on marketing website ?
- [ ] If this PR includes breaking changes, do we need to update any jobs in
      production? Is it safe to release?
- [ x] Are there any unit tests?
- [ ] Is there a changeset associated with this PR? Should there be? Note that
      dev only changes don't need a changeset.
- [ x] Have you ticked a box under AI Usage?

PR #1119, opened 2025-04-07
New adaptor: Google Drive (#1098)

josephjclark:
Release prep & QA of work from @aloundoye 

See #1098


mtuchi:
**Notes After Review**
Hiya @josephjclark, Here is my feedback after review 

> To test locally you need to obtain access token either by `gcloud auth print-access-token` If you have setup `gcloud` cli or Using postman  you can generate access token, Learn more here on [how to access Google APIs with postman](https://blog.postman.com/how-to-access-google-apis-using-oauth-in-postman/)

>**Thoughts on OAuth Setup**: I think we should have a blog/community post on how to setup OAuth on lightning. This is inspired by [how-to-access-google-apis-using-oauth-in-postman](https://blog.postman.com/how-to-access-google-apis-using-oauth-in-postman/)
 
- Function Naming
After testing `create()`, `get()` and `update` It's clear that each function it's main purpose is to work with files only and not folders. The appropriate name will be `createFile()`, `getFile()` and `updateFile()`.

- File Types
Currently the  mimeType is hardcode to `mimeType: 'application/octet-stream',` While you still can add file extension when specify the file name and google will know the type [eg: hello.txt, hello.pdf], You can't create Google specific files types like. 

    File Type | MimeType
    -- | --
    Google Doc | application/vnd.google-apps.document
    Google Sheet | application/vnd.google-apps.spreadsheet
    Google Slide | application/vnd.google-apps.presentation
    Google Drawing | application/vnd.google-apps.drawing
    Google Form | application/vnd.google-apps.form
    Folder | application/vnd.google-apps.folder
    
- Examples 
Job code examples for `get()` and `update()` are wrong. They don't match the function signature. They should be
`get('drive-file-id')`, `update($.data.id, 'SGVsbG8gTWlrZQ==', 'new-filename');`

- Error Handling
For all 3 functions, we are catching an error then throwing it back. I think we should safely through errors by removing things like headers and throw the rest of error message. For example 👇🏽 
```js
.catch(err => {
  const { config, ...safeError } = err.response;
  const { headers, ...safeConfig } = config;
  throwError(safeError.statusText, { ...safeError, ...safeConfig });
});
```
Basically remove any sensitive information or information that is not very useful in error logs 

- `update()` function signature
I find it odd that the third option is the `new file name`. It's probably me being bias but would prefer the third option to be an object with optional params

mtuchi:
Here is the sample code that i used to test the functions 

```js
// Test create
create('SGVsbG8gV29ybGQ=', 'hello.txt');

// Test get
get($.data.id);
get('invalid-id');

// What if id is file id?
// get('1V8ZkXxdimBM1QJ0eF_hkQmsH_9rAgr_r');
//Yep confirmed, Get only works with files, If its folder id you will get `Forbidden`

// Testing Update
create('SGVsbG8gV29ybGQ=', 'hello.mp4', {
  folderId: '15tLwRj0lmr4mGIslEm5QEAS8YJ1EAXep',
});
update($.data.id, 'SGVsbG8gTWlrZQ==');

//To update file name
//update($.data.id, 'SGVsbG8gTWlrZQ==', 'mike-here');

```

josephjclark:
Hi @mtuchi - thank you for this. At this stage I really just wanted to verify that the functions worked as documented.

Documentation on this branch is very clear that `create` just creates a file. I'm quite happy with that. We can support file creation at a later date by adding an option to `create`, or maybe add a `createFolder()` function :shrug:  The same with `get` really, and `update` doesn't make sense on a folder.

Similarly, we can take `mimeType` as an option later. No problem. I presume the existing mimetype is just something generic that enables any arbitrary file content to be uploaded.

Great catch on the examples - that got missed when we changed the signatures in the original PR. Would you mind updating them?

The errors is interesting. We certainly shouldn't catch and re-throw so maybe the catch wrappers just need removing. Is there unsafe data in the returned error? What does it look like - is the token itself in the headers?

`fileName` is a strange on in `update`, I hadn't noticed that. This would allow us to rename the file - in which case you wouldn't need file contents. Good catch on this one, it's quite untidy. I would just remove that option for now.

josephjclark:
So to be clear let's:
- Fix those examples
- remove the `fileName` parameter

And then we can ship!

mtuchi (approved):

josephjclark:
@aloundoye your adaptor is now live! Thank you! https://docs.openfn.org/adaptors/packages/googledrive-docs

aloundoye:
That’s great—thank you both @josephjclark and @mtuchi for your valuable feedback. I learned a lot from this!

PR #1123, opened 2025-04-08
Openmrs 5 final fixes

josephjclark:
- [x] Fix an issue on the main branch in common
- [x] Remove http semantics from rest functions
- [x] One more tweak to paging
- [x] Tidy docs
- [ ]  Tidy release notes
- [x] Integration tests
- [x] Throw if required config isn't set

Pagination has changed again, it now works like this:

The only paginating API is `get(path, options)`

`get()` documentation says that options are query parameters, refer to openMRS docs

So `limit` works just like in regular openMRS. Pass a limit N and you'll get up to N items. No pagination.

By default we'll auto-page results up to `max` (defaulting to 10k), using pages of `pageSize`.  Pass `limit` to disable this. Or set max to `Infinity` to download everything.


josephjclark:
Happy with this now - I'll revise release notes and run QA in the morning (The demo openmrs server is down right now)

PR #1124, opened 2025-04-08
Improve auto-pagination in kobotoolbox

### LINKED ISSUE #988: kobotoolbox: improve pagination
kobotoolbox uses a fairly simple pagination right now. Basically `getSubmissions()` will always download everything, and the `http` helpers don't paginate at all.

What we want to support is:
- Add `limit`, `pageSize` and `start` options to any APIs that support pagination.
- Paginating APIs should download data until the `limit` is reached. If no limit, all data will be downloaded (I think this is fine for default behaviour BUT we should warn about it)
- I think only `getSubmissions()` and `http.get()` actually need to support this level of pagination.

mtuchi:
## Summary

This PR improve auto-pagination for `getSubmission()` by using it's own `util.requestWithPagination()` helper. This separates the paging behavior for `http.{get,post}`  helpers which uses `util.request()` function that does not support auto-paging.

This PR also adds a request operation

Fixes #988 #1191

## Details

- Update chai package to the latest version. 5.2.0
- Rename `Utils.js` to `util.js`
- Remove http export from common
- Add `maxRedirections` option in `HTTPRequesOptions`
- Add `util.requestWithPagination()` function
- Update `getSubmissions()` to use `util.requestWithPagination()` function
- Update `getSubmissions()` options for pagination 
- Update `getSumissions()` docs with examples
- Add few integration tests
- Update unit tests
- Log query params 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/kobotoolbox/src/Adaptor.js
```diff
@@ -52,52 +44,76 @@ export function execute(...operations) {
...
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', {limit: Infinity});
  * @example <caption>Get form submissions with a query</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2022-06-12T21:54:20" } } });
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2025-03-12T21:54:20" } } });
  * @function
  * @public
  * @param {string} formId - Form Id to get the specific submissions
  * @param {object} [options={}] - Optional query params for the request
+ * @param {number} [options.limit] - Limit the number of submissions to fetch
```
josephjclark, line 65: What do you think about adding something like: `restricts query to a single request. Use max to allow multiple requests`

mtuchi: I don't think this is true, Because it depends on `pageSize`

josephjclark: If limit is set, pageSize and max are ignored
```
if (keysToRemove.length) {
        console.warn(
          `Warning: ignoring option [${keysToRemove.join(
            ','
          )}] as "limit" is set`
        );
        delete resolvedOptions.max;
        delete resolvedOptions.pageSize;
      }
```

josephjclark:
@mtuchi I'm having a couple of problems here, but I think I need to put this down until tomorrow.

First off, I'm getting a 301 back from my test server, which is triggering an error. Not sure what that's about.

Secondly, I've confused myself over the handling of query. It  doesn't help that kobo's docs on this are awful. I'm not convinced we got this right in the last release. Trying to verify that.

Thirdly, I don't think we  need `limit` AND `max` here  - but that depends a bit on the previous point.

The reason we have both in openMRS is that the properties `options` all get appended to the url - apart from the stuff documented as `openfn only`. So if the user sets `limit`, it'll set the `limit` query parameter, which has meaning to OpenMRS. And I didn't want to conflict that meaning, so I introduced max.

Here, I don't think the `options` get appended to the URL as query parameters. `options.query` expects a query string or object (which gets encoded into the query param).

Which means that `options.limit` is free, and doesn't have to map to a kobo parameter. And so we don't need limit AND max.

I need to re-read this in the morning and verify all this. It gets confusing :sweat_smile:

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,122 @@
...
+    },
+    query: {
+      format: 'json',
+      ...query,
+    },
+    parseAs,
+    baseUrl,
+  };
+
+  return commonRequest(method, requestPath, options).then(logResponse);
```
josephjclark, line 37: This needs to pass the query object through, like:
```
return commonRequest(method, requestPath, options).then(res =>
    logResponse(res, query)
  );
```

That we we get the query logged in the console.

We  fixed this in common recently but I fear we may have gotten it wrong, see #1163

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,122 @@
...
+        pageSize || maxResults,
+        maxResults - results.length
+      );
+    } else if (!isNaN(pageSize)) {
+      requestOptions.query.limit = pageSize;
+    }
+
+    const response = await request(state, 'GET', path, requestOptions);
+
+    if (response.body.results) {
```
josephjclark, line 76: This needs to be `response.body?.results`, in case there's no response body (I'm seeing this right now with a 301 response)

Or maybe if there's no body we exit the loop?

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    if (isFirstRequest)
+      if (!pageSize) {
+        pageSize = results.length;
+      }
+
+    isFirstRequest = false;
+
+    const hasMoreContent = response.body.next?.includes('start=')
```
josephjclark, line 97: This also needs to be `response.body?.next?`

.changeset/tame-eggs-notice.md
```diff
@@ -0,0 +1,27 @@
+---
```
josephjclark, line 1: I've made some tweaks (but this may change again)
````
---
'@openfn/language-kobotoolbox': major
---

Improved pagination support in `getSubmissions()`

- Removed `paginate` option
- Added new pagination options:
  - `max`: Maximum number of results to fetch (paginated fetch)
  - `pageSize`: Number of items per page in a paginated request
  - `limit`: Force a single request of a fixed size (disables pagination)

### Examples

```js
// Fetch submissions with default limit (10000)
getSubmissions('form-id');

// Fetch all submissions, with pagination
getSubmissions('form-id', { max: Infinity });

// Fetch no more than 500 submissions, with pagination
getSubmissions('form-id', { max: 500 });

// Fetch 100 submissions in a single request
getSubmissions('form-id', { limit: 100 });
```

````

File: packages/kobotoolbox/src/Adaptor.js [hunk 6]
```diff
@@ -52,52 +44,77 @@ export function execute(...operations) {
...
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { limit: Infinity });
  * @example <caption>Get form submissions with a query</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2022-06-12T21:54:20" } } });
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2025-03-12T21:54:20" } } });
  * @function
  * @public
  * @param {string} formId - Form Id to get the specific submissions
  * @param {object} [options={}] - Optional query params for the request
+ * @param {number} [options.limit] - Restricts query to a single request. Use max to allow multiple requests
```
josephjclark, line 65: Thank you for these docs!

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -52,52 +44,77 @@ export function execute(...operations) {
...
  * @example <caption>Get all submissions for a specific form</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { limit: Infinity });
  * @example <caption>Get form submissions with a query</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2022-06-12T21:54:20" } } });
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2025-03-12T21:54:20" } } });
  * @function
  * @public
  * @param {string} formId - Form Id to get the specific submissions
  * @param {object} [options={}] - Optional query params for the request
```
josephjclark, line 64: These are NOT URL query parameters, and they are all openfn only

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -52,52 +44,77 @@ export function execute(...operations) {
...
+ * @param {number} [options.limit] - Restricts query to a single request. Use max to allow multiple requests
+ * @param {object} [options.query] - Query parameters to filter the submissions. See query operators {@link http://docs.mongodb.org/manual/reference/operator/query/.}
```
josephjclark, line 66: This is a query but not `query parameters`

josephjclark:
From call:
1) Yes, let's drop `max`. All options are openfn options
2)  Add `?format=json` to all requests (I think this will fix the 301)

packages/kobotoolbox/src/util.js
mtuchi: Fixed

.changeset/tame-eggs-notice.md
mtuchi: Yeah this has changed to include `http` changes

packages/kobotoolbox/src/util.js
mtuchi: Fixed ✅

.changeset/tame-eggs-notice.md
```diff
@@ -0,0 +1,56 @@
...
+   // Fetch all submissions (no limit)
+   getSubmission('form-id', { limit: Infinity });
+
+   // Fetch submissions with max results
+   getSubmission('form-id', { limit: 500 }); // Returns up to 500 submissions
+   ```
+
+2. HTTP Changes:
+
+- For KoboToolbox API requests: Continue using the built-in methods
```
josephjclark, line 39: Was anyone using general purpose requests? I'm a bit suspicious of this.

I consider `http.get('https://any-api.com/endpoint');` to be invalid, undocumented behaviour. We don't tell anyone to do this (right??), and don't document it. So we can just remove this behaviour without making a big deal about it. The http namespace is also very new so not many users should be affected.

By documenting the breaking change like this, we sort of imply that the old behaviour was supported.

I wouldn't even report it as a bug fix or feature change or anything. I'd just say (when we're ready) `http methods will now throw an error when passing absolute URLs`. No other code examples, workarounds or docs are needed.

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -51,53 +43,69 @@ export function execute(...operations) {
...
-    console.log('✓', response.body.results.length, 'forms fetched.');
     return util.prepareNextState(state, response);
   };
 }
 
 /**
  * Get submissions for a specific form. Calls `/api/v2/assets/<formId>/data/`.
  * @example <caption>Get all submissions for a specific form</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { limit: Infinity });
```
josephjclark, line 59: Ok, but can we add a an example like:
```
 * @example <caption>Get submissions for a specific form</caption>
 * getSubmissions('aXecHjmbATuF6iGFmvBLBX');
 *
```

95% of our users do not need to worry about the limit we impose. And if they ever hit it they'll see a nice error right?

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -51,53 +43,69 @@ export function execute(...operations) {
...
  * @example <caption>Get form submissions with a query</caption>
- * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2022-06-12T21:54:20" } } });
+ * getSubmissions('aXecHjmbATuF6iGFmvBLBX', { query: { _submission_time:{ $gte: "2025-03-12T21:54:20" } } });
  * @function
  * @public
  * @param {string} formId - Form Id to get the specific submissions
  * @param {object} [options={}] - Optional query params for the request
+ * @param {object} [options.query] - (Openfn only) Query options to filter the submissions. See query operators {@link http://docs.mongodb.org/manual/reference/operator/query/.}
```
josephjclark, line 66: Please remove `openfn only` from these docs. It's super confusing.

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -51,53 +43,69 @@ export function execute(...operations) {
...
+ * @param {number} [options.limit=10000] - (Openfn only) Maximum number of submissions to fetch.
```
josephjclark, line 67: Here would be a good place to say `Pass Infinity to disable the limit and download all submissions"

.changeset/tame-eggs-notice.md
mtuchi: In v2 we don't have any job that use `http.get()` to access a different rest api. 
Regarding removing `http` export from common from v1.0.4 to v2.4.3 `http` method was exported from common.
So a request like this would work 
```js
http.get({ url: 'https://jsonplaceholder.typicode.com/todos' })
```

In v3.0.0 we introduces kobotoolbox `http` methods. So this means we can't make request outside kobotoolbox
```js
http.get('https://any-api.com/endpoint');
```
But the `http` common export was not removed.

In this PR, i have removed the common export of http from common

I have removed the example in changeset and left the one line description for http changes

packages/kobotoolbox/src/Adaptor.js
mtuchi: Example added ✅ ,

Regarding the limit we impose, There is a warning message that will appear if we have more items to download but have reached the maximum results limit

packages/kobotoolbox/test/adaptor.test.js
```diff
@@ -0,0 +1,242 @@
...
+  },
+};
+const configuration = {
+  username: 'test',
+  password: 'strongpassword',
+  apiVersion: 'v2',
+  baseUrl: 'https://test.kobotoolbox.org',
+};
+
+const responseWithPagination = (url, objects, query) => {
```
josephjclark, line 24: Didn't we agree to move `responseWithpagination` into its own file, write a set of unit tests against it, and re-use it in both `http.test.js` and `common.test.js`?

Quite important to get this done before we merge I think

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,126 @@
...
+      format: 'json',
+      ...query,
+    },
+    maxRedirections,
+    parseAs,
+    baseUrl,
+  };
+
+  return commonRequest(method, requestPath, options).then(res => {
+    logResponse(res, options.query);
```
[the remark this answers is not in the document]
mtuchi: My bad 🙌🏽 , i might have missed this during changes. Let me revert back

packages/kobotoolbox/test/adaptor.test.js
mtuchi: You're right, I had overlooked this. Appreciate the reminder — I’ll get on it!

mtuchi: Hiya @josephjclark  👋🏽 , I have created `responseWithPagination` helper for unit test and i have reused the helper in `adaptor.test.js` and `util.test.js`

packages/kobotoolbox/test/helper.js
```diff
@@ -0,0 +1,47 @@
+export const responseWithPagination = (url, objects, options) => {
+  const { pageSize, start, limit } = options;
```
josephjclark, line 2: This shouldn't have pagesize. The helper only ever returns a single page of data, right? So it just needs to take start and limit.

Having a pagesize here just adds confusion and means I can't entirely trust the tests

josephjclark: Or am I misunderstanding what this helper does?

packages/kobotoolbox/test/util.test.js
```diff
@@ -0,0 +1,221 @@
...
+  it('should return 2 items with start 1 and pageSize 2', () => {
+    const start = 1;
+    const pageSize = 2;
+    const { results } = responseWithPagination('www', items, {
+      pageSize,
+      start,
+    });
+    expect(results).to.eql([2, 3]);
+  });
+  it('should return all items', () => {
```
josephjclark, line 58: Why? What is this actually testing?

packages/kobotoolbox/test/util.test.js
```diff
@@ -25,100 +26,64 @@ const defaultObjects = [
...
-      'www',
+    const { next, results } = responseWithPagination('www', items, {
       pageSize,
       start,
-      items
-    );
+    });
     expect(results).to.eql([1, 2, 3]);
   });
   it('should return the correct next link', () => {
```
josephjclark, line 77: hmm, I think you should assert the `next` link in every test. Even if undefined. Then you need to rethink the purpose of this test.

because it's hard to test the url string, you might prefer to return like a `_next` object with all the key/value pairs. That's not used by the adaptor, but it can be used in these tests to ensure we're getting the correct behaviour

josephjclark (commented):
I'd like to see more tests generally. There's a lot of heavy logic in `responseWithPagination` and I don't really think the tests are covering all the cases

Basically, I need to be 100% satisfied that this helper function works. If I'm satsifed, I can quickly review the other tests and be confident that the adaptor is working as designed.

josephjclark:
@mtuchi why is this in blocked? Do you want to pair on it for an hour so that we can get it merged?

packages/kobotoolbox/src/Adaptor.js
```diff
@@ -66,8 +66,8 @@ export function getForms() {
  * @param {string} formId - Form Id to get the specific submissions
  * @param {object} [options={}] - Options to control the request
  * @param {object} [options.query] - Query options to filter the submissions. See query operators {@link http://docs.mongodb.org/manual/reference/operator/query/.}
- * @param {number} [options.limit=10000] - Maximum number of submissions to fetch. Pass Infinity to disable the limit and download all submissions
- * @param {number} [options.pageSize=1000] - Limits the size of each page of submissions. Maximum value is 30000.
+ * @param {number} [options.limit=30000] - Maximum number of submissions to fetch. Pass Infinity to disable the limit and download all submissions
```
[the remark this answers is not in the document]
josephjclark: I think `limit` - the absolute maximum number of items returned by our adaptor, should be 10k. It could be 5k or 20k or 13.5k but even 30k. It's arbitrary. This NOT trying to emulate the kobo limit or anything.

10k feels like a good number and I don't understand why it changed.

Similarly I don't know why `pageSize` changed. We could default page size in the adaptor (so we always send `limit=1000`), or we could leave it at the kobo default, which would be 30k. Which is fine, but we should't have a default page size that's bigger than the default limit - that makes no sense. And since w'ere controlling page sizes, having smaller pages means we're less likely to have requests timeout or fail for any reason.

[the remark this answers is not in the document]
mtuchi: I have updated the docs and add more unit tests

mtuchi:
Hiya @josephjclark , I have pushed some changes in this PR. I will take tomorrow on what needs to be updated on changeset and clean up the PR description

josephjclark (commented):
Hi @mtuchi - this is looking pretty good. I think it's the simplest, clearest implementation yet, which is a really good sign.

I'm sorry but I'm going to push you a bit further on the tests. I think it's super important that the behaviours are correct and the test functions do what they are supposed to do (or what they claim to do).

I know it's painful and taking a lot of time, but it's also a really important learning opportunity.

When writing your tests, try to be really clear in your mind _exactly_ what you want to test, make sure the behaviour/expectation is at least roughly reflected in the test name, and make sure the test code is as simple and focused as it can possibly be.

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,132 @@
...
+    ...otherOptions
+  } = options;
+
+  const isUsingDefaultLimit = limit === undefined;
+  const maxResults = limit ?? DEFAULT_MAX_LIMIT;
+
+  let isFirstRequest = true;
+  let requestOptions = { query: { start, limit: pageSize }, ...otherOptions };
+  let shouldFetchMoreContent = false;
+  const didUserPassLimit = Boolean(limit);
```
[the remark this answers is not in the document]
mtuchi: Thanks for spotting that out

packages/kobotoolbox/test/util.test.js
```diff
@@ -0,0 +1,336 @@
...
+  { uid: '4' },
+  { uid: '5' },
+  { uid: '6' },
+];
+
+// TODO - test URL building in a set of separate tests
+describe('responseWithPagination', () => {
+  const items = [1, 2, 3];
+
+  it('use default limit and default start', () => {
```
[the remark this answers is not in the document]
mtuchi: I have clean up the tests mentioned above, please give it another review @josephjclark

packages/kobotoolbox/src/util.js
```diff
@@ -0,0 +1,132 @@
...
+    } else if (!isNaN(pageSize)) {
+      requestOptions.query.limit = pageSize;
+    }
+
+    const response = await request(state, 'GET', path, requestOptions);
+
+    if (response.body?.results) {
+      results.push(...response.body.results);
+
+      if (!start) {
```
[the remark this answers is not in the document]
mtuchi: Good question 🤔 , Looking at the code now, it's not important. So i have removed that check `if(!start)`

packages/kobotoolbox/test/adaptor.test.js
```diff
@@ -0,0 +1,429 @@
...
+        {
+          ...jsonHeaders,
+        }
+      )
+      .times(totalRequests);
+    const { data } = await execute(getSubmissions('aXecHjmbATuF6iGFmvBLBX'))(
+      state
+    );
+    expect(data).to.eql(sampleResponse);
+    expect(requestCount).to.eql(totalRequests);
```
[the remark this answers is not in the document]
mtuchi: Okay i have clean up the test a bit to remove dynamic assertions

packages/kobotoolbox/test/util.test.js
```diff
@@ -0,0 +1,336 @@
...
+    const { results, next, previous, requestCount } = responseWithPagination(
+      items,
+      {},
+      { url: 'www' }
```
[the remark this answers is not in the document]
josephjclark: It could be defaulted though? we're not really using the value in a meaningful way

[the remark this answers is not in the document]
mtuchi: Yes, actually that's what i was thinking. Let me do that

packages/kobotoolbox/test/adaptor.test.js
```diff
@@ -0,0 +1,334 @@
...
+
+    const { data } = await getSubmissions('aXecHjmbATuF6iGFmvBLBX', {
+      pageSize,
+    })(state);
+
+    expect(data.length).to.eql(10);
+    expect(requestCount).to.eql(10);
+  });
+
+  it('should return all items if limit is greater than the number of items (ie, limit = Infinity)', async () => {
```
josephjclark, line 210: I would simply rename this test to "should get all items if limit is Infinity"

PR #1125, opened 2025-04-09
adaptor: automate release dates to adaptor changelogs

### LINKED ISSUE #721: Automate adding release dates and changelogs to adaptor docs
We have some adaptors that are quite old and have not been updated in a while. An example of these adaptors is the medic mobile and CHT adaptors. We also don't have a way for users to know how recent these adaptors are or when they were last updated. 

We have a changelog page on the adaptor docs page which should list the updates made to adaptors and the dates of these updates. We would love to have the changelog automatically updated whenever a new adaptor version is released with the release date. 

This will help users to know how recent an adaptor is.

hunterachieng:
## Summary

Automate adding release dates and changelogs to adaptor docs

Fixes #721

## Details

Use `snapshot` in `changeset` to automatically add dates to changelogs during adaptor releases.

The resulted output is a datetime looking like this: `20211213000730` and we have created another script to help convert this to `2021-12-13` for readability.

So instead of running 
`pnpm changeset version`

We run:

```
pnpm version:changeset  <adaptor-name>

```

The command runs :`pnpm changeset version --snapshot && node scripts/changeset-date.js` that ensures we get dates and update them accordingly 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -15,5 +15,9 @@
     "@openfn/parse-jsdoc",
     "@openfn/slack-notify",
     "import-tests"
-  ]
+  ],
+  "snapshot": {
+    "prereleaseTemplate": "{datetime}",
+    "useCalculatedVersion": true
+  }
 }
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ branch`.
 Version numbers should be bumped with `changeset` and git tags should be pushed
 to the release branch BEFORE merging.
 
-1. Run `pnpm changeset version` from root to bump versions
+1. Run `pnpm version:changeset` from root to bump versions and add release dates
 1. Run `pnpm install`
 1. Commit the new version numbers
 1. Push the branch
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "metadata": "pnpm -C tools/metadata cli",
     "slack:notify": "cd tools/slack && pnpm notify",
     "generate": "pnpm -C tools/generate cli generate",
-    "prerelease": "node scripts/prerelease.mjs --no-warnings"
+    "prerelease": "node scripts/prerelease.mjs --no-warnings",
+    "version:changeset": "pnpm changeset version --snapshot && node scripts/changeset-date.js"
   },
   "author": "Open Function Group",
   "license": "ISC",
--- a/scripts/changeset-date.js
+++ b/scripts/changeset-date.js
@@ -1,0 +1,46 @@
+import fs from 'fs';
+import path from 'path';
+
+const adaptor = process.argv[2];
+if (!adaptor) {
+  console.error(
+    '❌ Please provide the adaptor name. Usage: node changeset-date.js <adaptor-name>'
+  );
+  process.exit(1);
+}
+
+const pkgPath = path.join('./packages', adaptor);
+
+const formatDate = datetime => {
+  const year = datetime.slice(0, 4);
+  const month = datetime.slice(4, 6);
+  const day = datetime.slice(6, 8);
+  return `${year}-${month}-${day}`;
+};
+
+const fixVersionInText = text =>
+  text.replace(/-(\d{14})/g, (_, dt) => `-${formatDate(dt)}`);
+
+const updateFile = (filePath, label) => {
+  if (fs.existsSync(filePath)) {
+    const content = fs.readFileSync(filePath, 'utf8');
+    const updated = fixVersionInText(content);
+    if (content !== updated) {
+      fs.writeFileSync(filePath, updated);
+      console.log(`✔ Updated ${label}`);
+    }
+  }
+};
+
+const pkgJsonPath = path.join(pkgPath, 'package.json');
+if (fs.existsSync(pkgJsonPath)) {
+  const json = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
+  const newVersion = fixVersionInText(json.version);
+  if (json.version !== newVersion) {
+    json.version = newVersion;
+    fs.writeFileSync(pkgJsonPath, JSON.stringify(json, null, 2));
+    console.log(`✔ Updated version in ${adaptor}/package.json → ${newVersion}`);
+  }
+}
+
+updateFile(path.join(pkgPath, 'CHANGELOG.md'), `${adaptor}/CHANGELOG.md`);


package.json
```diff
@@ -20,7 +20,8 @@
     "metadata": "pnpm -C tools/metadata cli",
     "slack:notify": "cd tools/slack && pnpm notify",
     "generate": "pnpm -C tools/generate cli generate",
-    "prerelease": "node scripts/prerelease.mjs --no-warnings"
+    "prerelease": "node scripts/prerelease.mjs --no-warnings",
+    "version:changeset": "pnpm changeset version --snapshot && node scripts/changeset-date.js"
```
josephjclark, line 24: Does it work if this is just called `version`? Does pnpm pick our script over the default?

.changeset/config.json
```diff
@@ -15,5 +15,9 @@
     "@openfn/parse-jsdoc",
     "@openfn/slack-notify",
     "import-tests"
-  ]
+  ],
+  "snapshot": {
```
josephjclark, line 19: hmm. My memory is tingling and I've just checked the docs. I fear we're abusing snapshots with this.

What snapshots are doing is updating the version number to include the date string. So if you do `pnpm version:snapshot`, I think you'll find that the version in package.json updates too - which we don't want

josephjclark:
@hunterachieng maybe a better approach here is to change the changelog function

You can set the changelog function in config: https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#changelog-false-or-a-path

You basically specify a node module which export functions which are called to generate the changelog text. This is the templating function that generates the actual changelog.

Docs on this are a bit sketchy but luckily we don't really need to do a lot - just add the date to the existing string: https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md

You can probably for the default changelog package and add a fixed date string to it

package.json
hunterachieng: @josephjclark  using `pnpm version` does not work

README.md
```diff
@@ -203,7 +203,7 @@ branch`.
 Version numbers should be bumped with `changeset` and git tags should be pushed
 to the release branch BEFORE merging.
 
-1. Run `pnpm changeset version` from root to bump versions
+1. Run `pnpm version:changeset <adaptor-name>` from root to bump versions and add release dates
```
josephjclark, line 206: hmm, one obvious problem here is that `pnpm changeset version` will run for all adaptors that need their versions updating. If you bump common, that means it'll update version numbers for every adaptor. 

But this common only updates dates in the changelog for a single adaptor

scripts/changeset-date.mjs
```diff
@@ -0,0 +1,64 @@
...
+    '❌ Please provide the adaptor name. Usage: node changeset-date.js <adaptor-name>'
+  );
+  process.exit(1);
+}
+
+const pkgPath = path.join('./packages', adaptor);
+const pkgJsonPath = path.join(pkgPath, 'package.json');
+const changelogPath = path.join(pkgPath, 'CHANGELOG.md');
+
+const formatPrettyDate = (datetime) => {
```
josephjclark, line 16: This is fine, but `datefns` (which we already have in this repo) should do this for you way cleaner. I wonder if this is an AI generated solution? :robot:

josephjclark:
Ok, so I see that changesets don't give us enough control of the changelog file to do this, which is a real shame.

I do have some good news though: there is an ast parser for markdown files: https://github.com/syntax-tree/mdast This is _fantastic_.

I think we can implement the same basic functionality for new and existing versions.

Here's what I think we do:

- We have a function like updateChangelog(adaptorName, versionList)
- We write a script to call this function for all adaptors (later we can make this smarter)
- When running for old releases, we pass for each adaptor a list of all the versions and release dates
- When running for new releases, we only pass the latest version (or maybe a string like "latest" and today's date
- When the function does is:
  -  Parse the changelog.md file to build an AST
  - An AST is just a model of the contents of the file. Rather than using  regex to scrape the file for  various patterns, you use the AST to traverse key structures in the document. Like a section.
  - The function traverses the AST and finds all, uh, Heading 2 setcions (`##`)
  - If there isn't a date in there, we look for a corresponding version in the VersionList
  - If this is the first heading in the document, we can use the "latest" version
  - Either way, if we find a version, we add it to the section
  - We serialize the markdown text back out of the model and write to disk
  - And we're done

This is particularly cool because understanding ASTs is a really, really valuable skill in programming (and I can show you the how the auto-generated FHIR4 adaptor works!)

[MD AST guide](https://witve.com/codes/mastering-mdast-for-advanced-markdown-processing/)

scripts/changeset-date.mjs
```diff
@@ -0,0 +1,64 @@
...
+    console.log(`✔ Updated version in ${adaptor}/package.json → ${newVersion}`);
+  }
+};
+
+const updateChangelog = () => {
+  if (!fs.existsSync(changelogPath)) return;
+
+  let content = fs.readFileSync(changelogPath, 'utf8');
+
+  const versionDateRegex = /^## (\d+\.\d+\.\d+)-(\d{8})(\d{6})/gm;
```
josephjclark, line 50: If you think about what you're doing here, you don't actually need the snapshot stuff

Because you can run this regex against a changelog to get all the version headers out (or all the version headers without  dates), and for the first one in the doc, just insert today's date.

There isn't actually anything in `--snapshot` that you need. This regex is the hardest bit!

tools/changelog/src/update-latest-versions.js
```diff
@@ -0,0 +1,79 @@
...
+import fs from 'fs';
+import path from 'path';
+import { format } from 'date-fns';
+import { writeFile } from 'fs/promises';
+import getVersionHistory from './update-version-histories.js';
+
+const packagesDir = '../../packages';
+const versionOnlyRegex = /^\d+\.\d+\.\d+$/;
+
+function getAdaptorsFromDir() {
```
josephjclark, line 13: This function is used in both script - please move it out into a common utils.js file

tools/changelog/src/update-version-histories.js
```diff
@@ -0,0 +1,42 @@
...
+const packagesDir = '../../packages';
+
+function getAdaptorsFromDir() {
+  return fs
+    .readdirSync(packagesDir, { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => dirent.name);
+}
+
+async function fetchVersions(adaptor) {
```
josephjclark, line 12: Can we add some logging here, like in an adaptor code?

Just to give the user some feedback.

So each time you fetch versions, you should log "fetching common" or whatever, so we can see the traffic

tools/changelog/src/update-latest-versions.js
```diff
@@ -0,0 +1,79 @@
...
+    type: 'root',
+    children: updatedAstChildren,
+  };
+
+  const file = await remark().use(stringify).run(updatedAst);
+
+  const updatedMarkdown = String(await remark().use(stringify).stringify(file));
+
+  await writeFile(changelogPath, updatedMarkdown, 'utf8');
+  console.log('✔ Changelog updated');
```
josephjclark, line 66: Can we include the adaptor name?

tools/changelog/src/update-latest-versions.js
```diff
@@ -0,0 +1,79 @@
...
+  const pkgPath = path.join(packagesDir, adaptorName);
+  const changelogPath = path.join(pkgPath, 'CHANGELOG.md');
+
+  if (!fs.existsSync(changelogPath)) return;
+
+  let markdownContent = fs.readFileSync(changelogPath, 'utf8');
+  const ast = await remark().use(remarkParse).parse(markdownContent);
+
+  const updatedAstChildren = ast.children.map((item, index) => {
+    if (item.type === 'heading' && item.depth === 2) {
```
josephjclark, line 30: If we're not updating the whole version history, after checking the first header, we can exist this loop early. When updating the latest versions only, we literally only need to check the first header.

josephjclark:
Oh, and @hunterachieng can we include a very short readme file in the new changelog tool, just to explain what it does and maybe include the key commands? Just something really really basic

josephjclark (approved):
Fantastic Hunter, thank you :pray:

Issue #1117, opened 2025-04-04
Add `debug()` in common


mtuchi:
**Description**
Add `debug()` operation in common

All arguments are passed through to console.debug

References need to be expanded, so you can do `debug($.data)`

Its an operation, so must take state and return state

mtuchi:
This was added on #1138 and released on #1156

PR #1138, opened 2025-04-15
Add `log` and `debug` in common

### LINKED ISSUE #496: common: add a catch() operation
Catch is probably a reserved word so we'd need to think about that

But the idea is that if you put an operation in a catch, it'll call a callback on error. So you can do:
```
catch(upsert, (state, error) => {
   console.error(error)
})
```
This lets you intercept errors and could be useful for  debugging.

If you still want the job to fail, you need to re-throw the error object.

mtuchi:
## Summary

This PR adds `log` and `debug` function in common adaptor

Fixes #496, #1117

## Details

- Add `log(...args)` function with unit tests + jsdocs examples
- Add `debug(...args)` function with unit tests + jsdocs examples

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -932,3 +932,44 @@ export function assert(expression, errorMessage) {
     return state;
   };
 }
+
+/**
+ * Represents a reference to a value in the state object
+ * @typedef {Object} StateReference
+ * @property {string} key - JSON path to the value in the state object
+ */
+/**
+ * Creates a state logger that prints values to the console while maintaining the state chain
+ * @public
+ * @function
+ * @example <caption>Log entire state when no args provided</caption>
+ * log()
+ * // Logs entire state object
+ * @example <caption>Log specific values</caption>
+ * log('Hello', 'World')
+ * // Logs: Hello World
+ * log($.user.name)
+ * // Logs value of state.user.name
+ * @example <caption>Multiple arguments including state references</caption>
+ * log('User:', $.user.name, 'Age:', $.user.age)
+ * // Logs: User: John Age: 30
+ * @param {...(string|number|boolean|StateReference)} args - Values or state references to log. Can be:
+ *   - Primitive values (string, number, boolean)
+ *   - State references (using $ syntax to reference state properties)
+ *   - Objects
+ *   - Arrays
+ * If no args provided, logs entire state
+ * @returns {Operation}
+ */
+export function log(...args) {
+  return state => {
+    const [resolvedArgs] = newExpandReferences(state, args);
+    if (resolvedArgs.length === 0 || !resolvedArgs) {
+      console.log(state);
+    } else {
+      console.log(...resolvedArgs);
+    }
+
+    return state;
+  };
+}
--- a/packages/common/src/Adaptor.js
+++ b/packages/common/src/Adaptor.js
@@ -934,19 +934,11 @@ export function assert(expression, errorMessage) {
 }
 
 /**
- * Represents a reference to a value in the state object
- * @typedef {Object} StateReference
- * @property {string} key - JSON path to the value in the state object
- */
-/**
  * Creates a general-purpose logger that prints values to the console while
  * maintaining the state chain. Suitable for regular application logging
  * and monitoring runtime behavior.
  * @public
  * @function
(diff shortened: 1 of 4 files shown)


packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,53 @@ export function assert(expression, errorMessage) {
     return state;
   };
 }
+
+/**
+ * Outputs a message to the console while maintaining the state.
```
josephjclark, line 937: `while maintaining the state` seems needlessly confusing? What does maintaining state have to do with anything?

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,53 @@ export function assert(expression, errorMessage) {
...
+ * @public
+ * @function
+ * @example <caption>Log specific values</caption>
+ * log('Hello', 'World')
+ * // Logs: Hello World
+ * log($.user.name)
+ * // Logs value of state.user.name
+ * @example <caption>Log a specific object from state</caption>
```
josephjclark, line 945: This is a strange example. If you want to show how to log state without including credentials (and I don't think you need to do that?), then you should name the example that.

Because if I wanted to "log a specific value", I'd probably do `log($.data)`. I wouldn't use all that convoluted spread stuff.

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,53 @@ export function assert(expression, errorMessage) {
...
+ * log(({ configuration, ...rest }) => rest);
+ * @example <caption>Multiple arguments including state references</caption>
+ * log('User:', $.user.name, 'Age:', $.user.age)
+ * // Logs: User: John Age: 30
+ * @param {...*} args - Values to log: primitives, objects, or state references
```
josephjclark, line 950: No-one knows what a state reference is.

tbh I'd lie a little bit and do this:
```
@param {any} value - A value or message to display in the logs
```

It's true that you can pass multiple things (like in console.log) - but 90% of use cases only require a single argument. And by documenting this use-case, we make life easier for everyone

Don't change the implementation though

packages/common/test/index.test.js
```diff
@@ -1110,3 +1112,102 @@ describe('assert', () => {
     expect(error.message).to.eql(`assertion statement failed with false`);
   });
 });
+
+describe('log', () => {
```
josephjclark, line 1116: Great tests :raised_hands:

packages/common/src/Adaptor.js
mtuchi: Good question, Well the idea is your not mutating state with this function, But your right this is confusing. I will remove it. Because as long a a function takes state, you can still mutate state

packages/common/src/Adaptor.js
mtuchi: 🤔 , Right right
Okay i have removed the example for logging state without credential, We still have an example for log message from state, That should be enough

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,49 @@ export function assert(expression, errorMessage) {
     return state;
   };
 }
+
+/**
+ * Outputs a message to the console.
+ * @public
+ * @function
+ * @example <caption>Log specific values</caption>
```
josephjclark, line 940: Hey, sorry I'm sure I commented on this yesterday but it seems to have vanished?

Can we just have a single simple example for each of log and debug which is like `log($.data)`. We really don't need more than that. The existing example are slightly at odds with the result of the docs.

packages/common/src/Adaptor.js
```diff
@@ -932,3 +932,49 @@ export function assert(expression, errorMessage) {
     return state;
   };
 }
+
+/**
+ * Outputs a message to the console.
```
josephjclark, line 937: Maybe:
```
Outputs a message, like calling `console.log`. Use this at the top level of your job code, but not inside callbacks.
```

And then actually maybe you'd add an example like:
```
// log a line inside a callback
fn((state) => {
   console.log(state.data)
   return state
})
```


All this because I'm worried that users will get confused over when to use `log` and when to use `console.log()`

josephjclark (approved):

PR #1145, opened 2025-04-17
Adding Gmail send functionality to Gmail Adaptor

decarteret:
## Summary

Added support for sending messages. Upgraded zip library to support creating archives as well as reading archives.

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/gmail/src/Utils.js
+++ b/packages/gmail/src/Utils.js
@@ -138,12 +138,16 @@ export async function sendMessageWithAttachments(message) {
   const rawMessage = lines.join('\r\n');
   const encodedMessage = encodeEmail(rawMessage);
 
-  const result = await gmail.users.messages.send({
-    userId: 'me',
-    requestBody: { raw: encodedMessage },
-  });
+  try {
+    const result = await gmail.users.messages.send({
+      userId: 'me',
+      requestBody: { raw: encodedMessage },
+    });
 
-  return result.data;
+    return result.data;
+  } catch (error) {
+    throw new Error('Error fetching messages: ' + error.message);
+  }
 }
 
 function encodeEmail(email) {
@@ -185,8 +189,10 @@ async function createArchiveFromFiles(files) {
   const zip = new JSZip();
 
   for (const file of files) {
-    const content = await Promise.resolve(file.content);
-    zip.file(file.filename, content);
+    zip.file(file.filename, file.content, {
+      compression: 'DEFLATE',
+      compressionOptions: { level: 9 },
+    });
   }
 
   return await zip.generateAsync({ type: 'base64' });
--- a/packages/gmail/src/Utils.js
+++ b/packages/gmail/src/Utils.js
@@ -1,4 +1,4 @@
-import unzipper from 'unzipper';
+import JSZip from 'jszip';
 import { google } from 'googleapis';
 
 let gmail;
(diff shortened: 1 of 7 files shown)


packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
   }
 }
 
+// ---
```
josephjclark, line 102: Can we remove this comment?

packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
...
+// Send message start
+// ---
+
+export async function sendMessageWithAttachments(message) {
+  const BOUNDARY = '----=_Part_0_123456789.123456789';
```
josephjclark, line 107: This const should really go at the top of the file, rather than inside this function

packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
...
+
+  const attachments = await parseAttachments(message.attachments || []);
```
josephjclark, line 109: The attachment is actually optional in `parseAttachments`, so it's a bit confusing to default it here. Let's just make this `parseAttachments(message.attachments)`

packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
...
+  }
+
+  lines.push(`--${BOUNDARY}--`, '');
+
+  const rawMessage = lines.join('\r\n');
+  const encodedMessage = encodeEmail(rawMessage);
+
+  try {
+    const result = await gmail.users.messages.send({
+      userId: 'me',
```
josephjclark, line 143: Just to sanity check: `me` is a hardcoded string to mean "send this as teh user whose credential I'm using"?

packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
...
+    return result.data;
+  } catch (error) {
+    throw new Error('Error fetching messages: ' + error.message);
+  }
+}
+
+function encodeEmail(email) {
+  return Buffer.from(email)
+    .toString('base64')
+    .replace(/\+/g, '-')
```
josephjclark, line 156: We do these replacements after the base64 encoding? Is that right? Just feels something a bit abitrary that the google client should be doing :thinking:

packages/gmail/src/Utils.js
```diff
@@ -99,6 +99,104 @@ export async function getMessageContent(message, desiredContent) {
...
+
+  const content = await zip.generateAsync({ type: 'base64' });
+
+  return {
+    filename: attachment.filename,
+    content,
+  };
+}
+
+// ---
```
josephjclark, line 196: This can go too

josephjclark (commented):
Fantastic @decarteret ! Looks great

I've just got a couple of picky little comments (I have a reputation to uphold after all)

I haven't tested anything myself. Once you've addressed the comments would you like me to release?

packages/gmail/src/Utils.js
decarteret: Nice catch.

packages/gmail/src/Utils.js
decarteret: Yes, this is correct; it's a Gmail API definition.

packages/gmail/src/Utils.js
decarteret: Great catch. This is an artifact from an initial implementation that needed to url-encode the content which is no longer needed in this approach.

decarteret:
Great feedback. Changes have been incorporated. PR is ready to be merged pending your approval.

josephjclark (approved):

PR #1159, opened 2025-04-23
dhis2: Implement `tracker` namespace file

### LINKED ISSUE #1007: dhis2: add `tracker` in list of new tracker paths
The new `dhis2` tracker endpoints include `enrollement, trackedEntities, events, and relationships`

In order to effectively use the `upsert` and `update` functions, we need to add `tracker` in the list of the new paths.

The exact code to be updated is: 
```
export function shouldUseNewTracker(resourceType) {
  return /^(enrollments|relationships|events|trackedEntities)$/.test(
    resourceType
  );
}

```

hunterachieng:
## Summary

Implement a tracker namespaced file for `tracker.import()` and `tracker.export()` functions

Fixes #1007

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


packages/dhis2/src/tracker.js
```diff
@@ -0,0 +1,126 @@
...
+ *       {
+ *         attribute: 'w75KJ2mc4zz',
+ *         value: 'Gigiwe',
+ *       },
+ *     ],
+ *   },
+ *  ],
+ * });
+ */
+function _import(strategy, payload, options = {}, callback = s => s) {
```
mtuchi, line 46: Why callback support ?, i thought we are moving to a promise centric design?

hunterachieng: We have a branch that has removed callbacks. Until that is merged then we can fix it

josephjclark: hmm. To be fair this is a new API - are you sure we can't just remove the callbacks now?

packages/dhis2/src/tracker.js
```diff
@@ -0,0 +1,126 @@
...
+ * @typedef {Object} Dhis2State
+ * @private
+ * @property data - The response body (as JSON)
+ * @property response - The HTTP response from the Dhis2 server (excluding the body)
+ * @property references - An array of all previous data objects used in the Job
+ */
+
+/**
+ * Options object
+ * @typedef {Object} RequestOptions
```
mtuchi, line 15: I think request options should point to DHIS2 docs for each resource type instead of having `{ headers, query}` as request option.
This will make the code very tidy and precise
Eg:
```
tracker.export('trackedEntities' { orgUnits: "abc, fyc"})
```
Instead of
```
tracker.export('trackedEntities', { query: { orgUnits: "abc, fyc"}}
```
The only reason i could think of on when a user would want to parse something in headers is when they need a different output format. For example `text/csv , application/csv+gzip application/json+gzip` which could be mapped internally by providing a `parseAs` option


We should have examples for each resourceType and link to request params docs in that example

packages/dhis2/src/tracker.js
mtuchi, line 46: We should add `@alias` to avoid docs looking like this 
Screenshot 2025-04-24 at 2 35 41 PM

mtuchi:
@hunterachieng are you able to run unit tests locally ?, I am getting 4 failed unit test on my pc

josephjclark:
@mtuchi @hunterachieng The signature for import and export is like this:
```
import(strategy, payload, query, options = {})
```

Wouldn't it be easier to just have a single options object, where the options are appended to the query unless otherwise specified?

```
import(strategy, payload, options = {})
```

Then inside the function we extract the options we need and just append the rest:
```
const { apiVersion, parseAs, ...query } = options
```

We do something very similar in [openMRS get](https://docs.openfn.org/adaptors/packages/openmrs-docs#getoptions)

image


image

mtuchi:
Hiya @josephjclark, I like the suggestion, because i can do something like this 
```js
tracker.import(
  "CREATE_AND_UPDATE",
  {
    trackedEntities: [
      {
        orgUnit: "y77LiPqLMoq",
        trackedEntity: "Kj6vYde4LHh",
        trackedEntityType: "nEenWmSyUEp",
      },
    ],
    enrollments: [
      {
        enrolledAt: "2019-08-19T00:00:00.000",
        enrollment: "MNWZ6hnuhSw",
        occurredAt: "2019-08-19T00:00:00.000",
        orgUnit: "y77LiPqLMoq",
        program: "IpHINAT79UW",
        status: "ACTIVE",
        trackedEntity: "Kj6vYde4LHh",
        trackedEntityType: "nEenWmSyUEp",
      },
    ],
  },
  { reportMode: "FULL", importMode: "VALIDATE", atomicMode: "ALL",  parseAs: 'json' }
);
```

In that example above `reportMode, importMode and atomicMode` both are request params supported by DHIS2 when importing a tracker but `parseAs` that's OpenFn only option for parsing the response

hunterachieng:
@josephjclark  and @mtuchi  I have implemented the fix

josephjclark (commented):
Just a little docs tweak and I think we're good!

@mtuchi did you resolve the local tests thing? Do we have a problem there?

packages/dhis2/src/tracker.js
```diff
@@ -0,0 +1,126 @@
...
+ * @property {string} [parseAs='json'] - The response format to parse (e.g., 'json', 'text', 'stream', or 'base64'. Defaults to `json`
+ * @property {string} [apiVersion=42] - The apiVersion of the request. Defaults to 42.
```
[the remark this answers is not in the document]
hunterachieng: We only do this in `dhis2`. Originally we were allowing users to override the value from the options with this statement
```  
// Note that users can override the apiVersion from configuration with args
  if (options?.apiVersion) apiVersion = options.apiVersion;
```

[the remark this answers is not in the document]
josephjclark: Ok, so long as we're not inventing it here that's fine

packages/dhis2/src/tracker.js
```diff
@@ -0,0 +1,132 @@
...
+ * Import data into DHIS2 using the tracker endpoint.
+ * @alias import
+ * @public
+ * @function
+ * @param {string} strategy - The effect the import should have. Can either be CREATE, UPDATE, CREATE_AND_UPDATE and DELETE.
+ * @param {object} payload - The data to be imported.
+ * @param {TrackerOptions} [options] - An optional object containing parseAs, and apiVersion, and queries for the request
+ * @state {Dhis2State}
+ * @returns {Operation}
+ * @example <caption>Import a trackedEntity resource</caption>
```
josephjclark, line 30:
```
* @example <caption>Import some data and pass the `atomicMode` parameter</caption>
* tracker.import('CREATE', $.trackerData, { atomicMode: 'ALL' })
* ```

File: packages/dhis2/src/tracker.js [hunk 5]
```diff
@@ -0,0 +1,132 @@
...
+ * tracker.import('CREATE', {
+ * trackedEntities: [
```
josephjclark, line 32: Indenting needed here

hunterachieng:
@josephjclark  I am done with the fixes

josephjclark (approved):

PR #1160, opened 2025-04-24
common: remove deprecated expandreferences and update common versions

### LINKED ISSUE #1120: Common adaptor migration
This replaces #306 #307 and #317

The aim is to use the new common expandReferences APIs and remove the old stuff.

The aim is for all adaptors to be on workspace:* and to remove legacy code

# Use new expand references

For any affected adaptors:

- Remove `expandReferences` in `Adaptor.js`
- All adaptor should import `expandReferences` from `common/util`

# Migrate common Versions

Some adaptors are fixed on an old common version, eg, `progres`

Usually this is because of expandReferences or http

We should aim to:
- Update the expandReferences
- If the adaptor used to export common.http, it should now export the NEW http
- Replace the implementation on old http with an implementation on new http (we probably can't test this but changes should be minor)
- if anything gets complicated, make noise!!

For any of these adaptors, we should check if they are being used on v2

# Set adaptors to workspace:*

All adaptors should depend on `workspace:*`. Anything on semver should be able to be replaced trivally

# Bump Common ?

if we remove any functions, we have to bump common to a new major. This might involve more work

hunterachieng:
## Summary

Ensure all adaptors use the new `expandReferences` from `common.util`.
Update all adaptors use `workspace:*` .

Fixes #1120

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/common/src/Adaptor.js
```diff
@@ -431,7 +402,7 @@ export function fields(...fields) {
 export function merge(dataSource, fields) {
   return state => {
     const initialData = source(dataSource)(state);
-    const additionalData = expandReferences(fields)(state);
+    const [additionalData] = newExpandReferences(state, fields);
```
josephjclark, line 405: Here in common, `newExpandReferences` needs to be renamed to `expandReferences`.

packages/progres/test/index.js
```diff
@@ -40,7 +40,7 @@ describe('execute', () => {
   });
 });
 
-describe('post', () => {
+describe.skip('post', () => {
```
josephjclark, line 43: We're fixing this off-branch - but just a note that this shouldn't be skipped here

if this PR is breaking `progres`, it's super important that the CI tests fail. Tests failing is a good thing sometimes! It reminds us that work needs to be done before the PR can be merged safely.

josephjclark: I also don't think any changes on this branch right now should be causing the tests to fail :thinking:

hunterachieng: This test was giving an invalid url response. I will have to update the adaptor to `fetch()` and retest

josephjclark: Ah, that's because you've changed the http call for some reason.

OK, we can fix this on another branch. But  skipping the test isn't the right thing to do here. It should be failing, so let it fail.

.changeset/eleven-meals-bow.md
```diff
@@ -0,0 +1,35 @@
+---
+'@openfn/language-googlehealthcare': major
```
josephjclark, line 2: hmm. I think this is a major for common (because `expandReferences) has changed, but I think it's a patch for everyone else (because they don't export expandReferences)

So you probably need two changesets with slightly different notes?

packages/beyonic/src/Adaptor.js
```diff
@@ -1,9 +1,7 @@
-import {
-  execute as commonExecute,
-  expandReferences,
-} from '@openfn/language-common';
+import { execute as commonExecute } from '@openfn/language-common';
```
josephjclark, line 1: Yeah, just to clarify, looking at the code to beyonic there is no public change at all.

josephjclark:
@hunterachieng this is good to go. I've retargeted at a new epic branch `epic/common-3`, where we'll target the next common major changes before release.

Can you please resolve the conflicts, and once tests are passing, merge this into the epic? Thanks!

Re the conflict: we've removed `esno` on `main`, so I think you basically just need to make sure that `esno` isn't referenced on this branch.

hunterachieng:
@josephjclark  This is done

PR #1166, opened 2025-04-28
rapidpro: migrate from deprecated `common-http` to `fetch`

### LINKED ISSUE #1120: Common adaptor migration

hunterachieng:
## Summary

Migrate `rapidpro` from deprecated `common-http` to `fetch`

Fixes #1120

## Details

The current `rapidpro` is using the deprecated `common-http`.
We resorted to using `fetch` from `undici` as `common.request` returns `403`

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/rapidpro/src/Adaptor.js
```diff
@@ -49,30 +46,32 @@ export function execute(...operations) {
...
     const { host, apiVersion, token } = state.configuration;
 
     const url = `${host}/api/${apiVersion || 'v2'}/contacts.json`;
 
-    const config = {
-      url,
-      data: resolvedParams,
-      headers: { Authorization: `Token ${token}` },
-    };
+    return await fetch(url, {
```
josephjclark, line 56: The structure of this is wrong. You should never `return await` and you should never `await.then()`.

So this needs to be like:
```
const response = await fetch(...)
const result = await response.json()
// etc
return nextState
```

packages/rapidpro/src/Adaptor.js
```diff
@@ -49,30 +46,32 @@ export function execute(...operations) {
...
-        const nextState = {
-          ...composeNextState(state, response.data),
-          response,
-        };
-        if (callback) return callback(nextState);
-        return nextState;
-      });
+      console.log('Contact added with uuid:', result.uuid);
+
+      const nextState = {
```
josephjclark, line 68: It's hardly worth the effort but I'm going to say it just about is...

If you default callback to `s => s`  up in the parameter declaration, you can just do:
```
return callback({
  ...composeNextState(state, result),
  response,
})
```

Which removes quite a lot of junk code

packages/rapidpro/src/Adaptor.js
```diff
@@ -104,41 +103,45 @@ export function upsertContact(params, callback) {
...
-              console.log('Contact updated with uuid:', resp.data.uuid);
-              return resp;
-            });
-        } else {
-          console.log(JSON.stringify(data, null, 2));
 
-          delete err.response.request;
-          delete err.response.config;
+          delete config.data['urns'];
+          return await fetch(newUrl, {
```
josephjclark, line 120: Again I think we need to remove this `return await` pattern

packages/rapidpro/src/Adaptor.js
```diff
@@ -104,41 +103,45 @@ export function upsertContact(params, callback) {
...
-          throw err.response;
+            console.log('Contact updated with uuid:', res.uuid);
+            return res;
+          });
+        } else {
+          console.log('Contact added with uuid:', result.uuid);
+          return result;
         }
       })
+      .catch(async err => {
```
josephjclark, line 138: You don't need to keep this catch - just let the promise throw

packages/rapidpro/src/Adaptor.js
```diff
@@ -104,41 +103,45 @@ export function upsertContact(params, callback) {
       headers: { Authorization: `Token ${token}` },
     };
```
josephjclark, line 104: I can't comment in the right place because it's outside the diff - but this `config` object, which used to be used by axios, can now be removed. It has no value.

packages/rapidpro/src/Adaptor.js
```diff
@@ -177,16 +180,20 @@ export function startFlow(params, callback) {
       },
     };
 
-    return http
-      .post(config)(state)
+    return await fetch(url, {
+      method: 'POST',
+      headers: config.headers,
```
josephjclark, line 185: You can remove the config object here too - it's not doing anything

packages/rapidpro/src/Adaptor.js
```diff
@@ -226,16 +233,20 @@ export function sendBroadcast(params, callback) {
...
```
josephjclark, line 238: Config is also unused here

.changeset/six-crabs-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-rapidpro': minor
```
josephjclark, line 2: This is actually a major change, because we're removing the `axios` export (you used to be able to use axios in job code directly).

We need to mention this in the release notes.

packages/rapidpro/src/Adaptor.js
```diff
@@ -1,12 +1,9 @@
 import {
   execute as commonExecute,
   composeNextState,
-  http, // Important: this is the OLD axios-based http.
 } from '@openfn/language-common';
 import { expandReferences } from '@openfn/language-common/util';
-
-const { axios } = http;
```
josephjclark, line 8: Just to be clear - this is a breaking change. Removing an exported function is always a breaking change.

josephjclark (changes requested):
In the exports of `Adaptor.js`, at the bottom of the file, I can see that `http` is exported, with a comment.

This comment is now outdated. I also think we should NOT export `http` at this time. I'd just remove that whole line. And maybe mention this in the changelog too (the `http` namespace has been removed)

josephjclark (approved):
Great!

josephjclark:
`rapidpro` tests are passing so I'm happy to merge this with the failing test (which will be resolved elsewhere)

Issue #499, opened 2024-04-03
common: re-write `map() ` to work like Array.map


josephjclark:
common currently has a function called `map`, which so far as I can tell is exactly the same as `each` (but is documented wrongly).

I think `map` should work like Javascript's native Array.map.

Given an array, it will iterate over each item and call a function. The returned value is added to a new array. Once iteration is complete, the new array is written to state.

Unlike `each`, it does not take an operation as a callback function. Just a function. That's because the return value is not treated like state.

Like `each`, it should pass a scoped state object to the callback operation, where `data` is the item under iteration and the rest of your state is normal.

At the end, `map` should return the new state object, with mapped items on state.data.

It could take options:
- if `filter` is true, returning `null` will remove the item from the array. Default true.
- if `statePath`  is set, the new array will be written to this path on state

That leaves us with a signature like

```
map(path, callback, options)
```

it's pretty weird too have options after callback, but also weird to force options.

josephjclark:
I guess one question on this one is: should we actually RESERVE the word `map`? Does the implementation team use `map` in the same way as we do? I _think _ so but I just wonder if there's a semantic gap there

mtuchi:
After having a chat with @josephjclark this issue seems to be unimportant and needs more clarity on what's need to be implemented if we think people are going to use this helper function. 

@aleksa-krolls I suggest we put a pause until @josephjclark is back online

aleksa-krolls:
Noted @mtuchi pls move to `Blocked` for this week, and we can revisit this at a later date once we hear from @josephjclark

josephjclark:
This came up today

Still not sure about the semantic gap but I think it's fine. It's map as in "map an array of things to another array of things", not "lookup this value for this key", which is where I'm worried about confusion creeping in.

We should scrap the filter behaviour. It's a bit weird and lets us remove the option.

We should scrap the keyname. You can use `as()` or whatever we go with to alias the result. This is more canonical - other operations don't let you choose the key to save to state. why should this?

So the signature is:
```
map(pathOrArray, callback)
```

Which returns the mapped array to state.data and does not filter.

josephjclark:
These examples could sure use it https://github.com/OpenFn/docs/pull/651/files

josephjclark:
To stop this blocking us, we could remove `map()` from common in the next major release, then we can add a new map function whenever we like without the need for a breaking version. I quite like this idea

mtuchi:
[...]

PR #1172, opened 2025-04-29
common: remove map

josephjclark:
## Summary

This PR removes the existing `map()` operation from common.

This leaves us free to introduce a new, better map function later. See #499

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/beige-penguins-pump.md
+++ b/.changeset/beige-penguins-pump.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-common': major
+---
+
+remove `map()`
--- a/.changeset/beige-penguins-pump.md
+++ b/.changeset/beige-penguins-pump.md
@@ -2,4 +2,4 @@
 '@openfn/language-common': major
 ---
 
-remove `map()`
+Remove `map()`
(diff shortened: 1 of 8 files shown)


josephjclark:
Oh no, `mogli` uses `map()`! Let me check that out  hurrah for unit tests huh)

.changeset/beige-penguins-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-common': major
+---
+
+Remove `map()`
```
mtuchi, line 5: @josephjclark mogil and salesforce adaptor use this function, We should remove the export from those adaptor as well

josephjclark:
So `mogli` and `salesforce` both export `map`

That's a major change to salesforce. I'm only gonna call it a minor in `mogli` because I don't want this to be a mogli 1.0 version

josephjclark:
salesforce will have been updated by the time this comes out, so it;s a non-issue. We just need to resolve  any conflict.

PR #1173, opened 2025-04-30
common: move `field()` to util

### LINKED ISSUE #1118: common: move non-operations into util namespace
Common has a bunch of non-operations that should be moved into utils.

For example: `composeNextState`, and `expandReferenes` are utils, not operations.

Also maybe some of these should be private? Neither of those examples should be used in job code.

We need to go through each and make sure each is in the right place

hunterachieng:
## Summary

Move `field()` function to util in common

Fixes #1118

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


packages/http/src/util.js
```diff
@@ -195,4 +196,15 @@ export {
    * console.log(id); // Output:'3f4e254e-8f6f-4f8b-9651-1c1c262cc83f'
    */
   uuid,
+  /**
+   * Returns a key, value pair in an array.
```
josephjclark, line 200: This is the blocking problem: we cannot manually duplicate these docs across all adaptors.

josephjclark:
Ok @hunterachieng I think this is unblocked

I've done a couple of things:

First I've fixed the docs builder so that the http docs looks like this:

image

Note that `util.field` comes up at the bottom there. It's not terribly elegant but it works, and the link works.

Second I've moved the function into `util/index.js`, and set `@namespace` so that the docs render properly. It's not sustainable to have each util in its own file - the files we have there are all for historical reasons really. I'm not really sure how best to organise this - I like that `util/http.js` is self contained.  So let's just pack it all in to `index.js` for now.

josephjclark:
One more step for `field()` is to move the unit tests out of `adaptor.test.js` and into `util/index.test.js`

packages/http/src/util.js
```diff
@@ -196,3 +196,5 @@ export {
    */
   uuid,
 };
+
+export { field } from '@openfn/language-common/util';
```
josephjclark, line 200: This is better now: we can just export what we want out of common util. Docs will automatically be linked up

packages/http/src/util.js
josephjclark: These docs need removing :)

packages/common/src/util/field.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Returns a key, value pair in an array.
+ * @public
+ * @function
+ * @example
+ * field('destination_field_name__c', 'value')
```
josephjclark, line 6: This example needs updating to `util.field`

packages/mysql/src/Adaptor.js
```diff
@@ -298,9 +298,7 @@ export {
   fn,
   fnIf,
   each,
-  http,
```
josephjclark, line 301: Was removing `http` here deliberate? Also seen in `odoo`, so maybe there are others

josephjclark (changes requested):
Hi @hunterachieng - I've had a careful look through this and spotted a couple of things.

This is going to be such a huge piece of work :facepalm:  I'm going to do some user research before we  commit more time to this. I don't want to do it if it's not valuable (I think it's a big improvement, but if I'm the only who thinks that then what's the point?)

Based on the time it's taken to do `field`, can you estimate how long it will take to move the other 10 (I think?) util functions?

packages/mailgun/src/utils.js
```diff
@@ -0,0 +1 @@
+export { field } from '@openfn/language-common/util';
```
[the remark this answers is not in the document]
hunterachieng: Ah I was wondering about this. I shall fix them

packages/mysql/src/Adaptor.js
josephjclark: Ah, well yes you're right. But that's not mentioned anywhere in the PR title or description so it surprised me. And I thought that stuff had been done on the epic.

We need to make sure that this is recorded in a changeset because it's super important to document. Maybe you can generate a major changeset for this now?

We'll also need a major changset for every change to `field` - but maybe we  don't need to do that one yet. I'm not sure the easiest way to handle it.

packages/mailgun/src/utils.js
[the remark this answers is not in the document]
josephjclark: Don't wonder about it- test it! #WWJD

hunterachieng:
On it.

packages/mysql/src/Adaptor.js
josephjclark: Hang on, hang on, there's some confusion here.

When we removed `http` from `rapidpro`, we did so because it was exporting the _old_ http namespace. It was version locked to common v1, and we didn't want to export the common v1 http namespace anymore.

That was actually wrong to do -  we forgot that we'd updated to the latest common, so we should have left `rapidpro` to export `http`. Probably. But don't worry about that for now, I don't want to spend time on rapidpro.

But for other adaptors, we definitely don't want to change common `http` exports. Not here in this PR. We should not be touching `mysql`'s `http` export here. If we want to do that, it must be on a new branch.

So for now, please revert all changes to `http` exports on this branch. We'll look at the in isolation as a separate exercise.

And in future, please try to avoid making unrelated changes in a PR! Try to stay focused on the task in hand.

hunterachieng: I will improve on that. I have pushed the fixes

PR #1175, opened 2025-05-01
Salesforce 7

josephjclark:
Release PR for salesforce 7

* Update `query()`
* Remove `map()`


josephjclark:
So @mtuchi I'm going to go ahead and release the saleforce query stuff. We an introduce the new bulk namespaces as minor versions later on :)

PR #1176, opened 2025-05-01
Update changelogs

josephjclark:
Caught a couple of issues in the changelogs

The last release is missing dates - that's likely to be my fault for not using the latest version command :innocent: 

I also notice that whenever I run the versions script, bulleted list s change from `- thing` to `* thing`. But then when committing they get changed back somehow?? Weird. Anyway I've updated remark to prefer the `-` style

This should prevent weird diffs coming in next time I run the script

### DIFF
--- a/packages/asana/CHANGELOG.md
+++ b/packages/asana/CHANGELOG.md
@@ -1,13 +1,13 @@
 # @openfn/language-asana
 
-## 4.0.12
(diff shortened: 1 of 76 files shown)


josephjclark:
I've raised https://github.com/OpenFn/adaptors/issues/1177 with some other stuff

PR #1179, opened 2025-05-01
Release: common@3

josephjclark:
Release branch for common major 3.0

- remove map
- tidy and update expandreferences
- remove axios


josephjclark:
hmm. Since the util refactor is looking too big and complicated, maybe we should release this as-is :thinking: 

I'll try and look at the docs thing on https://github.com/OpenFn/adaptors/pull/1173. If there's a solution for the sake of a couple of hour's work, we'll do it. Otherwise we'll close that issue and never do it. It may just be too expensive

packages/common/src/Adaptor.js
```diff
@@ -196,41 +196,6 @@ export function lastReferenceValue(path) {
...
- *   create("SObject",
- *     field("FirstName", sourceValue("$.firstName"))
- *   )
- * )
- * @param {string} path - JSONPath referencing a point in `state.data`.
- * @param {function} operation - The operation needed to be repeated.
- * @param {State} state - Runtime state.
- * @returns {State}
- */
-export const map = curry(function (path, operation, state) {
```
mtuchi, line 218: @josephjclark i think we should remove `import curry from 'lodash/fp/curry.js';` since it's no longer used

josephjclark: Oh great catch, thanks!

josephjclark:
I think we made a mistake with `rapidpro` here, and maybe others. But I don't want to take time now to fix them - we can roll in fixes later if it becomes relevant.

Basically `rapidpro` no longer exports `http`. That's wrong, it should continue to export `http`. But it's a _different_ version of `http` because it's now exporting the common 3 http namespace, not the common 1 namespace.

Let's leave this for now, I just wanted to leave a note for the record

josephjclark:
Going to veto the util namespace on this branch. Let's just focus on getting common 3.0 out of door.

josephjclark:
This is ready to merge - but I don't want to do it before heading offline for 24 hours. I'll release on Thursday.

@mtuchi @hunterachieng if you have any time for a little QA against this branch, that would be much appreciated :+1:

PR #1243, opened 2025-06-18
Add `escape` util function in mssql

### LINKED ISSUE #1194: mssql: addParameter
To improve the security of `findValue()`, we should use [`addParameter`](https://tediousjs.github.io/tedious/parameters.html) rather than injecting arguments straight into an SQL string, cutting off a possible injection attack vector.

We need to  build the basic query as a string, but then add the where clause using `addParameter`, mixing terms and parameters.

We can refactor the incoming arguments if that makes this easier.

For the record, in practice, I think the way the SQL query is constructed in `mssql` is fairly robust and reduces the scope for injection attacks. We do quite a lot of manipulation of  the string.

But there's always room for improvement and this is a better solution

mtuchi:
## Summary

Add `escape` util function in mssql adaptor and updated `findValue` to use `escape` when building condition query

Fixes #1194

## Details

- Add `escape` function in `util.js`. This function was an internal function called `escapeQuote` in `Adaptor.js`, I moved it to `util.js` and exported the function in `index.js`
- Add unit test for `escape` function
- Add integration test and docker command for local mssql db setup
- Update `package.json` to include integration test script
- Update chai package to v5
- Update `findValue()` to use `escape` when building the sql query. Also use `reject(err)` instead of `throw err`
- Simplify `cleanupState` function

## AI Usage

know!):

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/mssql/ast.json
+++ b/packages/mssql/ast.json
@@ -94,7 +94,7 @@
           },
           {
             "title": "param",
-            "description": "The condition to use for the lookup. Values are automatically escaped.",
+            "description": "The condition to use for the lookup. Values are automatically escaped for security.",
             "type": {
               "type": "NameExpression",
               "name": "object"


packages/mssql/src/Adaptor.js
```diff
@@ -265,7 +254,9 @@ export function findValue(filter) {
     let conditionsArray = [];
     for (let key in whereData)
       conditionsArray.push(
-        `${key} ${operatorData ? operatorData[key] : '='} '${whereData[key]}'`
+        `${key} ${operatorData ? operatorData[key] : '='} '${escape(
```
josephjclark, line 257: Because this is a very dense template, and because the escaping is very important, can we break out the escaped value into its own variable?

packages/mssql/test/util.test.js
```diff
@@ -0,0 +1,54 @@
+import { expect } from 'chai';
+import { escape } from '../src/util';
+
+describe('util', () => {
+  describe('escape', () => {
```
josephjclark, line 5: Excellent testing

josephjclark (commented):
Can we add a line of docs to indicate that `findvalue.filter.where` will be escaped? There's some escaping in `sql` too but because that's so badly documented I don't want to go in and make a fix there today

packages/mssql/src/Adaptor.js
mtuchi: Added `escapedValue` variable

mtuchi:
@josephjclark regarding updating `chai` version. Because i looked at the unit test and all tests were skipped so i wanted to start a fresh with the latest version of `chai` and `v5.2.0` was in my head 🤷🏽. [Probably i worked on adaptor with that version recently]

PR #1248, opened 2025-06-19
Remove `query` from `logResponse`, add `response.query` in `request()`

### LINKED ISSUE #1163: common: we got log response with query wrong
We updated the `logResponse` function in common to add `query` as a second parameter.

But this is incompatible with all the adaptors that call `logResponse`. We got it wrong! Instead of this:

```
commonRequest(method, requestPath, options).then(logResponse);
```
You have to do this:
```
commonRequest(method, requestPath, options).then(res =>
    logResponse(res, query)
  );
```

What we _actually_ need to do is add `query` to the `response` object returned by common request.

mtuchi:
## Summary

Remove `query` argument from `logResponse` and add `response.query` in `request()` function.

Fixes #1163

## Details

- Remove `query()` argument in `logResponse`
- Refactor `logResponse` to use `response.query`
- Add `resposne.query` in `request` function
- Add unit test for `response.query` in `request()` func
- Add unit test for `logResponse`
- Update `inform` and `fhir-4`  to use the improved `logResponse` function 

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/empty-nails-push.md
+++ b/.changeset/empty-nails-push.md
@@ -1,24 +1,6 @@
 ---
-'@openfn/language-common': major
+'@openfn/language-common': minor
 ---
 
-### Key Changes
-
-**`logResponse` function:**
-
-- **Simplified API**: Removed the need to pass `query` parameters as a separate
-  argument
-- **Enhanced functionality**: Now automatically extracts `query` parameters from
-  the response object
-
-**`request` function:**
-
-- **Enhanced response structure**: Now includes `query` parameters in the
-  response object
-- **Better tracking**: Query parameters are attached to the response for
-  consistent logging
-
-### Breaking Changes
-
-- **API signature change**: `logResponse(response, query)` →
-  `logResponse(response)`
+- Updated internal `logResponse` API
+- Update request to response to include `query` parameters
--- a/.changeset/empty-nails-push.md
+++ b/.changeset/empty-nails-push.md
@@ -1,0 +1,24 @@
+---
+'@openfn/language-common': major
+---
+
+### Key Changes
+
+**`logResponse` function:**
+
+- **Simplified API**: Removed the need to pass `query` parameters as a separate
+  argument
+- **Enhanced functionality**: Now automatically extracts `query` parameters from
+  the response object
+
+**`request` function:**
+
+- **Enhanced response structure**: Now includes `query` parameters in the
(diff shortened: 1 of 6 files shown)


.changeset/empty-nails-push.md
```diff
@@ -0,0 +1,24 @@
...
+**`logResponse` function:**
+
+- **Simplified API**: Removed the need to pass `query` parameters as a separate
+  argument
+- **Enhanced functionality**: Now automatically extracts `query` parameters from
+  the response object
+
+**`request` function:**
+
+- **Enhanced response structure**: Now includes `query` parameters in the
```
josephjclark, line 16: As a general rule we should not be "selling" the changes to the adaptor. Especially for such a small technical change. These are technical release notes, not an advert. Focus on telling users what they need to know when migrating.

On a major version, or a big important minor, you can be a bit more dramatic and promotional. But still your primary audience is exhausted developers who just want to know how to upgrade to this adaptor version (and whether they should bother)

If you want to sell the update and encourage users to migrate, post to community and shout about it (and maybe you can edit a link back into the release notes later)

.changeset/empty-nails-push.md
```diff
@@ -0,0 +1,24 @@
+---
+'@openfn/language-common': major
+---
+
+### Key Changes
+
+**`logResponse` function:**
```
josephjclark, line 7: We don't need to document this publicly, or count it as a major change, because it doesn't affect the public facing API. It only affects our internal APIs.

I would probably just call this a very minor patch dependency with a line like `Updated internal logResponse API`. 99% of users don't need to know about this.

packages/inform/src/Utils.js
```diff
@@ -48,7 +48,5 @@ export const request = (configuration = {}, method, path, options = {}) => {
 
   const safePath = nodepath.join(path);
 
-  return commonRequest(method, safePath, opts).then(response =>
-    logResponse(response, query)
-  );
+  return commonRequest(method, safePath, opts).then(logResponse);
```
josephjclark, line 51: This is nice :+1:

.changeset/empty-nails-push.md
mtuchi: Makes sense

PR #1253, opened 2025-06-27
dhis2: update docs, and examples

### LINKED ISSUE #1081: dhis2: run QA and final review on the finished epic
A reminder issue

When dhis2 7 is "done", I'd like to do a review and run some QA on the whole adaptor. Final flight checks before launch.

hunterachieng:
## Summary

Implement QA changes for `dhis2` ; docs and examples

Fixes #1081

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/four-suits-eat.md
+++ b/.changeset/four-suits-eat.md
@@ -1,6 +1,0 @@
----
-'@openfn/language-dhis2': patch
----
-
-- Adjust logging in create to not include item details.
-- Fix issues in docs.
--- a/packages/dhis2/CHANGELOG.md
+++ b/packages/dhis2/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-dhis2
 
+## 7.0.2 - 01 July 2025
+
+### Patch Changes
+
+- aa2f46b: - Adjust logging in create to not include item details.
+  - Fix issues in docs.
+
 ## 7.0.1 - 20 June 2025
 
 ### Patch Changes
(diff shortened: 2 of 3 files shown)


josephjclark:
@hunterachieng this doesn't seem to be resolved here

image


If you move the examples below the `@state` declaration, does that change anything?

packages/dhis2/src/Adaptor.js
```diff
@@ -220,7 +223,7 @@ export function create(path, data, params = {}) {
     }
 
     const details = `with response ${JSON.stringify(
-      response.body?.message,
+      response.body?.message || response.body,
```
josephjclark, line 226: What are we expecting to see here? We don't actually want to log the output to stdout.

By `response` my first guess is http status, which we already know.

I think I'd prefer to just remove this log, unless we're certain there's a use for it?

hunterachieng:
@josephjclark  we had the examples after the state before and that is what we were seeing. I then switched the order

hunterachieng:
@josephjclark  As per my investigations, when we have `@state` the docs are built in the format below regardless of the position of the examples.  The last statement on the state table is combined with the example without leaving the extra line to signify different entities. 
- Could the issue be how the state object is being constructed?

This operation writes the following keys to state:

| State Key | Description |
| --- | --- |
| data | The response body (as JSON) |
| references | An array of all previous data objects used in the Job |
**Example:** a tracked entity instance. See [Delete tracker docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#webapi_nti_import)
```js
destroy('trackedEntities', 'LcRd6Nyaq7T');
```

PR #1262, opened 2025-07-02
adaptor: opencrvs adaptor

### LINKED ISSUE #736: Implement adaptor for OpenCRVS
For the first version of this adaptor we should do the MDP (minimum delightful product) implementation. We should be able  access two clean, well-written, simple adaptor functions:

- [ ] create a birth notification with https://documentation.opencrvs.org/technology/interoperability/event-notification-clients
- [ ] check for an existing record with https://documentation.opencrvs.org/technology/interoperability/record-search-clients

See: https://github.com/opencrvs/opencrvs-core/pull/9798

hunterachieng:
## Summary

Implement OpenCRVS adaptor with `createBirthNotification` and `searchEvents` name-spaced function with the `post` and `request` functions as well

Fixes #736 

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


packages/common/src/util/http.js
```diff
@@ -256,15 +256,19 @@ function encodeRequestBody(body) {
 }
 
 async function readResponseBody(response, parseAs) {
+  if (!response.body) {
```
josephjclark, line 259: @hunterachieng did we resolve this properly or is it still a problem?

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+const state = {
+  configuration: {
+    domain: 'gateway.fake.opencrvs.server.com',
+    clientId: 'someclientid',
+    clientSecret: 'someclientsecret',
+  },
+};
+
+before(() => {
+  state.configuration.domain = 'fake.opencrvs.server.com';
```
josephjclark, line 26: This is strange - why are we setting the domain here, when we've defined a state object up at the top of the file? The two values are fighting each other - which one do we want?

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+        resourceType: 'Bundle',
+        type: 'document',
+        meta: {
+          lastUpdated: new Date().toISOString(),
+        },
+        entry: birthRecordData,
+      })
+    )(state);
+
+    expect(finalState.data).to.eql({
```
josephjclark, line 73: No assertion on `state.response`?

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+      },
+    })
+    .reply(200, {
+      access_token: 'fake-token',
+    })
+    .persist();
+});
+
+describe('request', () => {
+  it('makes a post request to create a birth notification', async () => {
```
josephjclark, line 44: hmm. I think you need to remove all the birth notification stuff from this test.

If you're testing the raw http helpers, all you need to do is verify that the arguments you pass into `post()` are properly built into the request, and that the response is returned correctly.

You're not actually creating a birth notification in this request. You're calling a fake endpoint (that you defined) and receiving fake data (that you defined). All you're really testing is your own mock code. You're not testing any of the requirements of birth notifications - the data structures or required fields or query parameters or any of that.

You're not testing the birth notification endpoint - you're testing the request helper generally.

So probably a better way to structure this test, and the post one, is to send a request with a method, query parameters and body to an arbitrary endpoint (like `/test/`), and assert through the interceptor that all those values were included.

You also need to verify that `state.data` and `state.response` are correct in the result.  You don't need to return a birth records bundle, that's actually a bit confusing. Just return some random data and make sure it's processed properly.

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+      .intercept({
+        path: '/notification',
+        method: 'POST',
+        headers: {
+          Authorization: 'Bearer fake-token',
+          'Content-type': 'application/json',
+        },
+      })
+
+      .reply(200, {
```
josephjclark, line 131: I would like the reply handler in this test to assert that a bundle has been uploaded in the correct format.

Because if you change the `createBirthNotification` to not use a bundle, then the real OpenCRVS endpoint will fail, but the test will still pass. It's not strict enough.

So I guess I'm saying that in THIS test, you need to simulate the logic of the server a little bit. Because you are now testing that this endpoint _specifically_ is being called correctly.

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+          searchEvents: {
+            totalItems: 0,
+            results: [],
+            __typename: 'EventSearchResultSet',
+          },
+        },
+      });
+
+    const finalState = await execute(
+      queryEvents({
```
josephjclark, line 173: There is no assertion here that these query parameters are actually used. 

The test isn't strict enough. It's not really testing very much. It doesn't test for example that the `eventLocationId` parameter is actually mapped in the query (you could delete the query and the test would still pass)

Is there any way you can test what is sent to the server? Or how the query is assembled?

packages/opencrvs/test/Adaptor.test.js
```diff
@@ -0,0 +1,196 @@
...
+        declarationJurisdictionId: '',
+        eventLocationId: '704b9706-d729-4834-8656-05b562065deb',
+        fatherFirstNames: 'Dad',
+        motherFirstNames: 'Mom',
+      })
+    )(state);
+
+    expect(finalState.data).to.eql({
+      data: {
+        searchEvents: {
```
josephjclark, line 188: The nice thing about this test is that it's shown me how the function works without having to run the code. Which is great because I can form opinions on it :)

Do we really want to return the raw search object? Why do I have to do `state.data.searchEvents.results` in my job code?

I think you should just write the results array to `state.data`. Much easier for users.

packages/opencrvs/test/testData.js
```diff
@@ -0,0 +1,551 @@
+export const birthRecordData = [
```
josephjclark, line 1: There is a LOT of data in this file and it's hardly used by any of the tests. Do we really need to save so much to the repo? If we do, can we make sure that tests are actually using it?

josephjclark (changes requested):

packages/opencrvs/src/Adaptor.js
```diff
@@ -0,0 +1,300 @@
...
+    const response = await util.request(
+      state.configuration,
+      'POST',
+      '/graphql',
+      {
+        body: {
+          operationName: 'searchEvents',
+          query: searchEventsQuery,
+          variables: { advancedSearchParameters: resolvedVariables },
+          count: resolvedOptions.count || 10,
```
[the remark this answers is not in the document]
hunterachieng: It is deliberate. the default from the API is 10. I can document that

packages/common/src/util/http.js
hunterachieng: @josephjclark I am unable to fix it. I have tried different ways but its failing on the contentLength

packages/opencrvs/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: Given that we have no pagination, and pagination is kinda hard in openfn, are you sure we don't want this to be higher?

packages/opencrvs/src/Adaptor.js
```diff
@@ -0,0 +1,300 @@
...
+            code: '73764-3',
+            display: 'Birth attendant title',
+          },
+        ],
+      },
+      valueString: 'PHYSICIAN',
+    },
+  },
+  {
+    fullUrl: 'urn:uuid:eee21c26-67a2-40af-8cf7-5f4bc969153f',
```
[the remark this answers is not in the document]
josephjclark: Not sure I follow this. But if you're saying we can't generate bundle metadata properly, then we need to update our docs, because `body` is not an array of FHIR resources. We may as well just accept a Bundle directly, if we can't generate the bundle data.

Do we have a working workflow which uses this?

packages/opencrvs/src/Adaptor.js
```diff
@@ -0,0 +1,202 @@
...
+          ],
+        },
+      ],
+    },
+  },
+]);
+
+ * @function
+ * @public
+ * @param {Array} body - An array of Birth Notification FHIR resources
```
josephjclark, line 82: Can we list (here or in the main docs body) the required resources? And maybe link to the postman JSON?

This isn't super great practice for documentation, but I think it's OK in this instance

packages/opencrvs/src/Adaptor.js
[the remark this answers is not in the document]
josephjclark: @hunterachieng This question is still outstanding - a default of 10 feels very low to me! But if you're sure, then it's fine

PR #1265, opened 2025-07-04
http: add tls options  in config

### LINKED ISSUE #455: http: add TLS options to configuration schema
In new http, you can pass a `tls` object on any request. This can contain any of node's  [tls options](https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions) - but basically you use it to pass a cert and/or key.

Most usage is going to be something like this I think:

```
 get('/', {
        tls: {
          ca: myCertString,
       },
})
```

We should add tls/cert properties to the configuration schema and automatically load these properties into the requesdt.

The difficulty is that the list of supported options here is huge. Even if we were to add schema for them all, we'd end up with a massive and confusing form for http.

So whats the best way to support this? JSON input?

I would assume that anyone who needs to pass TLS options will already have a good guide of how to do it in node.js or curl, they just need to know where to paste the values they need.

hunterachieng:
## Summary

Add `tls` options  in `http` config

Fixes #455 

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/http/src/util.js
```diff
@@ -102,11 +104,11 @@ export function request(method, path, params) {
       resolvedParams.maxRedirections ??
       (resolvedParams.followAllRedirects === false ? 0 : 5);
 
-    const tls = resolvedParams.tls ?? resolvedParams.agentOptions;
+    const tls = tlsConfig ?? resolvedParams.agentOptions;
```
josephjclark, line 107: This is a breaking change, because any jobs who used to do `get($.url, { tls: $.tlsOptions })` will now stop working

We have to be super careful about not breaking existing user code

We also need to consider that I should be able to a) specify some properties on my credential (like `ca`), but I should also be able to override these config defaults explicitly by passing a `tls` object into the function

packages/http/src/util.js
```diff
@@ -89,8 +89,10 @@ export function request(method, path, params) {
         'Content-Type': CONTENT_TYPES[contentType] || 'application/json',
       };
     }
+    const fullConfig = state.configuration ?? {};
 
-    const baseUrl = state.configuration?.baseUrl;
+    const { username, password, access_token, baseUrl, ...tlsConfig } =
```
josephjclark, line 94: This is a little weird :thinking:  I agree that having the TLS stuff on the credential makes sense. But this assumes that any config options, apart from the documented ones, are TLS related.

Also the only TLS option we document in the configuration schema is `ca`

So I'd be more comfortable if this explicitly just read `ca` from the config, rather than passing everything through. Maybe we should also allow `state.configuration.tls` to be an object, in which case we pass that through. That will only work as raw credential in lightning but I think I'm OK with that

So I think state.configuration needs to be like:
```
tls: a custom tls object
ca: allow the certificate authority to be set
```

Probably if tls is set on the config object, other keys are ignored (like you can't pass both `tls` AND `ca` - just pick one)

packages/http/test/index.js
```diff
@@ -769,15 +763,13 @@ describe('tls', () => {
       configuration: {
         label: 'my custom SSL cert',
         publicKey: 'something@mamadou.org',
-        privateKey: 'abc123',
+        ca: 'abc123',
```
josephjclark, line 766: This test still passes if I comment out `ca`. Actually I can comment out the whole `configuration` object and it'll still pass.

So what is it testing? What's its value? Why do we have this test if it isn't checking the cert?

josephjclark (changes requested):
Nice idea @hunterachieng but we need to think a little more about the implementation and the tests. Also we'll need a changeset

josephjclark (commented):
I've pushed a commit to tweak the tests a little bit. I'm not sure we've got perfect coverage here but it's close enough.

I'm just a bit confused now by the special handling of `state.configuration.ca` if we're also supporting `state.configuration.tls`. Is this because we need `ca` in lightning and it's too hard to pass it through the `tls` option?

Maybe we need to take a call over this!

josephjclark:
Regarding TLS:

- My preferred option is `configuration.tls`, as a nested object
- I don't think the Lightning UI supports a nested object within a credential. Let's check with the team to confirm
- If, as I suspect, we have no support, our options are:
- a) use `configuration.tls` and define the credential as raw JSON
-b ) support `configuration.ca` AND `configuration.tls`. This is a hack to get us moving today. I don't want to make a habit of this!

josephjclark:
My preferred solution is to remove `ca` and use a raw JSON credential in the job

josephjclark (approved):

PR #1275, opened 2025-07-10
Fix common response

josephjclark:
This fixes the issue in OpenCRVS where auth requests don't include a content length and so the body isn't parsed.

A little nervous about this fix tbh

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/common/src/util/http.js
+++ b/packages/common/src/util/http.js
@@ -261,7 +261,14 @@ async function readResponseBody(response, parseAs) {
     contentLength = parseInt(response.headers['content-length']);
   }
 
-  if (contentLength == 0 || response.statusCode === 204) {
+  // Try to work out if the response is empty
+  // If so, we should just return without trying to parse it
+  // (turns out that this isn't easy to do!)
+  if (
+    contentLength == 0 ||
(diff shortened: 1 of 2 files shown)


packages/common/src/util/http.js
```diff
@@ -256,15 +256,24 @@ function encodeRequestBody(body) {
...
+  if ('content-length' in response.headers) {
+    contentLength = parseInt(response.headers['content-length']);
+  }
+
+  // Try to work out if the response is empty
+  // If so, we should just return without trying to parse it
+  // (turns out that this isn't easy to do!)
+  if (
+    contentLength == 0 ||
+    response.statusCode === 204 ||
```
josephjclark, line 269: Not at all happy about this stuff. I can't find a really good, robust way to determine if there's any body content in the response. Which seems crazy right?

josephjclark: Ah, a good note from undici:

```
Recommended way to ensure there is a body to deserialize is to check if status code is not 204, and `content-type` header starts with `application/json`.
```

https://github.com/nodejs/undici/pull/1516/files

PR #1276, opened 2025-07-10
openHIM: Rework openhim adaptor

### LINKED ISSUE #1256: Rework OpenHIM Adaptor
#Background 
Our current OpenHIM adaptor is outdated and only has one function that has some hardcoded logic that was built for demo purposes. The ask of this issue is to rebuild the OpenHIM adaptor  with authentication handled, basic http helpers and some basic function. An initial investigation of the current status of the adaptor was done in [this issue](https://github.com/OpenFn/adaptor-qa/issues/2). 

## Request
1. Authentication Handling: Handle OpenHIM authentication using basic auth or mutual TLS.
Auth documentation: [OpenHIM Authentication Guide](https://openhim.org/docs/api/introduction/authentication)
2. This adaptor should expose basic http helpers
3. This adaptor should include some basic OpenHim helper functions such as:

#### Transaction API
`getTransactions({ limit?: number, status?: string })`
* Fetch recent transactions
* Docs: [Transactions – GET /transactions](https://openhim.org/docs/api/transactions/read)

`getTransactionById({ id: string })`
* Fetch full transaction detail
* Docs: [Transactions – GET /transactions/:id](https://openhim.org/docs/api/transactions/read)

#### Clients API
`listClients()`
* Retrieve all registered client systems
* Docs: [Clients – GET /clients](https://openhim.org/docs/api/clients/read)

`registerClient({ name: string, roles: string[], ... })`
* Register a new client system (e.g., EMR or mediator)
* Docs: [Clients – POST /clients](https://openhim.org/docs/api/clients/create)

`updateClient({ id: string, updates: object })`
* Update existing client system details
* Docs: [Clients – PUT /clients/:id](https://openhim.org/docs/api/clients/update)

####  Routes API
`listChannels()`
* Retrieve all configured routing channels
* Docs: [Channel – GET /channels](https://openhim.org/docs/api/channels/read)

`createChannel(body)`
* Create a new channel
* Docs: [Channel – POST /channels](https://openhim.org/docs/api/channels/create)

####  Tasks API
`listTasks()`
* Retrieve all tasks
* Docs: [Task – GET /tasks](https://openhim.org/docs/api/tasks/read)

`createTask(body)`
* Create a new task
* Docs: [Task – POST /tasks](https://openhim.org/docs/api/tasks/create)

## Toggle
`Adaptors`

## Credentials
LP - `OpenHIM Demo Instance`

## Resources
1. [OpenHIM Docs](https://openhim.org/docs)
2. [Link to issue that tested adaptor functionality](https://github.com/OpenFn/adaptor-qa/issues/2)
3. [Link to Mtuchi's video on the testing](https://www.loom.com/share/190263b425584db8b40e889faedc019a?sid=d3d5eda1-1f49-4c78-92db-ad0036181d53)

### Logo Links
[Logo link](https://openhim.org/img/openhim-logo-green.svg)

hunterachieng:
## Summary

- Implement `http` namespace with `GET`, `POST`, `PUT`, and `DELETE`
- Implemented `client, tasks, transactions, and channels` functions 

Fixes #1256

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Review Checklist


packages/openhim/src/Adaptor.js
```diff
@@ -28,38 +36,374 @@ export function execute(...operations) {
...
+
+    const { taskId } = resolvedoptions;
+    delete resolvedoptions.taskId;
+
+    const response = await util.request(
+      state.configuration,
+      'GET',
+      taskId ? `/tasks/${taskId}` : '/tasks',
+      {
+        query: {
```
josephjclark, line 353: A small thing but we don't actually need to spread this: just do `query: resolvedOptions`

packages/openhim/src/Adaptor.js
```diff
@@ -28,38 +36,374 @@ export function execute(...operations) {
...
+    );
+
+    return util.prepareNextState(state, response);
+  };
+}
+
+/**
+ * Make a request to OpenHIM to create a new channel
+ * @example <caption>Create a channel</caption>
+ * createChannel({
```
josephjclark, line 246: A bit like with opencrvs, I'd love to see more concise example here, maybe with a full example linked elsewhere in the docs

Also, looking at the properties here, could we be more helpful to users? Does every single property need to be defined? Can we default any? Are any required and some optional?

Don't want to over think this but just flagging it. It could be an improvement for later.

packages/openhim/test/index.js
```diff
@@ -45,33 +69,326 @@ describe('execute', () => {
...
-      configuration: {
-        username: 'hello',
-        password: 'there',
-        apiUrl: 'https://openhim.org/demo',
+      managingOrganization: {
+        reference: 'Organization/1',
       },
-    };
+    })(state);
+    expect(data).to.eql('User Authenticated Successfully');
```
josephjclark, line 116: Is there really any value in writing this back to state.data?

packages/openhim/test/index.js
```diff
@@ -45,33 +69,326 @@ describe('execute', () => {
...
+      .intercept({
+        path: '/transactions',
+        method: 'GET',
+      })
+      .reply(200, [{ _id: '001', clientID: '1234' }], { ...jsonHeaders });
 
-    return execute(event(fixtures.event.requestBody))(state).then(state => {
-      let lastReference = state.references[0];
+    const { data } = await getTransactions()(state);
+    expect(data[0].clientID).to.eql('1234');
```
josephjclark, line 130: I would add ` expect(data.length).to.eql(1);` just to make it clear that data is an array

packages/openhim/test/index.js
```diff
@@ -45,33 +69,326 @@ describe('execute', () => {
...
+        { ...jsonHeaders }
       );
-    });
+
+    const { data } = await getTransactions({ transactionId: '001' })(state);
+    expect(data.clientID).to.eql('1234');
+    expect(data._id).to.eql('001');
+  });
+
+  it('should GET transactions with filters', async () => {
```
josephjclark, line 153: Can we have a test that undocumented query options get appended to the url? Like `getTransactions({ foo: 'bar' })` should generate the correct URL

packages/openhim/test/index.js
```diff
@@ -45,33 +69,326 @@ describe('execute', () => {
...
+        path: '/channels/0123',
+        method: 'GET',
+      })
+      .reply(
+        200,
+        {
+          _id: '0123',
+          name: 'Channel Test',
+        },
+        { ...jsonHeaders }
```
josephjclark, line 202: If you rebase against main, you should be able to remove all these JSON headers objects. Mock interceptors will now default json content type headers when appropriate

josephjclark (commented):
Looking nice and clean! Just a few comments and suggestions

packages/openhim/test/index.js
hunterachieng: The API returns that as a statement in the body

packages/openhim/src/Adaptor.js
hunterachieng: I do not think we can default any since this is used to create channels such as transactions and it depicts how it can be accessed through the app and API, so the properties should be left for the user to decide what they need

josephjclark: I'm not sure that's true. We could default `methods` to all the standard verbs, status to `enabled`, type to `http` and so on. There's loads of stuff that the user doesn't really need to provide. Also the `routes` object defined in this example has many empty fields.

OpenHIM's docs on this aren't terribly clear, but their schema also suggests that only a couple of properties are actually required. The rest, presumably, are defaulted by them.

I don't want to spend  great deal of time on this now, so I think I'll raise an issue to explore this, set sensible defaults on our end (if needed) and simpilfy examples.

For now, can we just make a few basic tidy ups?
- on `methods` we can just include one or two verbs, like `GET` 
- The Route can be simplified to something like this:
 ```
 {
 *       name: 'FHIR Server Testing',
 *       secured: false,
 *       host: 'localhost',
 *       port: '3447',
 *       primary: true
*     }
- `rewriteUrlsconfig` and `rewriteUrls` and `addAutoReweiteRules` can be removed

hunterachieng: @josephjclark  I have reduced the items in the example. If some items are pre-filled by openHIM, then there is no need for us to default them right?

josephjclark:
I've reconstructed this against the release branch as https://github.com/OpenFn/adaptors/pull/1281, will release imminently

PR #1286, opened 2025-07-17
odoo: Implement `search` and `searchReadRecord` functions with pagination

### LINKED ISSUE #1251: New helper function for Odoo to retrieve all records
New helper function for Odoo to retrieve all records
Add search_read method to Odoo adaptor

Our use case is a customer who wants to have a workflow that pushes data from Odoo to a spreadsheet in an automated way. They want to have all sales records updated in the spreadsheet.

## Request

We need a new method in the Odoo adaptor that allows a user to retrieve all records from a data model without having to give a record id. 

See below for the current `read` method functioning (it requires a value in second position - an  `id` or several - to return records)

Image

Yet when using Postman with the method `search_read` we do not need to specify a record `id` to return data from this model

Image

## Resources

[_Odoo API docs _
](https://www.odoo.com/documentation/18.0/developer/reference/external_api.html) has documentation about the different methods including the `search_read`

Image

Image

Image

Image

Actually if we predict that some users will have many records, we may want to use `search()` then `read()` separately for better memory management. By having a `search` method as well so that users can do batch processing by getting the array of `ids` first (with `search` method) then reading them in batches (using the existing `read` method).

We could also consider a wrapper method that handles pagination and the logic for batch processing, to facilitate implementation by users.

So ideally we want these additions: `search` and `search_read` , and potentially `batch_search_read` to handle pagination and maybe a last sync date stored.

hunterachieng:
## Summary

 Implement `search` and `searchReadRecord` functions with pagination.

Fixes #1251 

## Details

- Added a `search` function that returns the ids of the records.
- Added a `searchReadRecord` function that searches for a record, then returns all the details about the record.
- For `searchReadRecord`, I used `search` to get all items to use as the `totalCount` for pagination. 

## AI Usage

know!):

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/spicy-pillows-deny.md
+++ b/.changeset/spicy-pillows-deny.md
@@ -1,5 +1,5 @@
 ---
-'@openfn/language-odoo': major
+'@openfn/language-odoo': minor
 ---
 
 Add `search()` and `searchReadRecord()` functions.
--- a/.changeset/spicy-pillows-deny.md
+++ b/.changeset/spicy-pillows-deny.md
@@ -1,0 +1,7 @@
+---
+'@openfn/language-odoo': major
+---
+
+Add `search()` and `searchReadRecord()` functions.
+- `search()` Only returns the record IDs.
+- `searchReadRecord()` returns the records with the specified criteria or the full record if none is given. 
--- a/.changeset/spicy-pillows-deny.md
+++ b/.changeset/spicy-pillows-deny.md
@@ -1,7 +1,0 @@
----
-'@openfn/language-odoo': minor
----
-
-Add `search()` and `searchReadRecord()` functions.
-- `search()` Only returns the record IDs.
-- `searchReadRecord()` returns the records with the specified criteria or the full record if none is given. 
--- a/packages/odoo/test/Adaptor.test.js
+++ b/packages/odoo/test/Adaptor.test.js
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { MockOdooWithSearch } from './MockOdooWithSearch';
+import { MockOdooWithSearch } from './mock/client.js';
 
 import {
   create,
--- a/packages/odoo/test/Adaptor.test.js
+++ b/packages/odoo/test/Adaptor.test.js
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { MockOdooWithSearch } from './MockOdooWithSearch';
 
 import {
   create,
@@ -6,6 +7,7 @@ import {
   read,
   deleteRecord,
   setMockClient,
+  searchReadRecord,
 } from '../src/Adaptor.js';
 
 const configuration = {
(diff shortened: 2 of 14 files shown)


josephjclark:
Why do we need `search` and `searchRead` ? I think 90% of our use-cases would be handled by `searchRead`.

We don't need to match the `odoo` library here so I'd suggest a single `search` function. By default it returns full matches objects, but you can pass the fields you want to return. An empty fields array, or a flag, would just return the ids if that's really useful (ie, that would be a odoo `search` rather than `search_read`)

If we really do want `search` and `searchRead`, then we need to  make our docs a lot clearer.

These helper functions shouldn't assume users are deep experts in the odoo API. If they are they can use the basic HTTP helpers and will probably get on better with them.

packages/odoo/test/MockOdooWithSearch.test.js
```diff
@@ -0,0 +1,89 @@
...
+    is_company: true,
+    im_status: 'im_partner',
+    city: 'Nairobi',
+  },
+];
+
+describe('MockOdooWithSearch.searchRead', () => {
+  let data, mock;
+
+  beforeEach(() => {
```
josephjclark, line 33: Because the data doesn't change between tests, I think this is a `before`, not a `beforeEach()`

packages/odoo/test/Adaptor.test.js
```diff
@@ -178,3 +180,114 @@ describe('delete', () => {
...
+    rows.push({
+      id: 10002,
+      name: 'VendorOnly',
+      is_company: true,
+      im_status: 'im_vendor',
+    });
+    return rows;
+  };
+
+  it('fetches up to the limit using multiple pages and returns nextOffset=null at the end', async () => {
```
josephjclark, line 210: Good - but I disagree with the language of  `fetches up to the limit` here. It fetches all items, but only because the default limit is lower than the number of items.

packages/odoo/test/Adaptor.test.js
```diff
@@ -178,3 +180,114 @@ describe('delete', () => {
...
+      }
+    )(state);
+
+    expect(data.totalFetched).to.equal(200);
+    expect(data.rows).to.have.length(200);
+    expect(data.nextOffset).to.equal(null);
+    expect(data.rows[199]).to.deep.equal({ name: 'P200' });
+  });
+
+  it('stops early when data ends before the limit', async () => {
```
josephjclark, line 231: This isn't quite right because again, the number of items exceeds the limit.

So you have 120 items in the database, but you've passed a `limit` of 200. You're saying "don't give me more than 200 items", and it doesn't, but only by chance.

If you set `limit` lower than the dataset, then this test works

packages/odoo/test/Adaptor.test.js
```diff
@@ -178,3 +180,114 @@ describe('delete', () => {
...
+      { limit: 20, pageSize: 10, offset: 30 }
+    )(state);
+
+    expect(data.rows).to.have.length(20);
+    expect(data.rows[0]).to.deep.equal({ name: 'P31' });
+    expect(data.rows[19]).to.deep.equal({ name: 'P50' });
+    expect(data.nextOffset).to.equal(null);
+  });
+
+  it('returns empty when limit <= 0', async () => {
```
josephjclark, line 266: This is a great test case, well done :ok_hand:

packages/odoo/test/Adaptor.test.js
```diff
@@ -178,3 +180,114 @@ describe('delete', () => {
...
+      'res.partner',
+      { is_company: true },
+      ['name'],
+      { limit: 0, pageSize: 10 }
+    )(state);
+    expect(data.rows).to.deep.equal([]);
+    expect(data.nextOffset).to.equal(null);
+  });
+
+  it('returns empty when pageSize <= 0', async () => {
```
josephjclark, line 280: This is also excellent

.changeset/spicy-pillows-deny.md
```diff
@@ -0,0 +1,7 @@
+---
+'@openfn/language-odoo': major
```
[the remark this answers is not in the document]
hunterachieng: No we have not. Let me change to `minor`

josephjclark (approved):

josephjclark:
Thank you @hunterachieng - please bump and push versions

PR #1288, opened 2025-07-17
Add `searchTask` in asana

### LINKED ISSUE #1284: Asana Enhancement: Support Create-Only Task Behavior
## Background, context, and business value

The current `asana` adapter provides `upsertTask()`, which creates a task if none exists and **updates** it if a match is found (based on a given `externalId`). This is useful in many workflows where Asana tasks may need to be updated on resubmission or correction.

However, we’ve identified another common use case where tasks should be:
- **Created only if they do not exist**, and
- **Never updated** if they already exist.

This is typically needed in workflows where tasks are edited manually in Asana after creation, and updates via automation would overwrite those changes.

## The specific request, in as few words as possible

Support a create-only behavior:
- If a task with the given `externalId` **does not exist**, create it.
- If it **already exists**, do nothing (no update).
- Always log when skipping creation due to task existence.

#### Option 1: Add a New Function
Create a new helper:
```js
createOrIgnoreTask(projectGid, params)
```
- Same matching logic as `upsertTask()` (based on `externalId`)
- If no match → `createTask()`
- If match exists → log and exit without update
- Keeps responsibilities clear and avoids side effects for existing `upsertTask()`

#### Option 2: Extend `upsertTask()` with an `update: false` Option
Add an optional config flag to `upsertTask()`:
```js
upsertTask(projectGid, params, options = { update: true })
```

- `update: true`(default): current behavior (create or update)
- `update: false:` skips `updateTask()` if a match is found, logs, and exits
- Requires dropping support for callbacks in favor of promises

This avoids duplicating logic but introduces slightly more complexity to one function.

```js
export function upsertTask(projectGid, params, callback) {
  return state => {
    const [resolvedProjectGid, { externalId, data }] = expandReferences(
      state,
      projectGid,
      params
    );

    return util
      .requestWithPagination(state, `projects/${resolvedProjectGid}/tasks`, {
        query: { opt_fields: `${externalId}` },
      })
      .then(next => {
        const matchingTask = next.find(
          task => task[externalId] === data[externalId]
        );
        if (matchingTask) {
          console.log('Matching task found. Performing update.');
          const { projects, workspace, ...remainingData } = data;
          return updateTask(matchingTask.gid, remainingData, callback)(state);
        } else {
          console.log('No matching task found. Performing create.');
          return createTask(data, callback)(state);
        }
      });
  };
}
```
### Recommendation
Open to either approach. Option 1 (new function) offers cleaner separation, while Option 2 (extended config) avoids duplication. Please implement whichever approach aligns best with adapter design principles.

mtuchi:
## Summary

Add `searchTask` in asana adaptor. The function will search task in a workspace.  It requires a `workspaceGid` to be added in `state.configuration` 

Fixes #1284

## Details

- Add `searchTask` function in `Adaptor.js`
- Add `integration` and `unit` test for `searchTask`
- Change import from `dist/index.js` to `/src` in `integration.js`

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist


packages/asana/src/Adaptor.js
```diff
@@ -214,6 +214,54 @@ export function upsertTask(projectGid, params, callback) {
...
+ *   },
+ * });
+ * @function
+ * @param {string} projectGid - Globally unique identifier for the project
+ * @param {object} params - an object with an externalId and some task data.
+ * @param {string} params.externalId - The external id field name
+ * @param {object} params.data - The data to create.
+ * @returns {Operation}
+ */
+export function createTaskOnce(projectGid, params = {}) {
```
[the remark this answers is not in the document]
mtuchi: we need something that act as `upsertTask`. If we write `fnIf` or `create.catch()`. It's going to be a lot of work to refactor the job code

packages/asana/src/Adaptor.js
```diff
@@ -214,6 +214,54 @@ export function upsertTask(projectGid, params, callback) {
...
+  return state => {
+    const [resolvedProjectGid, resolvedParams] = expandReferences(
+      state,
+      projectGid,
+      params
+    );
+    const { externalId, data } = resolvedParams;
+
+    return util
+      .requestWithPagination(state, `projects/${resolvedProjectGid}/tasks`, {
```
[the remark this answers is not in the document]
mtuchi: We need to find the task first, if we use the default pagination then we might not find the task. 
Actually now that you mentioned it, I think we need to use [Search tasks in a workspace](https://developers.asana.com/reference/searchtasksforworkspace). The only problem is we will need to specify the `workspaceId` for this to work

packages/asana/src/Adaptor.js
```diff
@@ -214,6 +214,41 @@ export function upsertTask(projectGid, params, callback) {
...
+ * searchTask("Test Search Task", {
+ *   resource_subtype: "default_task",
+ *   sort_by: "modified_at"
+ * });
+ * @function
+ * @param {string} task - The text or name of the task to search for.
+ * @param {object} [query] - Query params. See {@link https://developers.asana.com/reference/searchtasksforworkspace Docs} for a list of valid parameters.
+ * @returns {Operation} An operation that, when executed, returns the search results in state.data.
+ */
+export function searchTask(task, query = {}) {
```
josephjclark, line 230: Can we add `options` with:

- `workspaceGid` (which overrides state.configuration if it's set)
- `projectId` which filters the project. I know this is probably just a query param but because most APIs use `projectId`, I want to see it represented in the signature

josephjclark: I think projects.any is what we want. But ok, let's leave it out for now. Users can build their own URL

packages/asana/configuration-schema.json
```diff
@@ -17,11 +17,20 @@
...
+        "12345"
+      ]
     }
   },
   "type": "object",
   "additionalProperties": true,
   "required": [
-    "token"
+    "token",
+    "workspaceGid"
```
josephjclark, line 34: Are you sure it's required?

josephjclark: Ok, but it was only ever required for `searchTask` other APIs didn't need it

josephjclark (approved):

PR #1299, opened 2025-07-24
chatgpt: add `deepResearch()` function and remove `reasoning_effort`

### LINKED ISSUE #1298: chatgpt: Add support for deepsearch
### Description

- The current `chatgpt` adaptor only supports `/v1/chat/completions` prompts. Models called with this version do not support deepsearch from OpenAI.
- Create a `deepsearchprompt()` that would allow for deepsearch prompts to chatgpt.
- Remove the hardcoded `reasoning_effort: 'low'` and make it dynamic.
- Update docs and tests.

hunterachieng:
## Summary

Add `deepResearch()` function for GPT deep research models
Remove `reasoning_effort` from `prompt()`. Fails with current models `4o`

Fixes #1298 

## Details

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/chatgpt/src/Adaptor.js
+++ b/packages/chatgpt/src/Adaptor.js
@@ -13,6 +13,15 @@ import OpenAI from 'openai';
  * @property {string} reasoning_effort - Use `low`, `medium`, or `high` to constrain effort on reasoning for some models.
  */
 
+/**
+ * Options provided to Chat Responses Create (https://platform.openai.com/docs/guides/deep-research)
+ * @typedef {Object} DeepResearchOptions
+ * @public
+ * @property {string} model - Which mode to use, i.e., `o3-deep-research`.
+ * @property {string} tools - An array of tools to use for the search. Default [{"type": "web_search_preview"}] See (https://platform.openai.com/docs/guides/deep-research#tools)
+ * @property {number} max_tool_calls - The maximum number of tool calls to make. Default: 1
+ */
+
 let client;
 
 /**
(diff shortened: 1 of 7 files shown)


packages/chatgpt/src/Adaptor.js
```diff
@@ -78,12 +86,60 @@ export function prompt(message, opts) {
...
+    console.log('√ Prompt operation completed');
+    return composeNextState(state, msg);
+  };
+}
+
+/**
+ * Prompt GPT deep research interface
+ * @example
+ * deepResearch(
+ *   `Verify if this healthcare facility is legitimate:
```
josephjclark, line 103: Is this example too close to your current project? We shouldn't be making client project logic public.

packages/chatgpt/src/Adaptor.js
```diff
@@ -78,12 +86,60 @@ export function prompt(message, opts) {
...
+ *     max_tool_calls: 1
+ *   }
+ * )
+ * @public
+ * @function
+ * @param {string} message - The prompt
+ * @param {DeepResearchOptions} options - Model, tools and other parameters (https://platform.openai.com/docs/guides/deep-research)
+ * @returns {operation}
+ */
+export function deepResearch(message, opts) {
```
josephjclark, line 123: As per my comment in the issue, are we sure this needs to be a different function? Isn't it just a case of setting a different model and setting different parameters?

I suppose openAI has its own docs section for deep research so maybe this is OK. I just want to hear your thoughts on the design and why you chose to introduce a whole new operation (with a very very similar signature to the existing one)

hunterachieng: The signatures are quite different. I figured since they take different parameters and call different clients; prompt calls `client.chat.completions.create(payload);` while deep research calls `client.responses.create(payload);`
I figured it is way easier to have them seperate

packages/chatgpt/src/Adaptor.js
```diff
@@ -78,12 +86,60 @@ export function prompt(message, opts) {
...
+      state,
+      message,
+      opts
+    );
+
+    const payload = {
+      model: 'o3-deep-research',
+      input: resolvedMessage,
+      tools: [{ type: 'web_search_preview' }, ...(resolvedOpts?.tools || [])],
+      max_tool_calls: resolvedOpts?.max_tool_calls || 1,
```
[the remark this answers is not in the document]
hunterachieng: @josephjclark  When I make a request without `max_tool_calls`, I get this `"max_tool_calls": 225,`.  Should we just let the user pass what they want?

PR #1370, opened 2025-09-11
Add MotherDuck adapter for OpenFN

aseidasvera:
## Summary
- Adds a new adapter for MotherDuck, the serverless cloud analytics platform powered by DuckDB
- Enables OpenFN users to integrate with MotherDuck for cloud-based analytics workflows
- Provides full CRUD operations with enterprise-grade security features

### Core Functionality
- **Cloud-native analytics**: Direct integration with MotherDuck's serverless platform
- **Full CRUD operations**: `query`, `insert`, `insertMany`, `update`, `remove`
- **Table management**: `createTable`, `dropTable`, `describeTable`
- **Database operations**: `listDatabases`, `listTables`
- **Batch processing**: Automatic chunking of large datasets (>1000 records)

### Security & Reliability
- **SQL injection protection**: Comprehensive validation of identifiers and WHERE clauses
- **Input sanitization**: All user inputs validated and escaped
- **BigInt serialization**: Handles DuckDB's BigInt types for JSON compatibility
- **Error handling**: Detailed error messages without exposing sensitive data
- **Connection management**: Automatic cleanup and proper state management

## Testing
- ✅ 61 comprehensive tests covering all operations
- ✅ Security validation tests for SQL injection prevention
- ✅ Integration tests with real MotherDuck connection
- ✅ Error handling and edge case coverage
- ✅ Linting passes with no errors

## Documentation
- Complete JSDoc documentation for all public functions
- README with practical OpenFN job examples
- Configuration examples for various use cases
- ETL pipeline demonstration

## Technical Details
- Built on DuckDB Node.js Neo client (`@duckdb/node-api`)
- Follows OpenFN adapter best practices
- State transformer pattern with `composeNextState`
- Client stored in closure (not in state) for security

## Breaking Changes
None - this is a new adapter

## Checklist
- [x] Tests pass locally
- [x] Linting passes
- [x] Documentation complete
- [x] Security features implemented
- [x] BigInt serialization handled
- [x] Real MotherDuck connection tested

Gen AI Usage (Claude Code) :

- [x] Code generation
- [x] Unit Test Generation
- [x] DocGen annotation


packages/motherduck/src/Adaptor.js
```diff
@@ -0,0 +1,673 @@
...
+ * Describe a table structure
+ * @public
+ * @example
+ * describeTable('users')
+ * @function
+ * @param {string} tableName - The name of the table to describe
+ * @param {object} [options] - Optional options argument
+ * @returns {Operation}
+ */
+export function describeTable(tableName, options = {}) {
```
josephjclark, line 553: Since you have a generic `query` function, I'm going to ask: what is the value of some of these smaller functions? Are you adding them because you need them for your use-case, or just because they're useful to users?

Our usual approach to adaptors is to provide one low-level function which can do basically anything - in this case, `query()`. And then provide higher-order  functions either where they add value, or where users are asking for them specifically.

In the case of `describe` - how does the result get written back to state? Is it a JSON object? Does it make more sense to just log a string?

I don't want to push this too far, but every function has a cost (testing, maintenance, documentation, usability). Do we think the cost of this function is worthwhile?

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+          SELECT COUNT(*) as total_records FROM test_data
+        `)
+      );
+      
+      const result = await operation(baseState);
+      expect(result.data).to.be.an('array');
+      expect(result.response.command).to.equal('WITH');
+    });
+
+    it('should expand references in query text', async () => {
```
josephjclark, line 121: This isn't how expand references works - a "reference" needs to be a function

So like this:
```
query(() => `SELECT * FROM sometable WHERE age >= 18`)
```
You can also feed state through there if you like (just make sure to set the values you need on state beforehand)
```
query((s) => `SELECT * FROM ${s.table} WHERE age >= ${s.age}`)
```

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+      const minAge = 18;
+      
+      const operation = execute(
+        createTable(tableName, [{ name: 'id', type: 'INTEGER' }, { name: 'age', type: 'INTEGER' }]),
+        insert(tableName, { id: 1, age: 25 }),
+        query(`SELECT * FROM ${tableName} WHERE age >= ${minAge}`)
+      );
+      
+      const result = await operation(baseState);
+      expect(result.data).to.exist;
```
josephjclark, line 132: Another little problem we have in this test is that we don't know whether the expand reference actually worked. If you comment out the `expandReferences` call from `createTable` the test will probably still pass

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+      expect(result.response).to.exist;
+      expect(result.response.command).to.equal('SELECT');
+    });
+
+    it('should include query in response when writeSql option is true', async () => {
```
josephjclark, line 137: Great, but I don't think we've tested the inverse?

Also this would help add better assertions to the test above, making sure the sql is generated correctly

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+      const userData = { name: 'Dynamic User', age: 25 };
+      
+      const dynamicInsert = execute(insert('users', userData));
+      
+      const result = await dynamicInsert(baseState);
+      expect(result.data).to.deep.equal(userData);
+      expect(result.response.command).to.equal('INSERT');
+    });
+
+    it('should validate table name for SQL injection', async () => {
```
josephjclark, line 213: Ideally I would like to see a range of tests against `validateSqlIdentifier()` directly, rather than through the main operations, just to validate the behaviour. Not necessary to do before merging though.

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+        id: i + 1,
+        name: `User ${i + 1}`,
+        age: 20 + (i % 30),
+        email: `user${i + 1}@example.com`
+      }));
+      
+      const insertOp = execute(insertMany('users', largeDataset));
+      const result = await insertOp(baseState);
+      
+      // Should handle large batch efficiently
```
josephjclark, line 249: This is a good test but why is this "efficient"? :thinking:  What is actually under test here?

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+      expect(result.response.command).to.equal('INSERT');
+      expect(result.response.rowCount).to.equal(50);
+    });
+
+    it('should handle empty array gracefully', async () => {
+      const insertOp = execute(insertMany('users', []));
+      const result = await insertOp(baseState);
+      
+      // Should handle empty array gracefully
+      expect(result.data).to.equal('No records to insert');
```
josephjclark, line 260: I would usually expect to see this as `console.log` line, rather than a special state object. `state.data` should probably have `{ rowCount: 0 }` for consistency

packages/motherduck/test/Adaptor.test.js
```diff
@@ -0,0 +1,885 @@
...
+        }
+      };
+      
+      const queryOp = query('SELECT 1');
+      
+      try {
+        await queryOp(incompleteConfig);
+        // Should handle missing token
+      } catch (error) {
+        expect(error).to.be.an('error');
```
josephjclark, line 822: Can we be a bit more specific here?

aseidasvera:
Hi @josephjclark , I've been offline the last couple of weeks, I'll be picking this up this week, thanks for the feedback, expect a PR by early next week!

packages/motherduck/src/Adaptor.js
```diff
@@ -0,0 +1,673 @@
...
+      return queryHandler(state, sqlQuery, options);
+    } catch (error) {
+      console.error('Describe table operation failed:', error);
+      throw error;
+    }
+  };
+}
+
+/**
+ * Connect to a shared MotherDuck database
```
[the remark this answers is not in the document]
aseidasvera: Agreed, I've removed it, accessing a shared DB could be handled in the createConnection code and is an edge case anyway.

aseidasvera:
Hi Joe,

I'm just back from a trip to our India office, I thought I would have more
time while there. I started working on the adaptor using the PR comments as
a running list but have not done a commit of the changes yet. I will try to
get the remaining changes done today/tomorrow and let you know if I need
any help! Thanks !

Best, Aseidas

packages/motherduck/src/Adaptor.js
```diff
@@ -0,0 +1,673 @@
...
+/**
+ * List all tables in the current database
+ * @public
+ * @example
+ * listTables()
+ * @function
+ * @param {object} [options] - Optional options argument
+ * @returns {Operation}
+ */
+export function listTables(options = {}) {
```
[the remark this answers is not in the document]
aseidasvera: Yup, that makes sense, I'm removing these anyway as they don't add immediate value, I will follow that recommendation for a later release.

josephjclark (approved):

aseidasvera:
Thanks for the input and fast turnaround @josephjclark , this adaptor is really important for us!

PR #1418, opened 2025-10-23
throw an error if `maxRedirections` not set in common http

### LINKED ISSUE #1169: common should warn if there is a redirect
**Description**
Currently common does not support automatic follow redirects, which is a good default for security reasons but it does not warn or log what option you should pass to allow follow redirects.

mtuchi:
## Summary

This PR  adds an error message if `maxRedirections` is not set but the response has [redirect statusCodes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#redirection_messages)

## Details
- Add `throwOnUnhandledRedirect: true` option for internal use only in request function. Currently this option is used in `inform` adaptor to bypass `maxRedirection` check because inform handles redirect manually.
- Throw an error if response has redirect status code and `maxRedirections` is not set.
- Add unit test for handling redirect with and with out `maxRedirections` option
- Set `throwOnUnhandledRedirect: false` in inform adaptor to avoid the throwing an error when redirecting 

Fixes #1169

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/old-buttons-live.md
+++ b/.changeset/old-buttons-live.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-inform': minor
----
-
-disable redirect warning
--- a/.changeset/twelve-plants-juggle.md
+++ b/.changeset/twelve-plants-juggle.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-common': minor
----
-
-warn if there is redirect
--- a/packages/arcgis/CHANGELOG.md
+++ b/packages/arcgis/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-arcgis
 
+## 1.0.2 - 12 November 2025
+
+### Patch Changes
+
+- Updated dependencies \[4d7a833]
+  - @openfn/language-common@3.2.0
+
 ## 1.0.1 - 04 November 2025
 
 ### Patch Changes
--- a/packages/arcgis/package.json
+++ b/packages/arcgis/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openfn/language-arcgis",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "OpenFn arcgis adaptor",
   "type": "module",
   "exports": {
--- a/packages/asana/CHANGELOG.md
+++ b/packages/asana/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-asana
 
+## 5.0.6 - 12 November 2025
+
+### Patch Changes
+
+- Updated dependencies \[4d7a833]
+  - @openfn/language-common@3.2.0
+
 ## 5.0.5 - 04 November 2025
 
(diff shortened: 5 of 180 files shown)


mtuchi:
@hunterachieng good question, I didn't know `undici` handle redirect automatically which raise a question why do we need `maximumRedirects` ?

hunterachieng:
@mtuchi  In unidici, we added a maxRedirect if we dot get it. Do you mind checking that in common `http.js`?

mtuchi:
@hunterachieng yes it's in common http util, I think i nee do sync with @josephjclark on this

packages/common/test/util/http.test.js
```diff
@@ -1135,3 +1135,101 @@ describe('generateAgentKey', () => {
...
+          headers: {
+            location: `https://redirect-test-${code}.com/new-path`,
+          },
+        });
+
+      const response = await request(
+        'GET',
+        `https://redirect-test-${code}.com/old-path`,
+        {
+          errors: false, // Don't throw on non-2xx status
```
josephjclark: We only throw on 400+ errors by default

Please check every line of your AI generated code super carefully. It doesn't take long before our test suites are full of useless test  cases and bad practice examples. You really have to thoroughly vet every line and think about whether it's fit for purpose

packages/common/test/util/http.test.js
```diff
@@ -1135,3 +1135,101 @@ describe('generateAgentKey', () => {
...
+      expect(warningMessage).to.include(
+        `https://redirect-test-${code}.com/new-path`
+      );
+      expect(warningMessage).to.include(
+        `https://redirect-test-${code}.com/old-path`
+      );
+    });
+  });
+
+  it('should not log warning for 301 when maxRedirections is explicitly set to 0', async () => {
```
josephjclark: This is actually a good test :+1:

packages/common/test/util/http.test.js
```diff
@@ -1135,3 +1135,101 @@ describe('generateAgentKey', () => {
...
+    { code: 301, name: 'Moved Permanently' },
+    { code: 302, name: 'Moved Temporarily' },
+    { code: 303, name: 'See Other' },
+    { code: 307, name: 'Temporary Redirect' },
+    { code: 308, name: 'Permanent Redirect' },
+  ];
+
+  redirectTestCases.forEach(({ code, name }) => {
+    it(`should log a friendly warning for ${code} (${name}) redirects when maxRedirections is not set`, async () => {
+      const redirectClient = enableMockClient(
```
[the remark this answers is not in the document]
mtuchi: We can't use the same client because it doesn't have `maxRedirections` option setup. I have refactor my implementation to show when exactly do we need redirectClient

mtuchi:
@josephjclark Done ✅ , Sorry about force push, i forgotten to rebase before bumping versions

Issue #731, opened 2024-08-21
Add `mGet()` and `mSet()` function to `redis` adaptor


mtuchi:
### Background
The  `redis` adaptor supports basic Redis commands, it lacks support for the `mGet()` and `mSet()` commands, which are essential for efficiently handling multiple key-value pairs in Redis. 

### Feature Request
I propose adding two new helper methods to the Redis adaptor: `mGet()` and `mSet()`. These methods should mirror the functionality of Redis's native `json.mGet` and `json.mSet` commands, allowing us to retrieve and set multiple key-value pairs in a single operation.

### Implementation Details
1. **`mGet(keys: Array<string>)`:**
   - Accepts an array of strings, where each string is a key in Redis.
   - Returns an array of values corresponding to the provided keys.

2. **`mSet(keyValuePairs: Object)`:**
   - Accepts an object where each key-value pair corresponds to a Redis key and its associated value.
   - Sets all the key-value pairs in Redis in a single operation.

### Use Cases
- **`mGet()`:** This will be particularly useful when retrieving multiple related values from Redis in a single call, reducing the need for multiple `jGet` operations and improving performance.
- **`mSet()`:** This will allow us to efficiently store multiple key-value pairs at once, which is beneficial when working with bulk data.

PR #1486, opened 2026-01-14
This PR addresses critical Windows compatibility issues across the repo and adds new batch operations to the Redis adaptor. on issue #731

triskyi:
### Fixes
- **Docs Generation**: Fixed `ERR_UNSUPPORTED_ESM_URL_SCHEME` by using `pathToFileURL` in build tools.
- **Test Infrastructure**: Resolved `MockNotMatchedError` in `asana` and `collections` by replacing `path.join` (backslashes on Windows) with `path.posix.join` for URL construction.
- **Build Tools**: Patched `@openfn/buildtools` to support Windows paths in `tsup`.

### New Features (Redis)
Implemented batch operations using native Redis `MSET` and `MGET` commands for improved performance.

- **`mSetString(values)`**: Sets multiple string keys in one request.
- **`mGetString(keys)`**: Retrieves multiple string keys in one request.

## 2. Verification
- [x] **Redis**: Added `packages/redis/test/mGetSetString.test.js` (Passed).
- [x] **Redis**: Verified with valid live test script against Cloud Redis.
- [x] **Collections**: All tests passed (fixed normalization bug).
- [x] **Asana**: All tests passed.
- [x] **Build**: Verified `docs` and `src` build targets on Windows.

## 3. Changesets
- Added 4 changeset files covering the fixes and the new feature.

### DIFF
--- a/packages/common/ast.json
+++ b/packages/common/ast.json
@@ -6,7 +6,7 @@
         "operations"
       ],
       "docs": {
-        "description": "Execute a sequence of operations.\nMain outer API for executing expressions.",
+        "description": "Execute a sequence of operations.\r\nMain outer API for executing expressions.",
         "tags": [
           {
             "title": "public",
@@ -15,7 +15,7 @@
           },
           {
             "title": "example",
-            "description": "execute(\n   create('foo'),\n   delete('bar')\n )"
+            "description": "execute(\r\n   create('foo'),\r\n   delete('bar')\r\n )"
           },
           {
             "title": "private",
@@ -63,7 +63,7 @@
           },
           {
             "title": "example",
-            "description": "fn(state => {\n  // do some things to state\n  return state;\n});"
+            "description": "fn(state => {\r\n  // do some things to state\r\n  return state;\r\n});"
           },
           {
             "title": "param",
@@ -146,7 +146,7 @@
         "path"
       ],
       "docs": {
-        "description": "Picks out a single value from a JSON object.\nIf a JSONPath returns more than one value for the reference, the first\nitem will be returned.",
+        "description": "Picks out a single value from a JSON object.\r\nIf a JSONPath returns more than one value for the reference, the first\r\nitem will be returned.",
         "tags": [
           {
             "title": "public",
@@ -198,7 +198,7 @@
         "path"
       ],
       "docs": {
-        "description": "Picks out a single value from source data.\nIf a JSONPath returns more than one value for the reference, the first\nitem will be returned.",
+        "description": "Picks out a single value from source data.\r\nIf a JSONPath returns more than one value for the reference, the first\r\nitem will be returned.",
         "tags": [
           {
             "title": "public",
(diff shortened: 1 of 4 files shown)


packages/collections/src/collections.js
```diff
@@ -468,7 +468,7 @@ export const request = async (state, client, path, options = {}) => {
   const { headers: _h, query: _q, ...otherOptions } = options;
   const query = parseQuery(options);
   const args = {
-    path: nodepath.join(basePath, path),
+    path: nodepath.posix.join(basePath, path),
```
josephjclark: I didn't know about this - thanks!

triskyi:
Okay problem

packages/redis/src/Adaptor.js
```diff
@@ -201,6 +223,26 @@ export function set(key, value) {
...
+ * @param {Record<string, string>} values - Key-value pairs to set
+ * @state {RedisState}
+ * @returns {Operation}
+ */
+export function mSetString(values) {
+  return async state => {
+    const [resolvedValues] = expandReferences(state, values);
+    console.log(`Setting string values for ${Object.keys(resolvedValues).length} keys`);
+
+    const entries = Object.entries(resolvedValues).flat(); // [key1, value1, key2, value2,...]
```
josephjclark: The comments here (and the next line) aren't particularly helpful for future users :thinking:

packages/redis/test/manual-live-test.js
```diff
@@ -0,0 +1,79 @@
+import { mSetString, mGetString, setMockClient } from '../src/index.js';
```
josephjclark: Love to see an integration test here! These are always helpful for testing changes.

However, this test sort of misses the point. The test code - which is AI generated right? - doesn't know about OpenFn and sort of simulates a runtime environment. It makes for strange code, which doesn't reflect the expected usage of the adaptor, and doesn't properly exercise the code anyway.

If you want to include an integration test for us to use later, I would much prefer that it was implemented as OpenFn job code to be executed by the CLI (users would have to pass in a state object with credntials)

Alteranatively, a good trick is to call the adaptor's `execute` function and pass the operations you want to run. This at least gives you a much better simulation of how the adaptor code would be used in a real workflow. You can see some examples of this in `http.test.ts` - the stuff inside the execute arguments looks a lot like job code, so you get a way nicer test :)

tools/adaptor-apis/src/parse.ts
```diff
@@ -1,5 +1,6 @@
 import jsdoc2md from 'jsdoc-to-markdown';
 import path from 'node:path';
+import { pathToFileURL } from 'node:url'; // ADDED
```
josephjclark: This comment needs removing please

josephjclark:
Thanks @triskyi ! It'll be great to get this over the line in the next couple of days.

triskyi:
@josephjclark Thanks for the feedback. I’ve made the requested updates based on your comments:

Removed mSetString / mGetString after confirming they were redundant.

Added the Windows-related notes to the changelog and included them in the relevant changesets.

Removed the manual Redis live test.

Cleaned up unnecessary comments and logging.

Please let me know if there’s anything further you’d like me to adjust. Thanks again for the guidance.

PR #1506, opened 2026-01-30
Update undici to 7.x

josephjclark:
## Summary

This PR bumps `undici` up to v7.0

Note that undici 7.0 does NOT run on node 18. We need to consider whether that's a problem - our servers use 22.0 but we haven't official announced dropping support for node18. This would effectively do that, and I'm not sure where that hurts us

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "lodash": "^4.17.21",
     "mocha": "^10.7.3",
     "tsx": "^4.19.2",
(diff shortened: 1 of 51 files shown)


josephjclark:
Too hairy for a friday afternoon and needs a proper changeset on anything that might break on node 18.

I'll merge this next week

PR #1514, opened 2026-02-03
Fix fhir generators with IG guides

### LINKED ISSUE #955: fhir4: things I'm breaking

josephjclark:
This PR updates the fhir generator to restore support for generating from an IG spec. This had been dropped a bit while building out the `fhir-4` generator (and also the switch to typescript).

This PR introduces minimal changes to fhir-4, and mostly updates the generator and adds a new reference `fhir-eswtatini` adaptor.

Fixes #955

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/green-singers-wave.md
+++ b/.changeset/green-singers-wave.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-fhir-4': patch
----
-
-Update d.ts defintion structure
--- a/.changeset/honest-facts-agree.md
+++ b/.changeset/honest-facts-agree.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-fhir-4': patch
----
-
-Rebuild on latest generator (should be no behaviour differences)"
--- a/packages/fhir-4/CHANGELOG.md
+++ b/packages/fhir-4/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/language-fhir-4
 
+## 0.2.7
+
+### Patch Changes
+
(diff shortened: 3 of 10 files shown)


josephjclark:
Basic adaptor is now working - tested in local lightning and CLI

Big caveat: it looks like all code  assist support has been dropped from Lightning, so it's not easy to test the type definitions. Bit of a bummer.

I think I'm gonna release this to lock the work in - but I'll be back soon so make improvements.

Issue #1319, opened 2025-08-05
Only when we need more powerful PDF adaptor :)) Browserless adaptor (for pdf conversion)


mtuchi:
**Description**
#1260 issue add [pdfshift](https://docs.pdfshift.io/#credits-usage) adaptor which is very useful for converting html to pdf. However their free tier limit are [**50 request/month**](url) which might not be enough for workflow that need to convert  more than 50 documents in a month.  

A better alternative for this adaptor would be [browserless](https://www.browserless.io/pdf-screenshot-screencast). Their open source, have more free credits (1k/month) and they have both [REST API ](https://docs.browserless.io/open-api#tag/Browser-REST-APIs/paths/~1chrome~1pdf/post) and a [nodejs sdk](https://github.com/browserless/browserless/blob/main/bin/scaffold/README.md#routing) (A bit advanced)

mtuchi:
@laszcz have you tried [browserless](https://www.browserless.io/) before ?

PR #1517, opened 2026-02-04
feat(browserless): add Browserless adaptor for PDF conversion

triskyi:
## Details
Hello @mtuchi, @josephjclark, @martalovescoffee 
This PR adds a new @openfn/language-browserless adaptor for converting HTML or URLs to PDF using Browserless.
The goal is to provide a practical alternative to pdfshift with higher free-tier limits and more flexible PDF options, as discussed in #1319.

## AI Usage
- [ ] This PR was created with the assistance of copilot. All changes were reviewed and tested by a human maintainer

Thanks for reviewing

### DIFF
--- a/packages/browserless/README.md
+++ b/packages/browserless/README.md
@@ -20,13 +20,13 @@ for required and optional `configuration` properties.
 
 ## Exports
 
-- `createPDF(input, options)` — Operation that calls the `/pdf` endpoint and returns a base64 string for binary responses.
-- `request(method, path, options)` — Generic Browserless-authenticated HTTP operation.
+- `createPDF(input, options)`:  Operation that calls the `/pdf` endpoint and returns a base64 string for binary responses.
+- `request(method, path, options)`:  Generic Browserless-authenticated HTTP operation.
 
 ## browserless Configuration
 
-- `state.configuration.baseUrl` (optional) — Base URL for Browserless (default: `https://production-sfo.browserless.io`).
-- `state.configuration.token` (required) — Browserless API token for authentication.
+- `state.configuration.baseUrl` (optional): Base URL for Browserless (default: `https://production-sfo.browserless.io`).
+- `state.configuration.token` (required):  Browserless API token for authentication.
 
 ---
 


packages/browserless/src/Utils.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const safePath = nodepath.join(path);
+  const wantsPdfBase64 = options && options.forcePdfBase64;
+  if (wantsPdfBase64) {
+    opts.parseAs = 'base64';
+  }
+
+  return commonRequest(method, safePath, opts).then(response => {
+    try {
+      const contentType = (response && response.headers && (response.headers['content-type'] || response.headers['Content-Type'])) || '';
```
josephjclark: undici - our http library - should be returning all headers as lowercase. We should be able to significantly reduce the complexity of this code.

You can also use conditional chaining - `consdt contentType = response?.headers?.['content-type']. But you should ALWAYS get a response object back from request, and you it should always have a headers object

triskyi:
@josephjclark   I have refactored the Browserless adaptor to expose a minimal public interface (http.request and createPDF), documented the forcePdfBase64 option, and added JSDoc.

josephjclark (changes requested):

triskyi:
Hi @josephjclark,  today I made updates based on the feedback you provided. I have addressed the suggested changes and would appreciate it if you could review them again.

Thank you.

packages/browserless/src/Adaptor.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Generic Browserless-authenticated HTTP request operation.
+ * @public
+ * @param {string} method - HTTP method
+ * @param {string} path - URL or relative path
+ * @param {object} options - Request options
+ * @returns {Operation} Returns state with `data` and `response`
+ */
+export function request(method, path, options) {
```
josephjclark: This function seems to be duplicated here and in `http`. We only need one public request function and I'd prefer the `http` one :)

packages/browserless/src/Adaptor.js
```diff
@@ -0,0 +1,92 @@
+import { expandReferences } from '@openfn/language-common/util';
+import * as util from './Utils.js';
+import { request as lowLevelRequest } from './Utils.js';
+
+/**
+ * Create a PDF from HTML or URL.
+ * @public
+ * @param {string|object} input - HTML string, URL string, or { html } / { url }
```
josephjclark: I'm still not sure why we're supporting the `{html}` or `{url}` objects here. What is the advantage to the user of doing it this way? Happy to keep it so long as there's a reason

triskyi:
Hi @josephjclark  i have fixed request duplication, PDF handling, schema, and updated tests
Thanks.

packages/browserless/assets/square.png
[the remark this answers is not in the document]
josephjclark: These images are still quite low quality - are you able to get better resolution and fit?

josephjclark (commented):
Sorry just a couple of little details to go

triskyi:
> Thanks @triskyi - I'll try and get this done today
> 
> Just one little bit of general feedback - how are you generating your commit messages? Just tracking the files isn't a terribly useful way for me to understand changes as they come up, or to understand the history of the work at a glance. Those messages are actually useful metadata for reviewers (as well as for understanding the history of the codebase when tracking bugs and changes over time)
> 
> A good way to level up your dev skills would be to work on those commit messages. They shouldn't be super rich - a single sentence is usually fine (don't fill your commits with AI bloat either!).

Thank you, I will add comments on them on how I have solved them

PR #1567, opened 2026-02-25
dhis2: Add  pagination to `get()`

### LINKED ISSUE #1546: dhis2: add pagination support
### Description

Need to add pagination support in `get` to allow for getting large amounts of data

hunterachieng:
## Summary

Implemented pagination capabilities in the `get()` function.

Use `limit` for the exact number of items to be returned; default is 10000
Use `offset` for the page number to start with; default is 1
Add a callback to enable handling of data returned during each pagination call

Fixes #1546 

## AI Usage

know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


packages/dhis2/src/Adaptor.js
```diff
@@ -247,9 +248,13 @@ export function create(path, data, params = {}) {
  * get('programs', { orgUnit: 'TSyzvBiovKh', fields: '*' });
  * @example <caption>Get a single tracked entity given the provided ID. See [TrackedEntities docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/tracker.html#tracked-entities-get-apitrackertrackedentities)</caption>
  * get('tracker/trackedEntities/F8yKM85NbxW');
+ * @example <caption>Get only 2 programs for an organization unit</caption>
+ * get('programs', { orgUnit: 'TSyzvBiovKh', limit: 2 });
  * @function
  * @param {string} path - Path to the resource
  * @param {object} params - Object of query parameters to include in the request
+ * @options {number} [options.limit] - Optional limit to the number of records returned. By default, this is 10,000.
```
josephjclark: hmm. These docs are a little strange because they sort of imply that limit and offset are the only options. But you can put in any dhis2 query parameter here right?

I'll come back to this later and decide what to do

josephjclark (commented):
I've started a review but I need to put this down today as it sounds like you're not blocked on this.

I don't think the pagination rules are right though - so we need to take a real close look at that.

I also wonder if we need to add a callback like `onPageDownloaded`, and trigger that callback every time we finish a page. That makes it a lot easier for you to process data in batches, rather than downloading everything in one huge hit (this is a problem that affects a lot of our pagination APIs, but maybe this is an opportunity to start a new design pattern)

mtuchi:
Okay @hunterachieng here are the examples
```
// Fetching all events without pagination
// Be cautious: This can degrade performance if you have a lot of events, use with caution
// v2.41+ for older version `skipPaging: true`
tracker.export('events', { paging: false });

// Fetching all events with pagination
// DHIS2 Defaults: paging: true, pageSize: 50
tracker.export('events', { totalPages: true, pageSize: 1e4 });
fn(state => {
  state.results = state.data.events;
  const { page, pageSize, pageCount, total } = state.data.pager;
  const remainingPages = pageCount - page;

  state.pages = Array.from({ length: remainingPages }, (_, i) => page + i + 1);
  state.pageSize = pageSize;
  return state;
});

each(
  $.pages,
  tracker
    .export('events', { pageSize: $.pageSize, page: $.data })
    .then(state => {
      state.results = state.results.concat(state.data.events);
      return state;
    }),
);
```
i don't think you should do the first one especially in the dhis2 demo server because it has lot's of records.

The second example is how you can fetch records in chunks aka pageSize. [Here is my loom video walk through](https://www.loom.com/share/53060a0ac4a74997adbc19e982754aef)

mtuchi (changes requested):
See my comment here https://github.com/OpenFn/adaptors/pull/1567#issuecomment-4488666651

hunterachieng:
Closing this as we are not implementing pagination anymore

Issue #1588, opened 2026-03-10
fhir-gen: put a space in between profile names


josephjclark:
A small docs issue

In a builder function, we list the valid types for a the `type` string

<img width="905" height="301" alt="Image" src="https://github.com/user-attachments/assets/e2e8114d-2d27-4679-94e9-d8c5735e3456" />

But when we generate the list, there's no spaces. So HTML renders the valid options as a single word with no line breaks, making it hard to read.

Also should we embed each value in a `code` tag?

PR #1589, opened 2026-03-11
feat: change builders map

### LINKED ISSUE #1587: fhir-gen: in docs, list props alphabetically
If you look at the docs for a generated fhir adaptor, you'll see that the properties for each builder are shown in a random order:

Image

This makes it super hard to understand and find the properties available.

So we should sort them alphabetically.

I don't know whether we should do the sorting just in the JSDocs generation bit, or if we should do it in the schema. I suspect just the jsdocs but I am not sure.

I suppose there's also an option to do this in the doc generator, rather than just for fhir, so that all objects are sorted alphabetically. But that might be too big, and sometimes maybe we have a particular order of options we want to keep. So let's just do it for fhir for now.

hunterachieng:
## Summary

Format type in builder function to add spaces in between and wrap in code formatter.

New output:

```
A profile type: one of `SzCauseOfDeath`, `SzClinicalObservation`, `SzLabResult`, `SzMannerOfDeath`, `SzVitalSigns`

```
Alphabetically organise generated props

Fixes #1587 , #1588

## Details

## AI Usage

know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/bright-terms-pump.md
+++ b/.changeset/bright-terms-pump.md
@@ -1,8 +1,0 @@
----
-'@openfn/language-fhir-eswatini': patch
-'@openfn/generate-fhir': patch
-'@openfn/language-fhir-4': patch
----
-
-Alphabetically sort generated props
-Correctly format generated types with a space and a surrounding code block: '``'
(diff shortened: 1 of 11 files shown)


josephjclark (approved):
Works great! Please rebuild and commit changes

No need to re-run this for other fhir adaptors

Please add a changeset

PR #1651, opened 2026-04-19
Update collections adaptor to include project ID

josephjclark:
## Summary

Changes to how Lightning handles collections requires the collections adaptor to include `project_id` in the request.

Very simply PR actually - most of the logic is in an update in the mock to ensure unit tests are consistent with the new design. But the actual adaptor changes very little.

This requires a change in the worker to send `project_id` on the credential. But if it doesn't, nothing breaks.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/packages/collections/CHANGELOG.md
+++ b/packages/collections/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Patch Changes
 
 - 0df6720: Add docs for a missing argument on `keygen` param
+- Add support for `project_id` on config
 
 ## 0.8.5 - 07 April 2026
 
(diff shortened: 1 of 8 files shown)


josephjclark:
I've tested this adaptor again the worker and lightning on main. Works great. No risk of breakage there (literally no behaviour different until the credential is modified)

josephjclark:
When hitting a conflict in lightning, the old adaptor on main says this:
```
 {
  "code": 409,
  "description": "The server returned an error, see attached message",
  "message": "{\"error\":\"Multiple collections found with this name. Add ?project_id=<uuid> to disambiguate.\"}"
}
```
Bit of a misleading error in the adaptor, but a reasonable one from lightning. I think we have to live with it - I'm not patching the adaptor to add handling for this


But THIS adaptor can handle the error. The error would only occur if the project-id has not automatically been fed through to the worker.

PR #1655, opened 2026-04-24
gsheets: Update input data format and allow for multiple range updates

### LINKED ISSUE #1476: GoogleSheets adaptor should support multi-range update and throttling
### Problem
The current batchUpdateValues implementation only supports updating a single range at a time. When users need to update multiple non-contiguous ranges, they must call batchUpdateValues multiple times, which:

- Quickly exhausts Google's rate limit (60 writes/minute/user)
- Defeats the purpose of the batch API
- Causes failures in parallel workflow scenarios

### Current Behavior
```
batchUpdateValues({
  spreadsheetId: '1abc...',
  range: 'Sheet1!A1:B2',
  values: [['a', 'b'], ['c', 'd']],
  valueInputOption: 'RAW'
});
```
This sends a single ValueRange object to the API.
### Proposed Behavior
Support a data parameter that accepts an array of ValueRange objects, aligning with Google's [spreadsheets.values.batchUpdate](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/batchUpdate) API:

```
batchUpdateValues({
  spreadsheetId: '1abc...',
  data: [
    { range: 'Sheet1!A1', values: [['value1']] },
    { range: 'Sheet1!B5', values: [['value2']] },
    { range: 'Sheet1!D10:E11', values: [['a', 'b'], ['c', 'd']] }
  ],
  valueInputOption: 'RAW'
});
```

### Problem
Google Sheets API enforces a limit of 60 write requests per minute per user. When multiple OpenFn workflows run in parallel or a single workflow makes rapid sequential calls, users hit `429 Too Many Requests` errors.

hunterachieng:
## Summary

Updated `appendValues()`, `batchUpdateValues()`, and `getValues()` to use positional arguments instead of a single params object.

Callback parameter has been removed from `appendValues()`, `batchUpdateValues()`, and `getValues()` in favor of a promise-based API.

Fixes #1476 

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

- [x] If this PR includes breaking changes, do we need to update any jobs in


packages/googlesheets/src/Adaptor.js
```diff
@@ -147,13 +148,25 @@ export function appendValues(params, callback = s => s) {
  *     ['Really now!', '$100', '1', '3/20/2016'],
  *   ],
  * })
+ * @example
+ * <caption>Update multiple ranges</caption>
+ * batchUpdateValues({
```
mtuchi: i am tempted to suggest a different signature 
`batchUpdateValues(sheetId, data: [{}], options)`
For single range entry,

```js
batchUpdateValues(
  "1O-a4_RgPF_p8W3I6b5M9wobA3-CBW8hLClZfUik5sos",
  [{ range: "", value: "" }],
  { valueInputOption: "RAW" },
);
```

For multiple 
```js
batchUpdateValues(
  "1O-a4_RgPF_p8W3I6b5M9wobA3-CBW8hLClZfUik5sos",
  [
    { range: "", value: "" },
    { range: "", value: "" },
  ],
  { valueInputOption: "RAW" },
);
```

i think that signature have a better seperation of concern. 
Of-course it's a major version bump but i think it's way better than having a data key in the params object

@josephjclark do you agree with that signature i am proposing ?

mtuchi (changes requested):
I have asked joe, for his opinion on new signature 
`batchUpdateValues(sheetId, data =  [{range, values}], options = {valueInputOption})`

If he agrees i think, we should update `appendValues` as well and remove support for callback in `getValues()`

.changeset/new-things-change.md
```diff
@@ -1,5 +1,45 @@
 ---
-'@openfn/language-googlesheets': minor
+'@openfn/language-googlesheets': major
 ---
 
-`batchUpdateValues()` now accepts a `data` array of `{ range, values }` objects, enabling multi-range updates in a single API call. The existing `range` + `values` params continue to work unchanged.
+Updated `appendValues()`, `batchUpdateValues()`, and `getValues()` to use positional arguments instead of a single params object.
```
mtuchi: I don't know if this makes it better or worse, but the `positional argument` threw me off a bit. I would love to see before and after for each function 

For example
**Updated `appendValues()`, `batchUpdateValues()`, and `getValues()` to use positional arguments instead of a single params object**

- `appendValues(sheetId, data, options)` — previously `appendValues({}, callback)`
- `batchUpdateValues(sheetId, data, options)` — previously `batchUpdateValues({}, callback)`
- `getValues(sheetId, range)` — previously `getValues(sheetId, range, callback)`

Callbacks have been removed across all three methods in favor of a promise-based API.


Keep the Migration Guide it's good as is

packages/googlesheets/src/Adaptor.js
```diff
@@ -78,26 +78,36 @@ export function execute(...operations) {
...
+ * @param {Object} data - Data to append.
+ * @param {string} data.range - The range to append to.
+ * @param {array} data.values - A 2d array of values to append.
+ * @param {Object} [options] - Optional settings.
+ * @param {string} [options.valueInputOption] - Defaults to 'USER_ENTERED'.
+ * @param {function} [callback] - Optional callback function.
  * @returns {Operation}
  */
-export function appendValues(params, callback = s => s) {
+export function appendValues(spreadsheetId, data, options = {}, callback = s => s) {
```
mtuchi: Maybe we should have a ticket to convert `data` to be an array of object similar to how `batchValuesUpdate()` works now.

packages/googlesheets/src/Adaptor.js
```diff
@@ -123,12 +124,10 @@ export function appendValues(params, callback = s => s) {
           } else {
             console.log('Success! Here is the response from Google:');
             console.log(response.data);
```
mtuchi, line 127: @hunterachieng is it possible to confirm that this log does not print access_token ?

hunterachieng: Confirming

hunterachieng: @mtuchi  It does not return `access_token`

packages/googlesheets/test/index.js
```diff
@@ -40,31 +42,119 @@ describe('execute', () => {
...
-      values: state.values,
-    })(state);
+  it('should return early if the values array is undefined or nullish', async() => {
+    const state = { data: [] };
+
+    const result = await appendValues(
+      '123-456-789',
+      [{ range: 'Sheet!A1:E1', values: state.values }],
+    )(state);
     expect(result).to.eql(state);
```
mtuchi: maybe just simplify to 
```js
const result = await appendValues(
      '123-456-789',
      [{ range: 'Sheet!A1:E1', values: null}],
    )(state);
```

mtuchi (commented):
@hunterachieng i think the changes are looking good, can you tidy up two things
- Create an issue supporting array of data in `appendValues(id, data = [{}], options)` to match what we have in `batchUpdateValues`
- Clean up PR description and title to match the changes and don't forget to update the review checklist as well

packages/googlesheets/src/Adaptor.js
```diff
@@ -78,26 +78,27 @@ export function execute(...operations) {
...
   return state => {
-    const [resolvedParams] = expandReferences(state, params);
-    const { spreadsheetId, range, values } = resolvedParams;
+    const [resolvedSpreadsheetId, resolvedData, resolvedOptions] = expandReferences(
+      state,
+      spreadsheetId,
+      data,
+      options
+    );
+    const { range, values } = resolvedData[0];
```
mtuchi: Hey @hunterachieng, please make `resolvedData` an object and not an array for now, because it doesn't support multiple range yet and people might think you can pass multiple range. 

If you can add support for appending multiple ranges then it's fine but since it doesn't support multiple range, lets revert back to a an object and create an issue to make

hunterachieng: @mtuchi  It is fixed

packages/googlesheets/src/Adaptor.js
```diff
@@ -78,26 +78,27 @@ export function execute(...operations) {
...
   return state => {
-    const [resolvedParams] = expandReferences(state, params);
-    const { spreadsheetId, range, values } = resolvedParams;
+    const [resolvedSpreadsheetId, resolvedData, resolvedOptions] = expandReferences(
+      state,
+      spreadsheetId,
+      data,
+      options
+    );
+    const { range, values } = resolvedData;
```
mtuchi: The `appendDataValues` doesn't support multiple ranges ? is this intentional ? if it is we should update the pr notes

packages/googlesheets/src/Adaptor.js
```diff
@@ -139,56 +138,57 @@ export function appendValues(params, callback = s => s) {
...
- *     ['From expression', '$15', '2', '3/15/2016'],
- *     ['Really now!', '$100', '1', '3/20/2016'],
+ * <caption>Update a single range</caption>
+ * batchUpdateValues(
+ *   '1O-a4_RgPF_p8W3I6b5M9wobA3-CBW8hLClZfUik5sos',
+ *   [{ range: 'Sheet1!A1:E1', values: [['From expression', '$15'], ['Really now!', '$100']] }],
+ *   { valueInputOption: 'RAW' }
+ * )
+ * @example
+ * <caption>Update multiple non-contiguous ranges</caption>
```
mtuchi: @hunterachieng what does `non-contiguous` ranges mean ?

mtuchi:
Hey @hunterachieng sorry for long review, i wanted to confirm few things on `appendValues` function before approving this pr. Here is my findings 

The current implementation does not multiple range when using `appendValues()`, See issue https://github.com/OpenFn/adaptors/issues/1656. And i don't think it's an easy fix you will have to change the entire payload on `client.spreadsheets.values.append()` to look something like this 
```
{
  "[valueInputOption](https://developers.google.com/workspace/sheets/api/reference/rest/v4/ValueInputOption)": "VALUE_INPUT_OPTION",
  "[data](https://developers.google.com/workspace/sheets/api/reference/rest/v4/spreadsheets.values#resource-valuerange)": [
    {
      "range": "Sheet1!A1:A4",
      "[majorDimension](https://developers.google.com/workspace/sheets/api/reference/rest/v4/spreadsheets.values#dimension)": "COLUMNS",
      "values": [
        ["Item", "Wheel", "Door", "Engine"]
      ]
    },
    {
      "range": "Sheet1!B1:D2",
      "[majorDimension](https://developers.google.com/workspace/sheets/api/reference/rest/v4/spreadsheets.values#dimension)": "ROWS",
      "values": [
        ["Cost", "Stocked", "Ship Date"],
        ["$20.50", "4", "3/1/2016"]
      ]
    }
  ]
}
```

[See writing multiple range docs](https://developers.google.com/workspace/sheets/api/samples/writing#multiple-ranges). I honestly don't think we need to support multiple range in `appendValues` because `batchUpdateValues` function already does that.  I will edit the pr notes and approve this pr

mtuchi:
[These examples from google](https://github.com/googleworkspace/node-samples/tree/main/sheets) are really useful, It's was really hard to find a good documentation on how to use the googleapis sheets api in node

mtuchi (approved):
@hunterachieng I have update pr notes and remove this issue from the notes https://github.com/OpenFn/adaptors/issues/1656,  See [my comments here](https://github.com/OpenFn/adaptors/pull/1655#issuecomment-4620722914). 

I am happy to handover this issue to joe for final round of review

packages/googlesheets/src/Adaptor.js
```diff
@@ -74,30 +74,31 @@ export function execute(...operations) {
...
  * @example
- * appendValues({
- *   spreadsheetId: '1O-a4_RgPF_p8W3I6b5M9wobA3-CBW8hLClZfUik5sos',
- *   range: 'Sheet1!A1:E1',
- *   values: [
- *     ['From expression', '$15', '2', '3/15/2016'],
- *     ['Really now!', '$100', '1', '3/20/2016'],
- *   ],
- * })
+ * appendValues(
```
josephjclark, line 81: If we're moving towards required positionals, shouldn't the signature be `appendValues(id, range, values, options)` ? @mtuchi 

I know it's different for the batch API but the batch API is different :)

packages/googlesheets/src/Adaptor.js
josephjclark: Sorry for the late reply. Yes, this looks like a much cleaner API. I'm just going through the details now.

josephjclark:
Re appendValues: yes, appendValues should take a single range/value set, and bulk should take multiple.

As the original issue hints at, I don't understand the difference really between `client.spreadsheets.append` and `client.spreadsheets.batchUpdate()`? Can anyone explain that to me? It is just that append only takes a single range and batch accepts multiple ranges? Or is there more going on in batch?

packages/googlesheets/test/index.js
```diff
@@ -40,31 +42,119 @@ describe('execute', () => {
...
-      values: state.values,
-    })(state);
+  it('should return early if data is empty', async() => {
+    const state = { data: [] };
+
+    const result = await batchUpdateValues('123-456-789', [])(state);
     expect(result).to.eql(state);
   });
+
+  describe('with mocked Google Sheets client', () => {
```
josephjclark: I like the mocked tests (but is this way out of scope?) but we shouldn't have a describe block just for the mocked tests. Also the pattern is inconsistent with `getValues()`, because the test there just uses the mock without shouting from the rooftop about it.

Looking at the prior tests, they don't really test anything. They might catch an exception I suppose.

I would be tempted to remove all the old no-op tests entirely, and instead just add your new tests (at least one test per function though).

It is confusing to have the old un-mocked tests which don't do anything PLUS the new mocked tests which actually have value. Let's clean it up.

packages/googlesheets/src/Adaptor.js
```diff
@@ -139,56 +138,57 @@ export function appendValues(params, callback = s => s) {
 /**
  * Batch update values in a Spreadsheet.
  * @example
- * batchUpdateValues({
- *   spreadsheetId: '1O-a4_RgPF_p8W3I6b5M9wobA3-CBW8hLClZfUik5sos',
- *   range: 'Sheet1!A1:E1',
- *   values: [
- *     ['From expression', '$15', '2', '3/15/2016'],
- *     ['Really now!', '$100', '1', '3/20/2016'],
+ * <caption>Update a single range</caption>
```
josephjclark: It would be nice for the single range API to take a single object, rather than an array of one. Just a little API convenience

mtuchi: @josephjclark can you say more 🤔 , values are rows data do you mean we should pass `values: [{columnHeader: value, colHeader: value}, {...}]`?

josephjclark: So multiple ranges is:
```
batchUpdateValues(id, [set1, set2])
```

Where a set is `{ range, values }`
And a single range so far is:
```
batchUpdateValues(id, [set1])
```

I'm saying allow a single range to be passed like this:
```
batchUpdateValues(id, set1)
```

BUT THEN AGAIN

That makes the API the same as the `appendValues` API, which only confuses the two.

Ok, so we should force data to be an array of ranges. But do not show an example with an array of just 1 range because it's wierd

mtuchi:
> difference really between `client.spreadsheets.append` and `client.spreadsheets.batchUpdate()`

 `values.append`
  - Finds the last row with data in the range and adds rows after it — it never overwrites
  - The range param is really just a hint to locate which table/sheet to append to
  - Designed for "add new rows to a growing list" use cases (logs, form submissions, etc.)
  - Single range only, yes

  `values.batchUpdate`
  - Writes to exact cell coordinates — it overwrites whatever is there
  - The range param means "put data starting at exactly this cell"
  - Multiple ranges in one API call (hence "batch")
  - Designed for "set these cells to these values" use cases

packages/googlesheets/src/Adaptor.js
mtuchi: I think function naming is what causing the confusion here. There are singular actions and plural actions
For example: 
- If we had `appendValue(id, range, data, opt)` this is clear, add new row to gsheet
- For `appendValues(id, [{set0, set1}], opt)`, this is clear adding multiple rows

The same could apply for `batchUpdateValues()`
- `batchUpdate(id, range, value,opt)` 
- `batchUpdate(id, [{set0}, {set1}])`

Of course adding good docs that explain what append does, and what does batchUpdate do? because their different. 

I will remove the single batch update example

packages/googlesheets/src/Adaptor.js
mtuchi: I have updated the function to be `appendValues(id, range, values, options)`

packages/googlesheets/test/index.js
mtuchi: I have cleaned up, the tests

josephjclark (approved):
Ok cool, I think this is fine. Can @hunterachieng or @mtuchi please bump versions in the morning and merge

PR #1678, opened 2026-05-26
fhir: map primitive elements

hunterachieng:
## Summary

Add ability to map primitive elements in `fhir` by focusing on elements starting with an `_`.
Regenerated `fhir-eswatini` with tests to check behavior

- Ensured we do not throw away primitive elements in `generate-schema` then mapped by checking if the parent has an extension.

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


packages/fhir-eswatini/test/resources/Patient.test.ts
```diff
@@ -439,4 +439,21 @@ describe('SzPatient', () => {
       valueDateTime: '2025-06-01T10:00:00Z',
     });
   });
+  it('should map primitive values', ()=>{
+  const state = {};
+
+  const resource = b.patient('SzPatient',{
+    birthDate:'10/07/1990',
```
josephjclark: You should assert that birthdate is set on the resource too.

packages/fhir-eswatini/test/resources/Patient.test.ts
```diff
@@ -439,4 +439,21 @@ describe('SzPatient', () => {
       valueDateTime: '2025-06-01T10:00:00Z',
     });
   });
+  it('should map primitive values', ()=>{
```
josephjclark: This isn't a primitive value though, it's a primitive extension. A primitive value is just `gender: 'male'`, which we already support right?

This work is about primitive extensions

packages/fhir-eswatini/src/profiles/SzVitalSigns.ts
```diff
@@ -4,43 +4,43 @@
 // SEE THE README FILE FOR DETAILS
 
 import _ from "lodash";
-import { b as dt } from "@openfn/language-fhir-4";
-import type { builders as FHIR } from "@openfn/language-fhir-4";
+import * as dt from "../datatypes";
+import type * as FHIR from "../fhir";
 type MaybeArray<T> = T | T[];
 
 export type Observation_SzVitalSigns_Props = {
```
josephjclark, line 11: Nervous about this diff! This looks unrelated to the work. And I think all the types here are actually broken?

packages/fhir-eswatini/types/builders.d.ts
```diff
@@ -1,705 +1,25 @@
...
-declare const lookupValue: (url: any, code: any) => any;
-declare const mapSystems: (obj: any) => any;
-declare const mapValues: (obj: any, hints: any) => any;
-declare const ref: (ref: any, opts?: {}) => any;
-declare const reference: (ref: any, opts?: {}) => any;
-declare const setSystemMap: (newMappings: any) => (state: any) => any;
-declare const setValues: (url: any, values: any, type?: string) => void;
-declare const value: (value: any, system: any, ...extra: any[]) => any;
-
+import "./values";
```
josephjclark: Quite nervous about this cahnge too

tools/generate-fhir/src/generate-schema.ts
```diff
@@ -343,6 +343,7 @@ async function parseProp(
   data,
 ) {
   let [parent, prop] = path.split('.');
+  const rawProp = prop;
```
josephjclark: Is `rawProp` actually needed?

tools/generate-fhir/test/generate-code.test.ts
```diff
@@ -177,6 +177,69 @@ run('builds typeDef with nested extension', t => {
   t.is(dt.addExtension.calls, 1);
 });
 
+run('builds primitive sibling extension from underscored slice input', t => {
```
josephjclark: I i hadn't noticed this `run` pattern before. I know it's unrelated but can we remove the alias and restore these to `test.serial`?

Actually do they even need to be serial?

josephjclark (commented):
Maybe a bit of confusion over whether we're handling primitve _types_ or primitive _extensions_? I'm expecting that primitive types work already, and we're just adding support for extensions here

packages/fhir-eswatini/src/profiles/SzVitalSigns.ts
hunterachieng: @josephjclark  I used this command to regenerate the schema `pnpm generate-fhir fhir-eswatini --respec` could this be the cause of the changes? Because I got this error when testing and  I had to export it from builders in `fhir-4`
` TypeError: dt.ensureConceptText is not a function`

josephjclark: Strange :thinking:

When I rebuild from the root I also get strange type errors and wierd syntax. When I go into the adaptor folder and do `pnpm build:src --respec`, I get different output.

The two commands really should be aliases of each other. I fear some path has gotten misaligned and is doing wierd things.

if you go to main and run `pnpm build:src --respec` from inside the adaptor folder, what happens?

hunterachieng: Screenshot 2026-05-27 at 16 21 13

@josephjclark  It is only this that is updated ` "adaptorGeneratedDate": "2026-05-27T13:18:19.262Z",`

josephjclark: That's what I get too.

So I think we have a bug when generating from the repo root. Or maybe it's a bug when generating inside the folder. Either way that needs investigating.

Maybe you can run `pnpm build:src --respec` on this branch and commit the changes - that should resolve some of the issues I see on this PR.

hunterachieng: @josephjclark When I do this, the changes from `/tools/generate-fhir` do not get picked up

josephjclark: You mean the primitive extension no longer gets added? The test fails?

hunterachieng: It is not added, and the tests fail. I had to skip them

josephjclark: Don't skip them - if tests are failing its because something is broken. They should fail in CI until we fix the problem.

Something seems very messed up here. I'll try and take a look tomorrow but if I can't, it'll be Monday before I can investigate.

`npm generate-fhir fhir-eswatini --respec` and `pnpm build:src --respec` should be _exactly_ the same command. I'm really nervous that we seem to be getting different behaviours. Can you try and investigate what might be different when running the commands?

There's no magic here, just code. All the answers are in this repo .

I suspect something is wrong on main, and we need to update `fhir-eswatini` on a branch off main first, and then rebase your work here onto that branch. That would help me understand what's wrong on main, and what you've added on this PR. That will make my review a lot easier.

hunterachieng: Sure. On it

hunterachieng: @josephjclark  using `npm generate-fhir fhir-eswatini --respec` does not work. It flags that `npm error Missing script: "generate-fhir"`
When I try with pnpm when inside `fhir-eswatini`, I get these changes
Screenshot 2026-05-29 at 12 17 42

josephjclark: You need to use pnpm, not npm.

It might just be that eswatini is behind on the changes and wasn't updated when we last touched the generator.

I don't have time to investigate this myself. You'll need to diagnose the problem, understand this diff, and explain it back to me. If you have specific questions I can try to answer them on slack

tools/generate-fhir/test/generate-code.test.ts
```diff
@@ -68,15 +68,15 @@ const generateBuilder = (resourceType: any, props: any) => {
   return profiles[resourceType];
 };
 
-run('sets resourceType', t => {
+test.serial('sets resourceType', t => {
```
josephjclark: why are these serial now? is that deliberate? Each test should use isolated resources so I'm not sure why we're forcing serial mode now

packages/fhir-eswatini/test/resources/Patient.test.ts
```diff
@@ -439,4 +439,23 @@ describe('SzPatient', () => {
       valueDateTime: '2025-06-01T10:00:00Z',
     });
   });
+  it('should map primitive extensions', ()=>{
```
josephjclark: This is a generic test name. But it should be specific to patient.

So you need two tests here: one for `_birthdate` and one for `_birthtime`. You'e explicitly checking that those two properties map. You are _not_ generically checking that primitive exetnsions work (the generic test should happen elsewhere, in the generator)

Also the formatting on this is weird

tools/generate-fhir/test/generate-code.test.ts
```diff
@@ -177,7 +177,70 @@ run('builds typeDef with nested extension', t => {
   t.is(dt.addExtension.calls, 1);
 });
 
-run('skips nil properties', t => {
+test.serial('builds primitive sibling extension from underscored slice input', t => {
+  const profile = {
+    birthDate: {
+      type: ['date'],
+      isArray: false,
+      typeDef: {
```
josephjclark, line 190: I am little surprised as to the modelling of this in the schema. Why is `birthTime` a subtype of `birthDate`? The two are unrelated. Usually the typedef is used to describe an object type.

So two things:
1. This is an incredibly valuable test because it quickly helped me understand the implementation and how primitive extensions are handled in schema. Well done and thank you!
2. I think primitive extensions should map to a top level schema property like `_birthTime`. Not nested like this.

josephjclark (changes requested):

tools/generate-fhir/test/generate-code.test.ts
hunterachieng: @josephjclark  you had mentioned this comment before. 
```
I i hadn't noticed this run pattern before. I know it's unrelated but can we remove the alias and restore these to test.serial?
```

I can update and return it to `run` instead

josephjclark: Oh! Sorry Hunter, I got myself confused. `run` was an alias for `test.serial`.

I don't like the alias but we also shouldn't need `.serial`. It's all a bit weird.

I'm pushing a commit which sets these tests up correctly. Sorry for the confusion!

packages/fhir-eswatini/src/builders.ts
```diff
@@ -573,6 +573,7 @@ export function organization(type: any, props?: any) {
   * @public
   * @function
   * @param {object} props - Properties to apply to the resource (includes common and custom properties).
+  * @param {} [props._birthDate] - undefined
```
josephjclark, line 576: `undefined` in the docs

Shouldn't we also have a `_birthTime` property declared here?

packages/fhir-eswatini/test/resources/Patient.test.ts
```diff
@@ -439,4 +439,39 @@ describe('SzPatient', () => {
       valueDateTime: '2025-06-01T10:00:00Z',
     });
   });
+
+  it('should map _birthDate primitive extension shorthand', () => {
+    const resource = b.patient('SzPatient', {
+      birthDate: '10/07/1990',
+      _birthDate: '2000-01-01T14:35:45-05:00',
```
josephjclark, line 446: Can we add another test here that if you pass a full extension as _birthDate, it'll work as we expect and skip the shorthand stuff?

If that test fails we might have to think about the implementation. I hope it'll just work.

josephjclark (approved):

PR #1679, opened 2026-05-27
Security Updates

josephjclark:
Sundry security updates, see changelogs


josephjclark:
Typescript build has suddenly started failing on the adaptor-apis tool. Nothing has really changed that would affect that. Reverting adaptor-apis doesn't fix. All works fine locally :(

PR #1681, opened 2026-05-28
dhis2: Add paging examples and default `async:false` in `tracker.import`

### LINKED ISSUE #1546: dhis2: add pagination support

hunterachieng:
## Summary
Default `async:false` in `tracker.import` and add  paging examples to `tracker.export()`

Fixes #1546 

Related 
- https://github.com/OpenFn/docs/pull/797

## Details
- Default `async:false` in `tracker.import` function
- Add paging examples in `tracker.export` function
- Add integration tests for `tracker.export` pagination 
- Update unit tests for `tracker.import` (testing async modes), and `tracker.export` testing pagination params

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/hip-planes-see.md
+++ b/.changeset/hip-planes-see.md
@@ -2,6 +2,6 @@
 '@openfn/language-dhis2': minor
 ---
 
-Add async option to `tracker.import` and 
-Add paging examples to
-`tracker.export()`
+Add async option to `tracker.import`.
+Improve `tracker.export` docs by adding a pagination example and
+linking pagination query parameters.
--- a/.changeset/hip-planes-see.md
+++ b/.changeset/hip-planes-see.md
@@ -1,6 +1,7 @@
 ---
-'@openfn/language-dhis2': patch
+'@openfn/language-dhis2': minor
 ---
 
-Default `async:false` in `tracker.import` and add paging examples to
+Add async option to `tracker.import` and 
+Add paging examples to
 `tracker.export()`
--- a/.changeset/hip-planes-see.md
+++ b/.changeset/hip-planes-see.md
@@ -2,5 +2,5 @@
 '@openfn/language-dhis2': patch
 ---
 
-Default `async:false` in `tracker.import` and paging examples to
+Default `async:false` in `tracker.import` and add paging examples to
 `tracker.export()`
--- a/.changeset/hip-planes-see.md
+++ b/.changeset/hip-planes-see.md
@@ -2,4 +2,5 @@
 '@openfn/language-dhis2': patch
 ---
 
-Fix tracker exports and imports when no query object is provided by defaulting `async` to `false`
+Default `async:false` in `tracker.import` and paging examples to
+`tracker.export()`
--- a/.changeset/hip-planes-see.md
+++ b/.changeset/hip-planes-see.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-dhis2': patch
+---
+
+Fix tracker exports and imports when no query object is provided by defaulting `async` to `false`
--- a/packages/dhis2/src/tracker.js
+++ b/packages/dhis2/src/tracker.js
@@ -58,7 +58,12 @@ function _import(strategy, payload, options = {}) {
     const [resolvedStrategy, resolvedPayload, resolvedOptions] =
       expandReferences(state, strategy, payload, options);
 
-    const { apiVersion, parseAs, async: asyncOption, ...query } = resolvedOptions;
+    const {
+      apiVersion,
+      parseAs,
+      async: asyncOption,
+      ...query
+    } = resolvedOptions;
 
     const response = await util.request(state.configuration, {
       method: 'POST',
@@ -68,7 +73,7 @@ function _import(strategy, payload, options = {}) {
           ...resolvedOptions,
           resolvedStrategy,
         },
-        'tracker'
+        'tracker',
       ),
       options: {
         apiVersion,
@@ -96,9 +101,28 @@ export { _import as import };
  * @example <caption>Export all enrollment resources</caption>
  * tracker.export('enrollments', {orgUnit: 'TSyzvBiovKh'});
  * @example <caption>Export all events</caption>
- * tracker.export('events')
+ * tracker.export('events', { paging: false})
  * @example <caption>Export all events with pagination</caption>
- * tracker.export('events', { page: 1, pageSize: 100 });
+ * tracker.export('events', { totalPages: true, pageSize: 1e4 });
+ * fn(state => {
+ *   state.results = state.data.events;
+ *   const { page, pageSize, pageCount, total } = state.data.pager;
+ *   const remainingPages = pageCount - page;
+ *
+ *   state.pages = Array.from({ length: remainingPages }, (_, i) => page + i + 1);
+ *   state.pageSize = pageSize;
+ *   return state;
+ * });
+ *
+ * each(
+ *   $.pages,
+ *   tracker
+ *     .export('events', { pageSize: $.pageSize, page: $.data })
+ *     .then(state => {
+ *       state.results = state.results.concat(state.data.events);
+ *       return state;
+ *     }),
+ * );
  * @function
  * @param {string} path - Path to the resource, relative to the /tracker endpoint
  * @param {object} query - An object of query parameters to be encoded into the URL. Can include pagination parameters, filters, etc.
@@ -115,7 +139,7 @@ function _export(path, query = {}, options = {}) {
       state,
       path,
       query,
-      options
+      options,
     );
 
     const response = await util.request(state.configuration, {
@@ -123,7 +147,7 @@ function _export(path, query = {}, options = {}) {
       path: util.prefixVersionToPath(
         state.configuration,
         resolvedOptions,
-        `tracker/${resolvedPath}`
+        `tracker/${resolvedPath}`,
       ),
       options: {
         ...resolvedOptions,
--- a/packages/dhis2/src/tracker.js
+++ b/packages/dhis2/src/tracker.js
@@ -121,7 +121,11 @@ export { _import as import };
  * @function
  * @param {string} path - Path to the resource, relative to the /tracker endpoint
  * @param {object} query - An object of query parameters to be encoded into the URL. Can include pagination parameters, filters, etc.
- * @param {boolean} [query.async=false] - Whether to perform the export asynchronously. Defaults to false.
+ * @param {number} [query.page=1] - Page number to return
+ * @param {number} [query.pageSize=50] - Number of results per page
+ * @param {boolean} [query.totalPages=false] - Whether to return total number of elements and pages
+ * @param {boolean} [query.paging=true] - Set to false to return all rows without paging
+ * @param {string} [query.order] - Comma-separated field:sortDirection pairs, e.g. `createdAt:desc`
  * @param {TrackerOptions} [options] - An optional object containing parseAs, and apiVersion for the request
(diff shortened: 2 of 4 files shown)


mtuchi (commented):
Hey @hunterachieng i have made couple of changes for dhis2 pagination
- I removed the `async:fasle` in `tracker.export`.  After looking into dhis2 docs there is no `async` for `tracker.export` only in `tracker.import`
- I have added integration tests for pagination using `tracker.export`
- update `tracker.export` to include paging examples and jsdocs
- update tracker unit tests to test `async` mode for `tracker.import` and paging params for `tracker.export`

Please have a look and let me know if you have any feedback, If not we can request review from joe

mtuchi:
@hunterachieng One more thing i have added dhis2 `tracker.export` pagination examples on docs, See https://github.com/OpenFn/docs/pull/797. Please review and leave your feedback

mtuchi:
Hey @josephjclark i have added examples on docs as well and request your review https://github.com/OpenFn/docs/pull/797

.changeset/hip-planes-see.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/language-dhis2': patch
+---
+
+Default `async:false` in `tracker.import` and add paging examples to
```
josephjclark: Changing a default is usually a breaking change (and so a major version bump) because it can change the behaviour of user's code.

Ah, but we haven't changed the default value right? What we've actually done is kept the default the same, but allowed users to set the value.

That's a really important difference!

So I think this is a minor change like `add async option to tracker.import`

packages/dhis2/src/tracker.js
```diff
@@ -47,6 +47,7 @@ import * as util from './util.js';
  * @param {string} strategy - The effect the import should have. Can either be CREATE, UPDATE, CREATE_AND_UPDATE and DELETE.
  * @param {object} payload - The data to be imported.
  * @param {TrackerOptions} [options] - An optional object containing parseAs, and apiVersion, and queries for the request
+ * @param {boolean} [options.async=false] - Whether to perform the import asynchronously. Defaults to false.
```
josephjclark: But what does this actually mean? Why would you do it? Can we link to docs anywhere to explain this?

packages/dhis2/src/tracker.js
```diff
@@ -95,37 +96,62 @@ export { _import as import };
  * @example <caption>Export all enrollment resources</caption>
  * tracker.export('enrollments', {orgUnit: 'TSyzvBiovKh'});
  * @example <caption>Export all events</caption>
- * tracker.export('events')
+ * tracker.export('events', { paging: false})
+ * @example <caption>Export all events with pagination</caption>
```
josephjclark: Ah, I've picked @mtuchi up on this example in the docs site but I can see where he's got it from

I don't like this example because a) it's too long for here and b) it seems to do the same thing twice? I'm not sure what the story is or how this code should be used

packages/dhis2/src/tracker.js
```diff
@@ -95,37 +96,62 @@ export { _import as import };
...
+ *       state.results = state.results.concat(state.data.instances);
+ *       return state;
+ *     }),
+ * );
  * @function
  * @param {string} path - Path to the resource, relative to the /tracker endpoint
- * @param {object} query - An object of query parameters to be encoded into the URL
+ * @param {object} query - An object of query parameters to be encoded into the URL. Can include pagination parameters, filters, etc.
+ * @param {number} [query.page=1] - Page number to return
+ * @param {number} [query.pageSize=50] - Number of results per page
```
josephjclark: These defaults are ours, or dhis2's? If we're going to declare them here we should probably set them in code too - because if dhis2 every changes its defaults then our docs will be out of  date (also that would be a breaking change to our workflows). They shouldn't do this without a major API version bump but fixing our own defaults is cheap and easy and safe and gives us some insulation just in case.

josephjclark (changes requested):

packages/dhis2/src/tracker.js
hunterachieng: @josephjclark  These defaults are from dhis2. Should we just add a link for pagination params?

mtuchi: @hunterachieng i think we should link to dhis2 docs https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#request-parameters-for-pagination

hunterachieng: Okay

josephjclark: That'sa  better solution. Now DHIs2 owns and documents its defaults - we're not imposing anything, and we're not making claims we can't truly honor. Thanks @mtuchi

packages/dhis2/src/tracker.js
```diff
@@ -95,37 +96,37 @@ export { _import as import };
  * @example <caption>Export all enrollment resources</caption>
  * tracker.export('enrollments', {orgUnit: 'TSyzvBiovKh'});
  * @example <caption>Export all events</caption>
- * tracker.export('events')
+ * tracker.export('events', { paging: false})
  * @function
  * @param {string} path - Path to the resource, relative to the /tracker endpoint
- * @param {object} query - An object of query parameters to be encoded into the URL
+ * @param {object} query - An object of query parameters to be encoded into the URL. Can include [pagination parameters](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#request-parameters-for-pagination), filters, etc.
+ * @param {string} [query.order] - Comma-separated field:sortDirection pairs, e.g. `createdAt:desc`
```
josephjclark: This is the only query property we're actually documenting now. Why is this one so important?

josephjclark:
@hunterachieng can you update the PR description so that it's accurate?

We seem to have now removed most of the things this PR actually added :sweat_smile:  So I'm a little unsure about the value or point of it.

- We've set `async: false` in code but wasn't this the defacto default anyway? And couldn't users always pass it on the query? So what's new?
- The PR claims to have added pagination examples - but they've now been removed (that's not what  I was asking for!). All we have now is an example with `paging: false` to disable pagination and force all items to be downloaded in a single page. Is that enough? Is that OK? The changeset says "examples" plural, but all we've currently just _changed_ one example.

I basically just want to be sure that the PR description and changeset are accurate, and that we haven't lost sight of what this PR is supposed to be about.

packages/dhis2/src/tracker.js
hunterachieng: I must have missed it. i removed it

mtuchi (changes requested):
@hunterachieng See joe's comment here https://github.com/OpenFn/adaptors/pull/1681#issuecomment-4728618397

josephjclark:
@mtuchi what do you make of all this? I just have this weird feeling that this PR doesn't actually do anything?

I think the idea was to add examples but the examples we have are very thin now. Are they really telling the right story?

Is this example really all we need? If so, fine :)

```
tracker.export('events', { paging: false})
```

Even the `async` query is just written to a query param and works out of the box today (presumably with the same value).  So it's meaningless code.

hunterachieng:
@josephjclark  The only things we have added are links for pagination and the thin examples of how its done. 

Since pagination is not handled by us, the best we can do is with the examples and doc links.

For `async:false`, it was working with the default as `false` before, we just now allowed users to change the values if needed. To be honest, it is true that there is not much going on in the PR

mtuchi:
@josephjclark 
Just to add on what changed on this pr, If we put pagination stuff aside

1. Allow user to specify `async` mode in `tracker.import`, before even if you specify in query it will always be set to `false`. Now user have the ability to set it to `true`. I think this is important we shouldn't limit user to always wait for results 
2. This pr remove `async` option in `tracker.export` because it's not supported. You will always get `events, enrollments` etc. There is not `async: true or false` for `tracker.export`. This is more of a clean up

I am totally fine moving the pagination stuff into our job library docs which is what this pr on docs is doing https://github.com/OpenFn/docs/pull/797

But i would be very cautious on advocating for `tracker.export('events', { paging: false})`.  We should express the performance implication of fetching all records by using `paging: false` because it's very noticeable when you have lots of data.

mtuchi:
I did some test on app to compare paging with query param `paging:true` vs paging with chunking 

- [Here is paging with `paging:true`](https://app.openfn.org/projects/18bd2cb5-4991-4da8-98dc-c33b3878c707/w/adb99096-63d1-458c-a56c-c9da804eb60d?job=8ccb7935-e797-4f32-8384-9e15a0b3f781&panel=editor&run=68c42f5c-184e-40f9-9fcd-0bc6fd21bf66&step=df1d7f73-4373-4258-a5fa-2012e218cbae), The run was lost after 34min, I have re-tried the run hopefully it won't fail this time. 
- [Here is paging with chunking approach](https://app.openfn.org/projects/18bd2cb5-4991-4da8-98dc-c33b3878c707/w/40266ea4-03b0-475f-a8ba-72b0f9fe5a6f?job=a0f9a629-364b-4072-8a26-dd013970b915&run=785d1546-bd2d-420c-89ac-5b0ada19d739&panel=editor&step=dd5f506c-2873-435a-b412-fe15956972aa), It took about 5min

mtuchi:
Still getting a lost run when using `tracker.export("events", { paging: false })`.  [See lost run here](https://app.openfn.org/projects/18bd2cb5-4991-4da8-98dc-c33b3878c707/w/adb99096-63d1-458c-a56c-c9da804eb60d?run=8cdcb640-e189-4df6-b4b1-d8ac4974e116&job=8ccb7935-e797-4f32-8384-9e15a0b3f781&panel=editor&step=e3f3a6e9-f294-4e1f-8457-a27b6dc8be1e#log)

josephjclark:
Ok- so those lost runs are probably occuring because we're downloading too much data and kubernetes is killing the process. The worker should really be robust to this but we know there are problems in that area.

So `paging: false` can be used to download all data at once, but doing so might cause the run to exceed its memory limit. I'm OK with that - it's a natural limitation of the system.

Doing manual paging in a workflow is very very hard, but possible. 

Adding a better pagination API to dhis2 is possible of course. We closed down that PR on the grounds that a) it wasn't immediately needed and b) we can achieve the same thin through `paging:false` and good example (this only feels half true?).

There's also a c): I  worry that, the way the closed pagination API works, it is still in danger of downloading too much data into a single run and causing memory problems. What we really need there (and on other adaptors) is to start introducing paging APIs which allow users to process a single page of a data at a time - rather than just downloading everything in one go. That can _still_ cause problems but at least it gives us some options.

Anyway that's another story. Pagination is complicated.

So we all agree that this PR doesn't do much, but the extra bit of documentation it adds is probably worthwhile? But we need to do further and deeper work to paginate and safely process very large data sets (this problem is more than just pagination)

mtuchi:
I totally agree pagination is really hard 😮‍💨, The api we have are so easy to footgun yourself. We definitely need some deeper work on that area

But I still think these two points still need to be addressed

> 1. Allow user to specify `async` mode in `tracker.import`, before even if you specify in query it will always be set to `false`. Now user have the ability to set it to `true`. I think this is important we shouldn't limit user to always wait for results
> 2. This pr remove `async` option in `tracker.export` because it's not supported. You will always get `events, enrollments` etc. There is not `async: true or false` for `tracker.export`. This is more of a clean up

And this pr already address those two points. 

Documentation does help but for now in think the [job library pr](https://github.com/OpenFn/docs/pull/797) is the best place for it because we have more room to say how to handle pagination in dhis2.

PR #1682, opened 2026-05-29
add support for google service accounts

mtuchi:
## Summary

Added support for Service Accounts credential type in `gmail`, `googlesheets` and `googledrive` adaptor

**Related**
- https://github.com/OpenFn/docs/pull/777

## Details

Service Account credentials requires `private_key`, `client_email`, and `scopes`. I have modify `createConnection` function to check if `private_key` and `client_email` is specified in `state.configuration` to use `JWT` auth otherwise we default to use `OAuth2` auth.

Mandatory scopes are currently hardcoded for each adaptor.  I used the mandatory scope specified on app when creation OAuth credential for each each adaptor (gmail, googlesheets and googledrive).

## Credential

This PR introduces a set of three new credential keys. Users in the app must choose to either use a service account style credential, or an oauth style credential.

The only way to release this safely, without breaking anythingin the app, is to make all fields optional (non required) in the schema. This won't break older google credentials and won't break the Lightning UI.

Until we have better support for this sort of thing in Lightning, this is the safest way to release this feature (even if it's poor UX). 

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


mtuchi:
Hiya @hunterachieng can you give this pr, I was able to test, GoogleDrive and GoogleSheets adaptor but Gmail adaptor is failing with an error saying `"Error sending message: Precondition check failed.` I think the error has nothing to do with how we are handling service account credential rather on admin settings to allow the service account permission to read and compose email

mtuchi:
@hunterachieng see my walk through video
https://www.loom.com/share/0bcb55300e39480fb914551747abcc39

hunterachieng:
@mtuchi I see you mentioned that we have Gmail code, but then you have mentioned that it fails. Should we fix this first or?

mtuchi:
> @mtuchi I see you mentioned that we have Gmail code, but then you have mentioned that it fails. Should we fix this first or?

@hunterachieng it's not failing because of the changes, it's failing because of permission, can you spend sometime to figure out the permission we need ?

packages/gmail/configuration-schema.json
```diff
@@ -1,19 +1,54 @@
...
+      "description": "Gmail user to impersonate via domain-wide delegation (e.g. user@yourdomain.com). Required for service account access to Gmail.",
+      "examples": [
+        "user@yourdomain.com"
+      ]
     }
   },
-  "type": "object",
-  "additionalProperties": true,
-  "required": ["access_token"]
+  "oneOf": [
```
mtuchi: I am worried this will break on lightning 🤔

mtuchi:
Hey @hunterachieng when you get sometime please review this pr again for me, I have updated the README.md for updated adaptors

packages/gmail/configuration-schema.json
mtuchi: @josephjclark would love you're input on this, I think lightning is expecting `required` filed not "oneOf". 
Should we have have two different schema maybe 🤷🏽 ? But again lightning need to support both of those schema

josephjclark: I will do a test a bit later

josephjclark:
@mtuchi this needs rebasing now - hope the changes aren't too bad. I'll help you merge this quickly when the branchi is stable.

packages/gmail/configuration-schema.json
josephjclark: Ah dang, I should have tested earlier. Yes, this seems to break Lightning :(


Probably the only way to handle this is to remove all required properties.

In salesforce we have once schema for oauth and one for regular login. I don't actually know how that works technically - but could we do somethign similar here?

josephjclark (changes requested):
Blocked because the schema is currently incompatible with Lightning :(

packages/gmail/configuration-schema.json
mtuchi: @josephjclark  I have separated the service accounts configurations from the oauth2 schema. It should be good now

mtuchi:
Hiya @josephjclark i have rebase and separate the configuration schemas into separate files , `configuration-schema.json` and `service-configuration-schema.json`

mtuchi:
@josephjclark i have removed the `service-accounts-configuration.json` files and add it's content to configuration-schema.json for gmail, googledrive and googlesheet adaptor

josephjclark (approved):
Awesome stuff.

Just to note that googlehealthcare has a test but is otherwise unchanged. It has no changeset, which is correct

I'll release this now

PR #1683, opened 2026-05-29
add image util functions

mtuchi:
## Summary

_WIP_ Image processing util functions 

- https://github.com/OpenFn/plan-international/issues/1

## Details

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -47,9 +47,11 @@
     "csv-parse": "^5.4.0",
     "csvtojson": "^2.0.13",
     "date-fns": "^2.25.0",
+    "exiftool-vendored": "^35.20.0",
     "http-status-codes": "^2.3.0",
     "jsonpath-plus": "^10.3.0",
     "lodash": "^4.18.1",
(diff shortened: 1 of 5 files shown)


mtuchi:
Going the micro-service approach, See joe's comment 
> * We can't use node js. We can do _almost_ everything - but we're blocked on EXIF metadata writing, which requires a special program in the environment.
> * So we'll need a microservice with a single endpoint to process images
> * Microservice must be well tested, and provide an easy route to let Plan validate it
> * Need to figure out a simple auth strategy for the service. Probably token based. How will we rotate tokens?
> * Probably should be using imagemagick to do the processing (at least most of it).  Recommend wrapping in bunjs
> 
> Likely gotcha in the runtime: we don't really want to write images to state at the end of each step (don't want to serialize the image buffer to string, and don't want to include the images in dataclips send back to lightning). This may require core engineering to work (small-medium effort, can probably do something very minimal - like don't serailize or emit top-level streams on state)

PR #1685, opened 2026-06-02
feat: add base in options for fhir schema generator

### LINKED ISSUE #1684: fhir: write the base to package.json metadata
We've hit some confusion when rebuilding fhir-eswatini about whether it should be based on fhir 4 or not.

When generating and adaptor you can pass `--base fhir-4` to have your new adaptor import helpers from the already published fhir-4 (it's an open question whether we want to continue to do this). 

But one you've done this, you need to continue to keep passing the flag. Otherwise the build will keep flip flopping and production weird diffs. So `pnpm build:src` should just know whether to use fhir-4 or not.

I think the best way to do that is to add metadata to package.json, which the build process then reads. Our fhir adaptors already have a `fhir` key on the package.json which contains some build metadata. Which includes the URL the spec was downloaded from and an `options` object (which looks empty right now).

So we want to:
1) If provided on build write the `base` option to `package.json:fhir`
2) By default, use the `base` option on package.json

Oh, I suppose we could also add the option to `mappings.js`. But options in the fhir key is probably better.

hunterachieng:
## Summary

When generating a FHIR adaptor with `--base fhir-4`, all profile files import shared types from `@openfn/language-fhir-4`. However, if subsequent runs (e.g. `build:spec`) omit the flag, the generator defaults to local imports — causing large noisy diffs every time the spec is rebuilt.

Changes:

`generate.ts`: Add `base` to the list of options persisted to `package.json:fhir.options`
`generate.ts`: Move `base` destructuring to after `pkg.fhir.options`, so saved values are picked up correctly on flag-less runs

With these changes, you do not need to pass `--base fhir-4` in `pnpm build:spec` as it will be automatically added

Fixes #1684 

## Details

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/fhir-eswatini/package.json
+++ b/packages/fhir-eswatini/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "build": "pnpm clean && build-adaptor fhir-eswatini src ast docs",
     "build:src": "pnpm generate-fhir fhir-eswatini --base fhir-4",
-    "build:spec": "pnpm generate-fhir fhir-eswatini",
+    "build:spec": "pnpm generate-fhir fhir-eswatini --respec",
     "test": "TS_NODE_TRANSPILE_ONLY=true mocha --experimental-specifier-resolution=node --no-warnings  --loader=ts-node/esm \"**/*.test.ts\"",
     "clean": "rimraf dist docs",
     "pack": "pnpm pack --pack-destination ../../dist"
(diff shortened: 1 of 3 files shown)


josephjclark (approved):
Fantastic.

If you re-generate with `base` now, are there are changes that need to be added? Or just the metadata in package.json?

josephjclark (approved):

PR #1686, opened 2026-06-03
image processing adaptor

mtuchi:
## Summary
New `@openfn/language-image-utils` adaptor providing image processing utilities for OpenFn job code.

## Details

[Using joe's implementation](https://github.com/josephjclark/img-test/blob/main/index.ts) i was able to do the following

Five exported operations, each accepting a `base64 string`, `data URL`, or `Buffer` as input:
    - `resize(input, { width, height })` — resizes to target dimensions using `jimp; outputs { buffer, width, height }`
    - `compress(input, { maxBytes, minQuality })` — quality-loop compression to a byte target; outputs `{ buffer, size, quality }`
    - `stripMetadata(input)` — strips all EXIF metadata (including GPS) by re-encoding through Jimp; outputs `{ buffer }`
    - `embedMetadata(input, { comment })` — embeds a string in the JPEG EXIF UserComment field; outputs `{ buffer }`
    - `metadata(input)` — reads image dimensions and orientation without modifying the image; outputs `{ width, height, orientation, size }`
  - All image-returning functions (resize, compress, strip, annotate) accept `{ parseAs: 'base64' }` to return a `base64 string `instead of a `Buffer`, similar to `parseAs` convention in  `language-common`
  - Image processing functions use `jimp` package. The `piexifjs` package is used only for `EXIF UserComment` injection in `annotate` function

Here is sample job code
```js
const url = 'https://tinyurl.com/rucpr6tk';

http.get(url, { parseAs: 'base64', maxRedirections: 1 });
resize($.data, { width: 1200, height: 1600 });
embedMetadata($.data.buffer, { UserComment: 'patient-id=42' });
compress($.data.buffer, { maxBytes: 700 * 1024 });
```

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


mtuchi:
https://github.com/user-attachments/assets/f38db509-685c-483c-a7f2-0f337c0e6c7b

packages/image-fn/src/Adaptor.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {object} [options={}]
+ * @param {number} [options.width=1200] - Output width in pixels
+ * @param {number} [options.height=1600] - Output height in pixels
+ * @param {number} [options.maxBytes=716800] - Maximum output file size in bytes
+ * @param {number} [options.minQuality=20] - JPEG quality floor (1–100); compression stops here even if maxBytes is not met
+ * @param {string} [options.comment] - String to embed in the EXIF UserComment field
+ * @state {ImageState}
+ * @returns {Operation}
+ */
+export function process(base64ImgStr, options = {}) {
```
josephjclark: So if this is a generic image processing adaptor, we can't have a single `process` function like this. Because process only works for a single use-case.

As a generic adaptor you have to export functions like `resize()` and `compress()` etc, then your job code will use them in the right way.

We could structure this  as a specific adaptor for Childcare (which is nice because we get the logo in the diagram), in which case you'd have to rename the adaptor. And add helpers of course for other Childcare APIs

mtuchi:
Hiya @josephjclark i have added `processJsquash` which uses `@jsquash` library. 

## `jimp` vs `@jsquash/jpeg`

| | `jimp` | `@jsquash` |
|---|---|---|
| Output size (q80) | ~389KB | ~312KB (–20%) |
| Input formats | JPEG, PNG, BMP, GIF, TIFF | **JPEG only** |
| Setup cost | None | ~200ms WASM init |
| Compression loop | 4170ms | 2512ms (40% faster) |

**For this use case (single CommCare JPEG per job):** differences are minor. `@jsquash` wins on compression efficiency; `jimp` wins on format flexibility.

> ⚠️ If we ever need to accept PNGs or screenshots, `@jsquash` throws at runtime

I think the right choice depends on use case. But i will pick `jimp` over `@jsquash` because of multiple image support and simple implementation. `@jsquash` wasmBinary feels a bit complicated to tbh

josephjclark:
@mtuchi `jimp` is fine by me - let's optimise when we need to. I'm a little anxious about the wasm in the work (although it should be fine)

Please see my earlier comment about the design of the adaptor. Need to get that right before I can look closer

packages/image-fn/src/Adaptor.js
mtuchi: I think generic adaptor is easier and can prove our concepts, later we can do a specific adaptor because the child data api is not even built yet

mtuchi: if it's a generic adaptor what should we name it ? @josephjclark

mtuchi:
Lets call it image-utils

packages/image-fn/src/Adaptor.js
mtuchi: I have rename the adaptor to `image-utils` and add more functions to cover the transformation requirements for child data

mtuchi:
Hiya @josephjclark i have addressed you're early feedback and cleanup .avi and .gif images, i will wait for you're feedback before cleaning up more images

packages/image-utils/test/Adaptor.test.js
```diff
@@ -0,0 +1,490 @@
...
+}
+
+// ─── resize ──────────────────────────────────────────────────────────────────
+
+describe('resize', () => {
+  describe('output dimensions', () => {
+    it('downscales a large portrait to target dimensions', async () => {
+      const finalState = await resize(toBase64Str('portrait-large.jpg'), DEFAULT_RESIZE_OPTS)(state);
+
+      const out = await Jimp.read(finalState.data.buffer);
```
josephjclark, line 55: would be nice to assert on the original size here, to make the test really clear

packages/image-utils/test/Adaptor.test.js
```diff
@@ -0,0 +1,490 @@
...
+
+import { resize, compress, strip, annotate, metadata } from '../src/Adaptor.js';
+import { resizeImage } from '../src/Utils.js';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+const state = { configuration: {}, data: {} };
+
+// Fixtures — see test/images/ for full details
+// landscape-large.jpg    4000×2667  4.1MB  landscape
```
josephjclark: Rather than have this big comment, it would be nice just to name the images like `img-4000x2667.jpg` or `img-4000x2667-gps.jpg`

Then I really think we can reduce the number of image significantly

packages/image-utils/test/Adaptor.test.js
```diff
@@ -0,0 +1,490 @@
...
+
+    it('downscales a landscape image to target portrait dimensions', async () => {
+      const finalState = await resize(toBase64Str('landscape-large.jpg'), DEFAULT_RESIZE_OPTS)(state);
+
+      const out = await Jimp.read(finalState.data.buffer);
+      expect(out.width).to.equal(1200);
+      expect(out.height).to.equal(1600);
+    });
+
+    it('upscales a small portrait to target dimensions', async () => {
```
josephjclark: I don;t think we need upscale and downscale tests tbh? Which means we can remove some target images.

We basically trust `jimp`  to do the resize logic for us. The only test we need is that if I say "resize this image to X by Y`, the correct arguments are passed to jimp.

It might be worth testing what happens if the aspect ratio of the image is changed - just to validate that `jimp` doesn't throw.

Finally, don't pass `DEFAULT_RESIZE_OPTS` as a constant. It's weird. Just pass the size we want in  the test

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+ * @function
+ * @param {string|Buffer|Function} base64ImgOrBuffer - Base64 string, data URL, Buffer, or resolver fn
+ * @param {object} [options={}]
+ * @param {number} [options.width=1200] - Output width in pixels
+ * @param {number} [options.height=1600] - Output height in pixels
+ * @param {'buffer'|'base64'} [options.parseAs='buffer'] - Return format: `'buffer'` (default) or `'base64'`
+ * @state {ImageState}
+ * @returns {Operation}
+ */
+export function resize(base64ImgOrBuffer, options = {}) {
```
josephjclark, line 45: Why `resize(image, options)` and not `resize(image, width, height)`? What do we gain with the options object?

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Resize an image to the given dimensions.
+ * Writes `{ buffer, width, height }` to `state.data`.
+ * @example
+ * resize(state.data.photoBase64, { width: 1200, height: 1600 })
+ * @function
+ * @param {string|Buffer|Function} base64ImgOrBuffer - Base64 string, data URL, Buffer, or resolver fn
+ * @param {object} [options={}]
+ * @param {number} [options.width=1200] - Output width in pixels
```
josephjclark: Do not default the width or height values. We must accept both

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+ * Compress an image to fit within `maxBytes` using a quality loop.
+ * Optionally embeds an EXIF UserComment.
+ * Writes `{ buffer, size, quality }` to `state.data`.
+ * @example
+ * compress(state.data.buffer, { maxBytes: 700 * 1024, comment: 'patient-id=123' })
+ * @function
+ * @param {string|Buffer|Function} base64ImgOrBuffer - Base64 string, data URL, Buffer, or resolver fn
+ * @param {object} [options={}]
+ * @param {number} [options.maxBytes=716800] - Maximum output file size in bytes
+ * @param {number} [options.minQuality=20] - JPEG quality floor (1–100); compression stops here even if maxBytes is not met
```
josephjclark, line 72: Open question (I honestly don't know): does min quality make sense? Isn't it a relative value?

if I take a jpg and downscale it to 80% quality, I get a new image. If I take THAT image and downscale it by 80% quality again, I've got an even smaller image. That's 80% * 80%, not 80% - 20%

I think all you're doing is taking n% of the pixels out. So you can just downscale by 20% forever and it'll keep getting smaller. I'm not sure what a minimum quality means.

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+    );
+    return composeNextState(
+      state,
+      applyParseAs(result, resolvedOptions.parseAs),
+    );
+  };
+}
+
+/**
+ * Compress an image to fit within `maxBytes` using a quality loop.
```
josephjclark: Remove `using a quality loop` - it's not really meaningful. You could I suppose say "by reducing image quality until it reaches the criteria", but I don't really think we need to explain _how_ it works?

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+ * Optionally embeds an EXIF UserComment.
```
josephjclark: Why does it embed a comment? Is this a random comment?

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+ * @example
+ * strip(state.data.photoBase64)
+ * @function
+ * @param {string|Buffer|Function} base64ImgOrBuffer - Base64 string, data URL, Buffer, or resolver fn
+ * @param {object} [options={}]
+ * @param {'buffer'|'base64'} [options.parseAs='buffer'] - Return format: `'buffer'` (default) or `'base64'`
+ * @state {ImageState}
+ * @returns {Operation}
+ */
+export function strip(base64ImgOrBuffer, options = {}) {
```
josephjclark: I would call this `stripMetadata`

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+ * annotate(state.data.buffer, { comment: 'patient-id=42' })
+ * @function
+ * @param {string|Buffer|Function} base64ImgOrBuffer - Base64 string, data URL, Buffer, or resolver fn
+ * @param {object} options
+ * @param {string} options.comment - String to embed in the EXIF UserComment field
+ * @param {'buffer'|'base64'} [options.parseAs='buffer'] - Return format: `'buffer'` (default) or `'base64'`
+ * @state {ImageState}
+ * @returns {Operation}
+ */
+export function annotate(base64ImgOrBuffer, options = {}) {
```
josephjclark: Instead of annotate, shouldn't this be   `setUserComment` or somthing?

If a comment exists, is it related or added to?

packages/image-utils/src/Adaptor.js
```diff
@@ -0,0 +1,186 @@
...
+  return async state => {
+    const [resolvedImg, resolvedOptions] = expandReferences(state, base64ImgOrBuffer, options);
+    const result = annotateImage(resolveInput(resolvedImg), resolvedOptions.comment);
+    return composeNextState(state, applyParseAs(result, resolvedOptions.parseAs));
+  };
+}
+
+/**
+ * Read image metadata without modifying the image. Useful for validation
```
josephjclark: Remove the `useful for` stuff

packages/image-utils/src/Adaptor.js
mtuchi: Done

packages/image-utils/src/Adaptor.js
mtuchi: @josephjclark in options we have `parseAs` option which is by default set to `buffer`. 

But there are couple of other benefits like if we decide to add other options like,`fit` for managing fit image aspect ration, or  `quality` to control image quality

Should we consider 
`resize(image, width, height, options)`
OR 
`resize(image, dimensions ={width, height}, options= {parseAs})`

packages/image-utils/src/Adaptor.js
mtuchi: Thanks, i have updated the docs

packages/image-utils/src/Adaptor.js
mtuchi: Aaah, the initial requirements was for adding UserComment metadata, I think the function should support adding any type of metadata not only `UserComment`, i will fix that

packages/image-utils/src/Adaptor.js
mtuchi: **I got this from claude code**

---
What Jimp actually does:

```js
const image = await Jimp.read(inputBuffer); // decoded to raw pixels (lossless in memory)
  while (true) {
    outBuffer = await image.getBuffer('image/jpeg', { quality }); // fresh encode of same pixels
    ...
    quality -= 5;
  }
```

`Jimp.read` decodes the JPEG fully into uncompressed pixel data. Every iteration of the The loop always re-encodes the same decoded pixels, not the previous iteration's JPEG output. So quality 20 means
  "compress the original image at quality 20" — not "take the quality-25 JPEG and compress it again." Each step is independent, starting from the same source.

packages/image-utils/src/Adaptor.js
mtuchi: Done

packages/image-utils/src/Adaptor.js
mtuchi: I think `embedMetadata` is more precise to what the function is doing

packages/image-utils/test/Adaptor.test.js
mtuchi: Updated the image to use that naming convention

packages/image-utils/test/Adaptor.test.js
mtuchi: This is done

mtuchi:
Hiya @josephjclark i have added some integration test based on recent feedback from @jfeldman-openfn, [See justin comment here](https://github.com/OpenFn/plan-international/issues/2#issuecomment-4702774964). I would love you're feedback when you get back

josephjclark:
TODO: add a little image emoji on the the icon

josephjclark:
The icon is a bit hard to figure out - I had a plan but it doesn't sit nicely. It's not important.

packages/image-utils/configuration-schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
josephjclark, line 2: We might have a problem with configuration schema. I think modern Lightning will force us to have one.

I'll run a test locally a bit later. We might need a little patch

mtuchi: Noted

josephjclark: Ok I've tested this. Admittedly it makes no sense to open the Add Credential UI for this adaptor. But if you do:
- if there is NO configuration file, the app errors silently and the credential UI does not load
- If there is a configuration file but with empty `properties`, the app shows a UI which basically only lets you set the name.

So this adaptor is safe to release right now. I just needed to verify that.

Note that common is exempt from the UI - it simply says "this adaptor does not require credentials". Really this adaptor should be included in that exemption list (or we should have a better solution)

josephjclark (approved):

PR #1689, opened 2026-06-09
DHIS2: Schema should not require username/password when using a PAT

### LINKED ISSUE #1547: DHIS2: Schema should not require username/password when using a PAT
## Adaptor Name and Version

dhis2 (all versions)

## Summary

The configuration schema requires `username` and `password` even when authenticating with a PAT, forcing users to provide credentials that are silently ignored.

## Steps to reproduce

1. In Lightning, create a DHIS2 credential using a Personal Access Token.
2. You are forced to also fill in `username` and `password` because the schema marks them as required.
3. At runtime, `configureAuth` in `src/util.js` silently ignores the username/password and uses only the PAT.

Separately, via the CLI:

1. Provide a configuration with a `password` but no `username` (and no `pat`).
2. The adaptor constructs a broken Basic auth header and the request fails with an opaque `401` from DHIS2.

## Any more details

The schema currently has:

```json
"required": ["hostUrl", "password", "username"]
```

And `configureAuth` uses a priority chain — PAT wins over password:

```js
if ('pat' in auth) {
  Object.assign(headers, { Authorization: `ApiToken ${auth.pat}` });
} else if ('password' in auth) {
  Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password));
}
```

The `password` branch doesn't check that `username` is also present, so CLI users (where the schema isn't enforced) can end up with silent auth failures.

## Expected behaviour

1. **Schema** should express that credentials are one-of: either `pat`, or both `username` + `password` — not all three required. A `oneOf`/`anyOf` structure would work.
2. **Runtime**: when falling back to Basic auth (no PAT), `configureAuth` should validate that both `username` and `password` are present and non-empty, and throw a descriptive error if either is missing.

PiusKariuki:
## Summary

The configuration schema requires username and password even when authenticating with a PAT, forcing users to provide credentials that are silently ignored.

The schema currently has:
```
"required": ["hostUrl", "password", "username"]
```
And configureAuth uses a priority chain — PAT wins over password:
```
if ('pat' in auth) {
  Object.assign(headers, { Authorization: `ApiToken ${auth.pat}` });
} else if ('password' in auth) {
  Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password));
}
```
The password branch doesn't check that username is also present, so CLI users (where the schema isn't enforced) can end up with silent auth failures.

Fixes #1547 

## Details
The schema currently has:

```json
"required": ["hostUrl", "password", "username"]
```

And `configureAuth` uses a priority chain — PAT wins over password:

```js
if ('pat' in auth) {
  Object.assign(headers, { Authorization: `ApiToken ${auth.pat}` });
} else if ('password' in auth) {
  Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password));
}
```

The `password` branch doesn't check that `username` is also present, so CLI users (where the schema isn't enforced) can end up with silent auth failures.

## Expected behaviour

1. **Schema** should express that credentials are one-of: either `pat`, or both `username` + `password` — not all three required. A `oneOf`/`anyOf` structure would work.
2. **Runtime**: when falling back to Basic auth (no PAT), `configureAuth` should validate that both `username` and `password` are present and non-empty, and throw a descriptive error if either is missing.

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/dhis2/test/auth.test.js
+++ b/packages/dhis2/test/auth.test.js
@@ -14,29 +14,18 @@ const testServer = enableMockClient(hostUrl);
 
 
 describe('configureAuth', () => {
-  it('should set a PAT auth header', () => {
-    const headers = util.configureAuth({ pat: 'mytoken' });
+  it('should prefer PAT over username and password when both are provided', () => {
+    const headers = util.configureAuth({ pat: 'mytoken', username: 'admin', password: 'district' });
     expect(headers).to.eql({ Authorization: 'ApiToken mytoken' });
   });
 
-  it('should set a basic auth header', () => {
-    const headers = util.configureAuth({ username: 'admin', password: 'district' });
-    expect(headers).to.eql(makeBasicAuthHeader('admin', 'district'));
-  });
-
-  it('should throw if pat is empty', () => {
-    expect(() => util.configureAuth({ pat: '' })).to.throw(
-      'Invalid authorization credentials. Include a PAT or a username and password in state.configuration'
-    );
-  });
-
-  it('should throw if username is empty', () => {
+  it('should throw if password is provided without a username', () => {
     expect(() => util.configureAuth({ username: '', password: 'district' })).to.throw(
       'Invalid authorization credentials. Include a PAT or a username and password in state.configuration'
     );
   });
 
-  it('should throw if password is empty', () => {
+  it('should throw if username is provided without a password', () => {
     expect(() => util.configureAuth({ username: 'admin', password: '' })).to.throw(
       'Invalid authorization credentials. Include a PAT or a username and password in state.configuration'
     );
--- a/packages/dhis2/test/auth.test.js
+++ b/packages/dhis2/test/auth.test.js
@@ -1,0 +1,51 @@
+import chai from 'chai';
+import { execute, create, update, upsert, get } from '../src/Adaptor.js';
+import { dataValue } from '@openfn/language-common';
+import { enableMockClient } from '@openfn/language-common/util';
+import * as util from '../src/util.js';
+import {
+  makeBasicAuthHeader,
+} from '@openfn/language-common/util';
+
+const { expect } = chai;
+
+const hostUrl = 'https://auth.dhis2.org';
+const testServer = enableMockClient(hostUrl);
+
+
+describe('configureAuth', () => {
+  it('should set a PAT auth header', () => {
+    const headers = util.configureAuth({ pat: 'mytoken' });
+    expect(headers).to.eql({ Authorization: 'ApiToken mytoken' });
+  });
+
+  it('should set a basic auth header', () => {
+    const headers = util.configureAuth({ username: 'admin', password: 'district' });
+    expect(headers).to.eql(makeBasicAuthHeader('admin', 'district'));
+  });
+
+  it('should throw if pat is empty', () => {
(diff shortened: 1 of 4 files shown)


packages/dhis2/test/auth.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const headers = util.configureAuth({ pat: 'mytoken' });
+    expect(headers).to.eql({ Authorization: 'ApiToken mytoken' });
+  });
+
+  it('should set a basic auth header', () => {
+    const headers = util.configureAuth({ username: 'admin', password: 'district' });
+    expect(headers).to.eql(makeBasicAuthHeader('admin', 'district'));
+  });
+
+  it('should throw if pat is empty', () => {
```
mtuchi: I think on the unit test for `configureAuth` could be summarised into 
```
it('should throw if no valid credentials are provided', () => {
  expect(() => util.configureAuth({})).to.throw('Invalid authorization...');
});

it('should throw if username is provided without a password', () => {
  expect(() => util.configureAuth({ username: 'admin', password: '' })).to.throw('Invalid authorization...');
});
```

I think the description should be very specific and tidy @PiusKariuki

mtuchi (changes requested):
@PiusKariuki can you update the pr checklist notes and title, also since you updated the util we should have a changeset

PiusKariuki:
@mtuchi You would like less tests? I think that would reduce our coverage in testing what was written in the spec. **_[See link](https://github.com/OpenFn/adaptors/issues/1547)_** See below for things we would miss out on:

1. Not testing the scenario where an empty string is passed in place of a `PAT`.
2. Not testing the scenario where the `configureAuth` function would not check if a  ​`username` is present.


I have added a changeset file as requested and the PR checklist notes and title

packages/dhis2/test/auth.test.js
mtuchi: @PiusKariuki where i am getting at is for example this description is not accurate
"should throw if pat is empty" because it will only throw if pat is empty and username and password is  not provided

josephjclark: Yeah these tests are a little peculiar.

They look thorough at first glance but:

- in the test `should throw if pat is empty`, user name and password are all empty
- you're testing "empty" but not null or undefined. Why is "empty" so significant?

They're not very tight, accurate, controlled tests. And they're actually quite misleading: `should throw if pat is empty` is simply not a reflection of the spec. I agree with Mtuchi's approach

josephjclark: You might also test for example that  pat is preferred to username and password - that feels like an important thing  to a) document in a test and b) preserve in a regression test

mtuchi: - "should prioritize `pat` over `username/password` when both are present"
- "should set Authorization header with ApiToken when `pat` is provided"
- "should set Basic Auth header when `username and password` are provided (no pat)"

PiusKariuki:
@josephjclark @mtuchi Got it. Thank you both for clarifying those issues on the tests. Key changes I have made:
1. Dropped the misleading `should throw if pat is empty` test.
2. Updated descriptions to accurately reflect what is being tested.
3. Added the PAT priority test.

josephjclark (approved):

PR #1698, opened 2026-06-25
Salesforce: patch 6.x to fix connections dropping

josephjclark:
Retro fit #1697 onto  6.0


josephjclark:
released and tagged https://github.com/OpenFn/adaptors/tree/%40openfn/language-salesforce%406.0.4

PR #1699, opened 2026-06-25
Salesforce: patch 7.2.x to fix connections dropping

josephjclark:
Retro fit https://github.com/OpenFn/adaptors/pull/1697 onto 7.2


josephjclark:
This is released and tagged

PR #1700, opened 2026-06-25
Salesforce: patch 8.x to fix connections dropping

josephjclark:
Retro fit https://github.com/OpenFn/adaptors/pull/1697 onto 6.0


josephjclark:
tagged and released

PR #1711, opened 2026-07-06
New BDR API implementation #1

stevkky:
## Summary

This is a completely new implementation for the new Ghana Birth and Death (BDR) Registry API

## API Documentation below:
https://documenter.getpostman.com/view/26097188/2sB2qZE2N9

## Details
Changes to: 
Adaptor.js 
Adaptor.test.js
Utils.js
mock.js
Created integration.test.js, to send sample data to BDR beta endpoint
## AI Usage

know!):

- [ ] I have used Claude Code
- [ Yes] I have used another model =>  qwen3.6-27b
- [ ] I have not used AI

## Review Checklist


taylordowns2000 (changes requested):
Thanks @stevkky ! This is great. I'm assuming it's all tested and working properly when you're running it from your local OpenFn installation, so I'm just asking Claude Fable to focus on code/pattern rather than checking actual end to end stuff. I've also requested that we get an OpenFn adaptor engineer's eyes on this soon... I only expect this automated review to catch a few basic things and then the human review to do better 😄 

1. `test/integration.test.js:12` — I'm assuming this is not an actual API token, but would be good to confirm and maybe make it more obvious that this isn't valid.
2. `test/integration.test.js:9` — Live network tests run on every `pnpm test`. The package's test script is bare mocha with no exclude, so mocha's default `./test/*.js` glob picks up `integration.test.js`, which POSTs real birth records to the beta server on every unit-test/CI run. This violates the repo's testing rule in CLAUDE.md ("Every function needs a unit test using Undici MockAgent... No live network calls"). Follow the dhis2/kobotoolbox pattern: exclude it from the default glob and expose a separate test:integration script.
3. `src/Adaptor.js:20` — `execute()` installs the mock server in production whenever `configuration.baseUrl` is falsy, and the mock can never be evicted. Two failure modes: (a) with baseUrl missing, the run doesn't fail cleanly — getAccessToken throws an opaque ERR_INVALID_URL from new URL(path, undefined) (`src/Utils.js:46`); (b) because getClient only creates a real client when `!client` (`src/Utils.js:12`) and nothing ever resets the module-level singleton, a mock installed by one execution persists — a later execution in the same process with a real baseUrl is silently served fabricated success responses (fake document_number) instead of writing to BDR. Shipping mock.js inside production source and branching on config to enable it is the wrong altitude; keep mock wiring in the test suite (via setMockClient or language-common's enableMockClient pattern) and make a missing baseUrl throw a clear config error.
4. `src/Utils.js:163` — [BDR DEBUG] console.logs dump full request bodies containing PII. Every createBirthRecord call logs the entire payload — Ghana Card numbers, phone numbers, names, and addresses of newborns and parents — into run logs (also error bodies at lines 60 and 170). On app.openfn.org those logs are retained and visible to project collaborators. Remove the debug logging or redact the body.
5. `src/Utils.js:91` — The response/request encoding contract is unverified and internally contradictory. The old adaptor double-parsed responses and double-stringified request bodies because BDR sent/required string-encoded JSON; the new code parses and stringifies once, yet mock.js still comments "BDR data endpoints return double-encoded JSON (JSON string inside JSON string)" while its mock actually replies single-encoded. If the real API still double-encodes, JSON.parse(bodyText) yields a string, state.data.api_data is undefined in every job, and unit tests can't catch it because the mock encodes once. Please confirm the new API's actual encoding against the Postman docs and make the mock and comments match.
6. `src/Utils.js:70` — Unvalidated token response can produce Authorization: Bearer undefined with a NaN expiry. If the token endpoint's response lacks api_data (or expires_in), destructuring from data.api_data || {} silently yields undefined, expiresAt becomes NaN (so the cache check always fails and the token is refetched on every operation), and requests go out with Bearer undefined, failing with an opaque BDR_ERROR after the 401 retry refetches the same broken token. Validate access_token/expires_in and throw a clear "malformed token response" error.
7. `src/mock.js:236` — The mock's request wrapper drops the body, query, and origin, so tests can't catch payload regressions. It destructures data but Utils.request sends the payload as pre-stringified body; query isn't forwarded; and origin: url is hardcoded (which is why the get test passes despite using baseUrl http://tracker.chimgh.org). Combined with validateRequestBody now being a stub that always returns true and the deleted 417 error branch, no test can assert on what the adaptor actually sends or how it handles non-200 responses.
8. `src/Adaptor.js:76` — The createBirthRecord JSDoc @example is syntactically invalid. Most keys have a stray trailing quote (`informant_national_id_number": "34454344",`), and this broken snippet is already baked into ast.json, which feeds docs.openfn.org — users who copy the published example into a job get a syntax error. Same ~60-field payload also exists in four drifting copies (JSDoc, mock.js, both test files); worth extracting one fixture and trimming the example.
9. `throwError('BDR_ERROR', { description: response.statusMessage, ... })` — undici responses have no statusMessage, so users always see a bare BDR_ERROR with no detail; and the 401-retry only works because throwError happens to let the numeric code extra clobber the string code — worth a comment or an explicit error shape.
10. The hand-rolled undici stack (client management, URL joining, error handling) duplicates @openfn/language-common/util's request helper used by sibling adaptors; only the BDR-specific encoding genuinely needs custom code.
11. The unused refreshToken field, /auth/refresh mock, and validateRequestBody stub are dead scaffolding; and there's no changeset in the PR (the checklist item is unticked — one is needed for a release).

taylordowns2000 (approved):
@josephjclark , @mtuchi , this is a single-use adaptor developed by our colleagues in Ghana. It's passed our standard guidelines and their own testing, so as they're the only user (it's for a government API) I'm going to approve and merge this.

We may want to suggest follow-ups later. (Or even move these Ghana-specific gov't adaptors into a separate repo?)

PR #1714, opened 2026-07-07
Gmail: parse attachments (including xlsx)

### LINKED ISSUE #1712: gmail: saving an xls results in a lost run
Saving an XLS attachment in the gmail adaptor will result in a lost run. 

On the Lightning side it results in https://github.com/OpenFn/lightning/issues/4893

This is either because of a wierd character in the XLS file; or _much more likely_ it's because the XLS is being saved as a binary. And the binary is rejected by the database with that slightly cryptic error.

The most likely solution here is to save the XLS as a base 64 encoded string, where it can be processed downstream. 

This issue is to fix the gmail end of the problem. But we may want to investigate processing binaries and files more generally.

josephjclark:
## Summary

An important update to gmail which changes how file attachments are processed

* Spreadsheets are automatically converted to JSON
* JSON files are automatically parsed
* Plaintext and XML files are returned as stings
* All other files (which are typically binaries) are returned as base64 stings

This is a targeted, minimal fix designed around a single specific use-case (XLSX)

Fixes #1712

## Details

When parsing attachments from an email, the adaptor will automatically detect XLSX formats and convert them to JSON.

The conversion is delibrerately simple, just an array of arrays. This is a serializable and portable format and a huge improvement

I am not going to expose `xlsx` directly or expose too many options on the conversion. I just want the adaptor to fetch readable data - that's it. Maybe we'll create a sheetJS adaptor (see #1718)

If auto-parsing doesn't work, users can set parseAs to force the attachment parsing.

This is a major bump because I worry about behavioural changes while processing attachments. I don't think production workflows will be affected.

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Let's check that files by default are encoded into base64 strings so that they are safe on state. I think that there's code to try and handle this but it can't be working or else we woulnd't have had this problem in the first place

josephjclark:
hmm. So there are 33 jobs on the cloud using gmail which use `getContentsFromMessages`. A lot of those will break on this latest 3.0 version.

There's only one of those using `@latest` - and that's @mtuchi's

I think this is safe to release as a major version. I will add a migration guide.

PR #1725, opened 2026-07-19
Add changeset check workflow and docs

### LINKED ISSUE #1688: Add a github automation which fails PRs without a changeset
If a contributor opens a PR  without a changeset on the change, the build should fail.

The fail should have a clear error message about how to fix the fail. Ideally it would link to some docs in the repo. 

This action can be disabled, perhaps if  the phrase "no changeset required" is included in the PR body. Or perhaps if a particular label is added to the PR? Something easy and intuitive. 95% or PRs must have a changeset though.

aloundoye:
## Summary

Add a GitHub Actions workflow that enforces a changeset on PRs (.github/workflows/changeset.yaml). The check fails PRs that do not include a .changeset/*.md file unless a maintainer applied the `no changeset required` label (exemptions are recorded in the step summary). Update the PR template and README to document the label-based exemption and provide guidance on running `pnpm changeset` when required.

Fixes #1688 

## Details

## AI Usage

know!):

- [ ] I have used Claude Code
- [X] I have used another model
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -29,8 +29,9 @@ Before merging, the reviewer should check the following items:
 - [ ] If this PR includes breaking changes, do we need to update any jobs in
       production? Is it safe to release?
 - [ ] Are there any unit tests?
-- [ ] Is there a changeset associated with this PR? If this PR does not affect a
-      release, has a maintainer applied the `no changeset required` label?
+- [ ] Does every changed adaptor have a corresponding changeset? If an adaptor
+      change does not affect a release, has a maintainer applied the
+      `no changeset required` label? Repository-only changes need neither.
 - [ ] If there is a changeset, was `pnpm run version` used to bump versions (not
       `pnpm changeset version` directly)? This ensures changelog dates are stamped correctly.
 - [ ] Have you ticked a box under AI Usage?
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -29,8 +29,8 @@ Before merging, the reviewer should check the following items:
 - [ ] If this PR includes breaking changes, do we need to update any jobs in
       production? Is it safe to release?
 - [ ] Are there any unit tests?
-- [ ] Is there a changeset associated with this PR? Should there be? Note that
-      dev only changes don't need a changeset.
+- [ ] Is there a changeset associated with this PR? If this PR does not affect a
+      release, has a maintainer applied the `no changeset required` label?
 - [ ] If there is a changeset, was `pnpm run version` used to bump versions (not
       `pnpm changeset version` directly)? This ensures changelog dates are stamped correctly.
 - [ ] Have you ticked a box under AI Usage?
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -1,0 +1,75 @@
+name: Changeset
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+
+permissions:
+  contents: read
+
+jobs:
+  changeset:
+    name: Changeset required
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Record changeset exemption
+        if: contains(github.event.pull_request.labels.*.name, 'no changeset required')
+        run: |
+          echo "Changeset check skipped by the 'no changeset required' label."
+          {
+            echo "## Changeset check skipped"
+            echo
+            echo "A maintainer applied the \`no changeset required\` label."
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Checkout repository
+        if: "${{ !contains(github.event.pull_request.labels.*.name, 'no changeset required') }}"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for a changeset
+        if: "${{ !contains(github.event.pull_request.labels.*.name, 'no changeset required') }}"
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          changeset_files=()
+
+          while IFS= read -r file; do
+            if [[ "$(basename "$file")" != "README.md" ]]; then
+              changeset_files+=("$file")
+            fi
+          done < <(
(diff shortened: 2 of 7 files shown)


josephjclark:
Hi @aloundoye, thank you for this.

I have some suggestions but I want to be sure this works. I have a similar PR open in the [kit](https://github.com/OpenFn/kit/pull/1479) repo which acutally is failing to run in CI. I haven't had time to check it.

I've approved workflows to run so let's see.

But you can help me test this by opening two new PRs which branch off of this: One should have a change and a changeset, and pass in CI; the other should have a change but not a changeset, and should fail in CI.

aloundoye:
Hi @josephjclark , thanks for the feedback.

The current PR can serve as the test case without a changeset. The Changeset required check is failing as expected because this PR does not include a .changeset/*.md file and does not currently have the exemption label.
I’ll create another PR branching from 1688 that includes both a small change and a changeset, which should verify the passing case.
Once both cases are confirmed, please apply the no changeset required label to this PR, since it only changes repository tooling and documentation. Applying the label should rerun the workflow and allow this PR to pass through the exemption path.
I’ll share the second test PR when it is ready.

aloundoye:
Hi @josephjclark , I created the passing test PR: https://github.com/aloundoye/adaptors/pull/1.
I opened the test PR in my fork, targeting the 1688 branch. I’m not sure whether this is the setup you intended, so please let me know if you would prefer it elsewhere.
It contains a repository change and a changeset, and the Changeset required check passed successfully.
PR #1725 is exercising the opposite case and currently fails because it has no changeset or exemption label. Once you apply the no changeset required label to #1725, it should rerun and pass through the exemption path.

josephjclark:
Hi @aloundoye 

One immediate improvement I see is that if no adaptor was touched in a PR, then the `changeset required` should not fail. This PR, for example, probably shouldn't fail.

I wonder if the action needs to be like:
```
Check if changeset is required
Check that changeset was provided
```

The first action always runs and always succeeds. Depending on its outout, the second action might be skipped, or failed if no changeset was included, or passes if changesets were found for each adaptor.

The second big thing I'd ask is that the logic for this stuff doesn't live in inside the action, but is contained in a separate script file. Bash is fine, but node might be more convenient for us to maintain. Aside from tidiness, the benefit of that is that those scripts could actually be run locally - making testing easier. Maybe we could even add it to the local `pnpm test` command so that contributors can check this stuff before pushing (and asking for permissions for CI to run)

josephjclark:
Another improvement would be to run this test on a per package basis. If a user updates common and gmail, but only submits a changeset for `gmail`, they'll find themselves in trouble. I can see from https://github.com/aloundoye/adaptors/pull/1/changes that the changeset test is a totally blanket change: if _anything_ in the repo changes, ensure there is _a single changeset file_.

The test we really want is: for each _adaptor_ that is changed, make sure there's a corresponding changeset targeted at it. Which is a harder test!

aloundoye:
Hi @josephjclark, I’ve updated the implementation based on your feedback.

The workflow now detects which adaptor packages changed and verifies that every one is explicitly targeted by an added or modified changeset. PRs that do not touch packages/*, including #1725, no longer require a changeset.
I also moved the logic into a Node script with a local `pnpm changeset:check` command and added automated tests for single- and multi-adaptor coverage, missing targets, malformed changesets, renames, and package additions/deletions.
Passing test: https://github.com/aloundoye/adaptors/pull/1
Missing-changeset test: https://github.com/aloundoye/adaptors/pull/2
I also verified that applying no changeset required makes the failing test pass, and removing it makes the check fail again.

josephjclark:
Looking pretty good here @aloundoye, the GH actions seem to be doing the right thing.

I just want  to experiement with one more thing.

The  script in `check-changesets` is very complex. It's cool that we can generate all this stuff with AI now, but we've got to be able to maintain and I've got to be able to trust it.

I've just been looking at the changeset docs and they have a `status` command. So we can do this:

```
$ pnpm changeset status
```

And it'll give a nice readout (it'll even serialize to JSON if we add `--out output.json`):
```
🦋  info NO packages to be bumped at patch
🦋  ---
🦋  info NO packages to be bumped at minor
🦋  ---
🦋  info Packages to be bumped at major:
🦋  info 
🦋  - @openfn/language-arcgis
```

And it'll throw if it detects changes to a package, but does not find a changeset.

[here's the docs](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#status

Do you think we can utilise this for a simpler implementation?

I've just done a quick test and it doen';t understand the monorepo structure. So if I change two packages and add one changeset, it's happy. That might not be robust enough for our needs - I don't know if we can do more?

aloundoye:
Thanks @josephjclark. I agree that the current script is more complex than this check should be, and simplifying it would make it easier to review and maintain.
I tested changeset status with two changed adaptors—Common and Gmail—but a changeset targeting only Gmail. Even with --since=<base SHA>, the command exited successfully and only reported Gmail.
Looking at the Changesets implementation, status fails when packages changed and there are zero changesets. It does not fail when a changeset exists but only covers some of the changed packages. So it cannot enforce our per-adaptor requirement by itself.

aloundoye:
I think we can still simplify the implementation considerably. The minimum custom logic is:
- Determine the changed adaptor package names.
- Read the package names explicitly targeted by the submitted changesets.
- Fail if changed packages − targeted packages is not empty.

We are already using the official @changesets/parse package, so we do not maintain the changeset/YAML parsing ourselves.
We could also run changeset status --since="$BASE_SHA" --output=status.json for additional validation, but I would not use it as the coverage check. It can also fail because of unrelated release-graph issues, as happened in the Kit PR.
My suggestion is to refactor the current script down to the smallest implementation of those three steps while retaining the per-adaptor tests. What do you think ?

josephjclark:
Hi @aloundoye - sorry for the late replies here, I've had a whirlwind week and I'm just about to take go away on leave. I'll get back to you soon and we'll get this merged. Thanks for your time and your patience 🙏

aloundoye:
Hi @josephjclark, Thanks for the update! No worries at all. Enjoy your leave, and we’ll pick this up when you’re back.

PR #1730, opened 2026-07-22
Msgraph: encode a zip

josephjclark:
This PR adds a `zip()` function to msgraph, allowing files to be encoded into a zip and uploaded to a sharepoint drive.

The solution uses `adm-zip`, described asa pure JS library. Note that gmail uses `jszip`, but that's not had an update in 4 years. So let's try adm-zip.

Extra fixes:
- Fix a bug building the URL on `create`

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/packages/msgraph/src/Adaptor.js
+++ b/packages/msgraph/src/Adaptor.js
@@ -387,17 +387,20 @@ export function uploadFile(resource, data, callback) {
  * ])
  * @function
  * @param {Object[]} files - An array of `{ name, content }` objects to add to the zip
- * @param {function} callback - Optional callback function
  * @state data the generated zip archive, as a buffer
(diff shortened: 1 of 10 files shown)


packages/msgraph/src/Adaptor.js
```diff
@@ -77,7 +81,7 @@ export function create(resource, data, callback) {
 
     const { accessToken, apiVersion } = state.configuration;
 
-    const url = setUrl({ apiVersion, resolvedResource });
+    const url = setUrl(resolvedResource, apiVersion);
```
mtuchi, line 84: Hey @josephjclark i have added this fix, `setUrl` was not set properly, and added a changeset

PR #1727, opened 2026-07-21
feat:update publish yaml

hunterachieng:
## Summary

Added two steps to the publish workflow between pnpm test and pnpm publish: one to run pnpm run changelog (which stamps today's date onto the latest version heading in each CHANGELOG.md), and one to verify only .md files changed before committing and pushing back to main with [skip ci]. The git identity config was also moved up to before the commit step, where it's actually needed.

Fixes #1692 

## Details

Changelog dates in CHANGELOG.md files were not being updated automatically as part of the publish pipeline — this was a manual step.

Fix: Added two steps to publish.yaml between pnpm test and pnpm publish:

Update changelog dates — runs pnpm run changelog, which stamps the current date onto the latest version heading in every packages/*/CHANGELOG.md

Commit changelog date updates — runs a safety check before committing:

Inspects git diff --name-only for any non-.md files
Fails loudly if unexpected files were modified
Commits and pushes with [skip ci] to avoid re-triggering the workflow
Skips silently if nothing changed
Also moved the git config user.name/email steps up to before the changelog commit (they were previously after publish, which would have caused the commit to fail).

Result: On every publish run, changelog dates are stamped, committed back to main, and the working tree is clean before pnpm publish runs.

## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to
know!):

- [ ] I have used Claude Code
- [x] I have used another model
- [ ] I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Review Checklist

Before merging, the reviewer should check the following items:

- [ ] Does the PR do what it claims to do?
- [ ] If this is a new adaptor, added the adaptor on marketing website ?
- [ ] If this PR includes breaking changes, do we need to update any jobs in
      production? Is it safe to release?
- [ ] Are there any unit tests?
- [ ] Is there a changeset associated with this PR? Should there be? Note that
      dev only changes don't need a changeset.
- [ ] If there is a changeset, was `pnpm run version` used to bump versions (not
      `pnpm changeset version` directly)? This ensures changelog dates are stamped correctly.
- [ ] Have you ticked a box under AI Usage?

PR #1732, opened 2026-07-23
feat: update commcare

hunterachieng:
## Summary

Update `commcare` apis to `v1` and `v2`.

Fixes #1727 

## Details

Add support for CommCare's [Case Data API v2](https://commcare-hq.readthedocs.io/api/cases-v2.html) and make the API version configurable.

**Breaking change:** The URL pattern used by `get()`, `post()`, and `fetchReportData()` has changed from `/api/v0.5/<resource>` to `/api/<resource>/<apiVersion>`. Existing jobs are unaffected as long as the CommCare v1 API is still available — the default `apiVersion` is `"v1"`.

**`get()` now correctly handles individual resource fetches**, placing the ID after the API version in the URL (e.g. `get("case/12345")` resolves to `/api/case/v1/12345`).

**New: configurable API version**

Set `apiVersion` in `state.configuration` to target a specific CommCare API version (defaults to `"v1"`):

```json
{
  "configuration": {
    "hostUrl": "https://www.commcarehq.org",
    "username": "...",
    "password": "...",
    "domain": "my-domain",
    "apiVersion": "v2"
  }
}
```

**New: Case Data API v2 (`apiVersion: "v2"`)**

The v2 case API introduces JSON-based case creation and updates (no XForm construction required), filtering by project-specific case properties, bulk operations, and a new response envelope. See the [v2 docs](https://commcare-hq.readthedocs.io/api/cases-v2.html) for full details.

Note: auto-pagination is v1-only (uses `meta.next`). The v2 API uses cursor-based pagination via a top-level `next` field — follow it manually using `request()` if needed.

## AI Usage

know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/hip-bats-admire.md
+++ b/.changeset/hip-bats-admire.md
@@ -1,0 +1,31 @@
+---
+'@openfn/language-commcare': major
+---
+
+Add support for CommCare's [Case Data API v2](https://commcare-hq.readthedocs.io/api/cases-v2.html) and make the API version configurable.
+
+**Breaking change:** The URL pattern used by `get()`, `post()`, and `fetchReportData()` has changed from `/api/v0.5/<resource>` to `/api/<resource>/<apiVersion>`. Existing jobs are unaffected as long as the CommCare v1 API is still available — the default `apiVersion` is `"v1"`.
+
+**`get()` now correctly handles individual resource fetches**, placing the ID after the API version in the URL (e.g. `get("case/12345")` resolves to `/api/case/v1/12345`).
+
+**New: configurable API version**
+
+Set `apiVersion` in `state.configuration` to target a specific CommCare API version (defaults to `"v1"`):
+
+```json
+{
+  "configuration": {
+    "hostUrl": "https://www.commcarehq.org",
+    "username": "...",
+    "password": "...",
+    "domain": "my-domain",
+    "apiVersion": "v2"
+  }
+}
+```
+
+**New: Case Data API v2 (`apiVersion: "v2"`)**
+
+The v2 case API introduces JSON-based case creation and updates (no XForm construction required), filtering by project-specific case properties, bulk operations, and a new response envelope. See the [v2 docs](https://commcare-hq.readthedocs.io/api/cases-v2.html) for full details.
+
+Note: auto-pagination is v1-only (uses `meta.next`). The v2 API uses cursor-based pagination via a top-level `next` field — follow it manually using `request()` if needed.
--- a/packages/commcare/ast.json
+++ b/packages/commcare/ast.json
@@ -102,7 +102,7 @@
           },
           {
             "title": "example",
-            "description": "post(\"case\", {\n  case_type: \"patient\",\n  case_name: \"Elizabeth Harmon\",\n  owner_id: \"20cc9dda-b90a-4af3-aa3d-fc67184e73ef\",\n  properties: { dob: \"1948-11-02\" }\n})\n// Note: set apiVersion: \"v2\" in state.configuration",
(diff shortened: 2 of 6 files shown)


josephjclark:
Thank you for opening this @hunterachieng. This has helped me understand what's going on.

I think the problem is bigger than I first feared :(

First of all there appear to be 112 jobs on prod which use `commcare@latest`. I don't want to break all those with a new major. We could perhaps manually migrate them in the database - but we also have to consider that there are other openfn deployments which may be affected.

We should avoid a breaking change unless absolutely neccessary. And I don't think this _neccessary_ yet: we were confused by reading old docs, but actually once we straighted that out the adaptor worked just fine.

Secondly, the content of this PR makes me nervous. We are being too clever in how we assemble urls to make `get(case/123)` work. The `get()` is a HTTP get request helper. It should really just do what it's asked, it shouldn't make assumptions. We are working too hard and making too many assumptions in how we rebuild the URL in this PR. I don't like it :(

It actually creates more confusion because the URLs in job code don't match the URLs in commcare's docs.

Thirdly, the `apiVersion` in config is now wrong!! 

`apiVersion` is designed to switch _all _ API calls used in the adaptor to use a particular version.

But what commcare have done is changed how they version APIs. Rather than versioning the whole API, they're versioning endpoints individually. That's why case data has API v1 and v2, but form data only has an API v1.

What it means is that we can't globally set API versions from config. The API version has to be set on each resource call.

In summary: the commcare adaptor is drastically out of data with the modern commcare API, and we can't really change it because the risk of breaking changes is too high.

And here's what I think we should do here (it's a lot of work)
- Ensure the new docs links point to the new commcare docsite at https://commcare-hq.readthedocs.io/
- Add a note to docs to explain that most functions use the 0.5 API, and to use the modern 1.0 API you have to use `request` or the `http` namespace
- Document that to use the 1.0 API, you have to use the `request` function or `http` namespace
- Introduce the http namespace as a minor update which only prefixes `/a/<domain/api` and not the version
- Deprecate the `apiVersion` configuration, and the `get`, `post` and `request` operations. But don't remove them. Document that these only work with the 0.5 API.
- Create a `http` namespace with the basic HTTP helpers
- The HTTP helpers should only build `/a/<domain/api` from config, so users will do `http.get('form/v1/<form-id>')`. The API version is explicitly in the URL
- I'm also tempted to add a rule that says: in a HTTP helper, if the path starts with `/`, we append it straight to the baseURL. That gives users total 100% control of the URL. I don't know if there's any actual usecase for this in commcare so I'd want to investigate a bit further.

Work for later:

- Consider creating `fetch` and `list` apis in the main namespace, which look a bit like `fetch('form', $.formId, { apiVersion, attachmentName }`. This is not a HTTP API, it's a "get a resource from commcare" API. It might support different API versions under the hood, but also maybe not.
- Offer generic pagination support to `http.get()`. Pagination must be smart and auto-select offset/page style or cursor style (ie for the v2 commcare APIs). This must be opt-in, not default, so require a `paginate: 'auto'` flag or something


Then in the next major version of commcare, we can just remove the deprecated API and clarify things.

I think all these changes can modernise the adaptor nicely without breaking anything.

It's Friday afternoon after a long week. Let's recap this next week, maybe get another brain in the loop, and decide what to do.

josephjclark (changes requested):
Blocking merge because a major update is just too risky

hunterachieng:
@josephjclark  I was anticipating some of the issues you have raised. I like the idea of revisiting this next week

josephjclark:
Closing in favour of https://github.com/OpenFn/adaptors/issues/1737

PR #1736, opened 2026-07-25
fix: pass parseAs into assertOK so error response bodies parse as JSON

### LINKED ISSUE #1604: fhir-eswatini: errors come back with escaped JSON
When the fhir server returns an error from http.post, the response body has string encoded JSON.

I've truncated a response but it looks a bit like this:
```
{
  "body": "{\"resourceType\":\"OperationOutcome\",\"issue\":[{\"ext
  "duration": 695,
  "headers": {
    "connection": "keep-alive",
    "content-length": "22088",
    "content-type": "application/fhir+json; charset=utf-8",
    "date": "Fri, 20 Mar 2026 16:06:11 GMT",
    "keep-alive": "timeout=5",
    "vary": "Accept-Encoding",
    "x-openhim-transactionid": "69bd707380863bf16a1b4ff8"
  },
  "message": "POST to http://172.209.216.154:5001/fhir returned 422: Unprocessable Entity",
  "method": "POST",
  "statusCode": 422,
  "statusMessage": "Unprocessable Entity",
  "url": "http://172.209.216.154:5001/fhir"
}
```

EmmS21:
## Summary

Pass `parseAs` into `assertOK` so HTTP error response bodies are parsed as JSON instead of left as strings (e.g. FHIR `application/fhir+json` OperationOutcomes).

Fixes #1604

## Details

`parseAs` controls how response bodies are parsed to JSON. It was already passed through on successful responses, but not into `assertOK` — the function that throws when a request gets an error status (≥400).

Without `parseAs` on that error branch, `readResponseBody` fell back to auto-detect, which only treats content-types containing `application/json` as JSON. Content-types like `application/fhir+json` were returned as text strings, so callers saw escaped JSON in `error.body` instead of a nested object.

### Testing

- Added unit test in `packages/common/test/util/http.test.js`:
  `should parse error body as json when parseAs is json`
- Covers 422 + `application/fhir+json` + `parseAs: 'json'`
- Asserts `error.body` is the OperationOutcome object, not a string

## AI Usage

- [ ] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

## Test plan

- [x] Unit test for error-body JSON parsing with `parseAs: 'json'` and `application/fhir+json`
- [x] Run `@openfn/language-common` unit tests (`pnpm test` in `packages/common`)
- [x] Confirm success-path parsing is unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/common/src/util/http.js
+++ b/packages/common/src/util/http.js
@@ -171,7 +171,20 @@ const assertOK = async (response, errorMap, fullUrl, method, startTime, parseAs)
       : errMapMessage || response.statusCode >= 400;
 
   if (isError) {
-    const body = await readResponseBody(response, parseAs);
+    // Honour parseAs for error bodies (e.g. application/fhir+json), but if
+    // JSON parsing fails, keep the raw text so the HTTP status error is thrown
+    // instead of a parse error (plain-text 404 pages, etc).
+    let body = await readResponseBody(
+      response,
+      parseAs === 'json' ? 'text' : parseAs
+    );
+    if (parseAs === 'json' && typeof body === 'string') {
+      try {
+        body = JSON.parse(body);
+      } catch {
+        // leave body as text
+      }
+    }
 
     const statusText = getReasonPhrase(response.statusCode);
     const defaultErrorMessage = `${method} to ${fullUrl} returned ${response.statusCode}: ${statusText}`;
--- a/packages/common/src/util/http.js
+++ b/packages/common/src/util/http.js
@@ -171,9 +171,6 @@ const assertOK = async (response, errorMap, fullUrl, method, startTime, parseAs)
       : errMapMessage || response.statusCode >= 400;
 
   if (isError) {
-    // Honour parseAs for error bodies (e.g. application/fhir+json), but if
-    // JSON parsing fails, keep the raw text so the HTTP status error is thrown
-    // instead of a parse error (plain-text 404 pages, etc).
     let body = await readResponseBody(
       response,
       parseAs === 'json' ? 'text' : parseAs
@@ -181,9 +178,7 @@ const assertOK = async (response, errorMap, fullUrl, method, startTime, parseAs)
     if (parseAs === 'json' && typeof body === 'string') {
       try {
         body = JSON.parse(body);
-      } catch {
-        // leave body as text
-      }
+      } catch {}
     }
 
     const statusText = getReasonPhrase(response.statusCode);
--- a/packages/common/src/util/http.js
+++ b/packages/common/src/util/http.js
@@ -158,7 +158,7 @@ export const enableMockClient = (baseUrl, options = {}) => {
   return dispatcher;
 };
 
-const assertOK = async (response, errorMap, fullUrl, method, startTime, parseAs) => {
+const assertOK = async (response, errorMap, fullUrl, method, startTime) => {
   if (errorMap === false) {
     return;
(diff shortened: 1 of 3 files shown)


hunterachieng (approved):

hunterachieng:
@EmmS21  do you mind fixing the breaking check?

josephjclark:
Hi @EmmS21 :wave: Thank you for the contribution. I agree with the fix and would love to get this merged.

There's a failing test in the `fhir` adaptor which you can see in the Build and Test CI report
```
packages/fhir test:   1) post
packages/fhir test:        throws an error for a 404:
packages/fhir test:      AssertionError: expected '404: Error parsing the response body' to include 'POST to [https://hapi.fhir.org/baseR4/…](https://hapi.fhir.org/baseR4/%E2%80%A6)'
packages/fhir test:       at Context.<anonymous> (file:///home/runner/work/adaptors/adaptors/packages/fhir/test/Adaptor.test.js:237:26)
packages/fhir test:       at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

Are you able to look into this? It looks at first glance like its related to the changes here. It might just be that the test needs updating, or it could reveal a subtle breaking change.

EmmS21:
@hunterachieng @josephjclark I fixed the failing test

josephjclark:
Thank you @EmmS21 ! We'll run a few manual tests and get this released soon. Sorry for the slow response, we've been super busy within the team lately

packages/common/src/util/http.js
```diff
@@ -171,7 +171,15 @@ const assertOK = async (response, errorMap, fullUrl, method, startTime) => {
       : errMapMessage || response.statusCode >= 400;
 
   if (isError) {
-    const body = await readResponseBody(response);
+    let body = await readResponseBody(
```
josephjclark: Hi, sorry for the late reply on this. I've been away and also busy on some priority stuff.

Whenever I open this PR to close it out, I get hung up on this bit 🤔 

I understand the princple of the PR: we want to support JSON object scoming back in the body of error responses from the server. I see the value in that, its cool.

But `readResponseBody` should be doing that anyway - so long as `parseAs` isn't set, it'll read the contentType header and use that to decide if it wants to return JSON (admittedly the test for this is probably native, but if that's the case maybe that's the real bug here).

What this fix seems to do is: if `parseAs` on the original request was set to JSON, regardless of the content type of the payload, force `readResponseBody` to return a text representation, and then manually parse that text into JSON.

It doesn't feel right to me and makes me nervous to merge. 

Am I misunderstanding something? Do you have a specific use-case which you're blocked on with this, or have you just picked it up from the ticket?

josephjclark: Sorry that's a lot of words - I think what I'm saying is: wouldn't a simpler fix be to foward `parseAs` straight into `readResponseBody`?

That'll work for cases where the error object returns the expected format. Like you ask for JSON and you get a JSON error. But might cause problems in cases where you get a HTML login page back, for example. Which makes me a bit nervous.

EmmS21: I see your point. I'm reusing `readResponseBody` now instead of parsing JSON by hand in `assertOK`.

Errors still go through auto, so we're not forcing error bodies into JSON.

josephjclark (approved):
Great, I'm a lot happier with this change. Thank you so much @EmmS21 !

PR #1734, opened 2026-07-24
Add fetchAttachments option to gmail adaptor

aloundoye:
Includes documentation updates and comprehensive tests verifying both disabled and enabled attachment fetching behavior.

Adds a new `fetchAttachments` option (default true) to `getContentsFromMessages` that allows skipping the download of file and archive attachments while still retrieving message metadata and body content. This improves performance when attachment content is not needed.
Add a high-level, single-sentence summary of what this PR changes.

Fixes #1731 

## Details

Add technical details of what you've changed (and why).

## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to
know!):

- [ ] I have used Claude Code
- [x] I have used another model
- [ ] I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Review Checklist

Before merging, the reviewer should check the following items:

- [ ] Does the PR do what it claims to do?
- [ ] If this is a new adaptor, added the adaptor on marketing website ?
- [ ] If this PR includes breaking changes, do we need to update any jobs in
      production? Is it safe to release?
- [ ] Are there any unit tests?
- [ ] Is there a changeset associated with this PR? Should there be? Note that
      dev only changes don't need a changeset.
- [ ] If there is a changeset, was `pnpm run version` used to bump versions (not
      `pnpm changeset version` directly)? This ensures changelog dates are stamped correctly.
- [ ] Have you ticked a box under AI Usage?

PR #1743, opened 2026-07-28
`gmail` Add getMessageById and `fetchAttachments` message option

### LINKED ISSUE #1731: Update `getContentsFromMessages` on gmail adaptor
## Background 

The current `getcontent` function automatically downloads attachements which might result in data bloat and "lost runs" when processing . The function needs to be updated to support an optional parameter (e.g., fetch_attachment: true/false). This would allow the first workflow to retrieve email metadata (like IDs) based on a cursor date without automatically downloading the heavy attachments.

## The Two Options Discussed
1. Modify the existing get content function: Add a parameter to the current function to toggle whether attachments are downloaded. 
2. Build another function that only gets the email list: this function will only retrieve email metadata and IDs without attempting to download attachments.

mtuchi:
## Summary

Continue on https://github.com/OpenFn/adaptors/pull/1734 contribution, This pr add `getMessageById` function

Fixes #1731 

## Details

- Added a `fetchAttachments` option (default `true`) to `getContentsFromMessages`, so jobs can skip downloading file/archive attachments while still retrieving message metadata and body content
- Added `getMessageById(messageId, options)` to fetch a single message directly by its Gmail API message id.
- Moved the shared "resolve which content to fetch" and "fetch + assemble one message" logic into `Utils.js` (`resolveContentPlan`, `buildMessageContent`), 

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/gmail-fetch-attachments.md
+++ b/.changeset/gmail-fetch-attachments.md
@@ -2,7 +2,5 @@
 '@openfn/language-gmail': minor
 ---
 
-Added a `fetchAttachments` option (default `true`) to `getContentsFromMessages`
-and `getMessageById` to control whether file and archive attachments are
-downloaded. When `false`, matched attachments are returned as `{ filename }` /
-`{ archiveFilename }` objects without content instead of being omitted.
+- Added a `fetchAttachments` option (default `true`) to
+  `getContentsFromMessages` and `getMessageById`.
--- a/.changeset/gmail-fetch-attachments.md
+++ b/.changeset/gmail-fetch-attachments.md
@@ -1,0 +1,8 @@
+---
+'@openfn/language-gmail': minor
+---
+
+Added a `fetchAttachments` option (default `true`) to `getContentsFromMessages`
+and `getMessageById` to control whether file and archive attachments are
+downloaded. When `false`, matched attachments are returned as `{ filename }` /
+`{ archiveFilename }` objects without content instead of being omitted.
--- a/.changeset/gmail-fetch-attachments.md
+++ b/.changeset/gmail-fetch-attachments.md
@@ -1,6 +1,0 @@
----
-'@openfn/language-gmail': minor
----
-
-- Added a `fetchAttachments` option (default `true`) to
-  `getContentsFromMessages` and `getMessageById`.
--- a/packages/gmail/src/Adaptor.js
+++ b/packages/gmail/src/Adaptor.js
@@ -216,9 +216,7 @@ export function sendMessage(message) {
 
 /**
  * Downloads contents from a single message of a Gmail account, identified by
- * its Gmail API message id, instead of searching with a query. Note that this
- * is the id returned as `messageId` / in `state.processedIds`, not the RFC 822
- * Message-ID header.
+ * its Gmail API message id.
  * @public
  * @function
  * @param {string} messageId - Gmail API message id to fetch.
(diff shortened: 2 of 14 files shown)


.changeset/gmail-fetch-attachments.md
```diff
@@ -0,0 +1,8 @@
+---
+'@openfn/language-gmail': minor
+---
+
+Added a `fetchAttachments` option (default `true`) to `getContentsFromMessages`
```
josephjclark: Way too many words here - just a super super brief senence

packages/gmail/src/Adaptor.js
```diff
@@ -202,6 +203,74 @@ export function sendMessage(message) {
...
+ */
+
+/**
+ * Downloads contents from a single message of a Gmail account, identified by
+ * its Gmail API message id.
+ * @public
+ * @function
+ * @param {string} messageId - Gmail API message id to fetch.
+ * @param {MessageIdOptions} [options] - Customized options including desired contents.
+ * @state {Object} data - The returned message object, of the form `{ messageId, contents } `
```
josephjclark: Why would this function return the message id back? Shouldn't state.data just be the contents?

packages/gmail/src/Adaptor.js
```diff
@@ -202,6 +203,74 @@ export function sendMessage(message) {
...
+ * @returns {Operation}
+ * @example <caption>Download attachments for a specific message identified by an earlier step</caption>
+ * getMessageById(
+ *   $.data.find(m => m.report).messageId,
```
josephjclark: eh I wouldn;t do the lookup here, it confuses the story of the example. Assume the user has already found the id

packages/gmail/src/Utils.js
```diff
@@ -103,6 +122,72 @@ export async function getMessageContent(message, desiredContent) {
...
+    requestedContents,
+  );
+
+  const needsFullFormat = contentIndicators.some(
+    ({ type }) => type === 'body' || type === 'file' || type === 'archive',
+  );
+  const messageFormat = needsFullFormat ? 'full' : 'metadata';
+
+  if (!fetchAttachments) {
+    console.log('fetchAttachments is false: skipping attachment downloads');
```
josephjclark: logging here feels a bit to verbose? We seem to have two logs which  say the same thing

packages/gmail/src/Adaptor.js
```diff
@@ -36,11 +35,13 @@ import {
  * @typedef {Object} Options
  * @public
  * @property {string?} [query] - Gmail search query string.
- * @property {Array<string|MessageContent>} [contents=['from', 'date', 'subject', 'body']]
+ * @property {Array<string|MessageContent>} [contents=['from', 'date', 'subject']]
  *   An array of strings or MessageContent objects used to specify which parts of the message to retrieve.
  * @property {Array<string>} [processedIds] - Ignore message ids which have already been processed.
  * @property {string?} [email] - The user account to retrieve messages from. Defaults to the authenticated user.
  * @property {int?} [maxResults] - Maximum number of messages to process per request. Default is 1000.
+ * @property {boolean} [fetchAttachments=true] - Whether to download file and archive attachments.
```
josephjclark, line 43: Shouldn't this just be a flag on `MessageContent`? So I might ass a `metaOnly` flag like this:
```
getContentsFromMessages(
  {
    query: 'after:15/01/2025',
    contents: [
      { type: 'file', name: 'metadata', file: 'report.txt', metaOnly: true }
    ]
  }
)
```

josephjclark (approved):

PR #1753, opened 2026-08-06
Feature/1738 commcare http namespace

### LINKED ISSUE #1738: commcare: Add new HTTP namespace
Add a new `htttp` namespace to commcare with `get`, `post` and `request` functions.

The `http` workspace should be similar to eg [openMRS](https://docs.openfn.org/adaptors/packages/openmrs-docs) .

All HTTP functions should use the new commcare URL structures, where urls are prefixed with `/a/<domain/>api` (but not a version number). So `http.get('cases/v1')` would call `/a/<domain>/api/cases/v1`.

Note that the user has to include versions in every request path in this API. This is correct and good.

But additionally I want to introduce a new convention: if a request starts with `/` then we treat it almost as an absolute URL and don't prefix anything. So `http.get('/cases/v1')` would literally just call `/cases/v1`.

This gives users full control of the URL (rather than having to guess what the result would be). It also exposes the FHIR endpoints, which are like `/a/<domain>/fhir`. So users can do  `http.get('/a/<domain>/fhir')` to access FHIR APIs directly

hunterachieng:
## Summary

Added `http` namespace with `http.get()`, `http.post()`, and `http.request()` for direct access to the CommCare REST API.

All functions support two URL conventions:

- **Relative paths** are automatically prefixed with `/a/<domain>/api/`, so you only need to provide the resource and version.
- **Absolute paths** (starting with `/`) are passed through unchanged, giving you full control over the URL.

```js
// Relative path — resolves to /a/my-project/api/case/v1
http.get('case/v1');

// Absolute path — used exactly as written
http.get('/a/my-project/api/case/v1');
```

- Used `/code-review` for initial code review

Fixes #1738 

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/slick-wombats-travel.md
+++ b/.changeset/slick-wombats-travel.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-commcare': minor
----
-
-Add `http` namespace with `http.get()`, `http.post()`, and `http.request()` for direct access to the CommCare REST API.
--- a/.changeset/slick-wombats-travel.md
+++ b/.changeset/slick-wombats-travel.md
@@ -1,0 +1,18 @@
+---
+'@openfn/language-commcare': minor
+---
+
+Add `http` namespace with `http.get()`, `http.post()`, and `http.request()` for direct access to the CommCare REST API.
+
+All functions support two URL conventions:
+
+- **Relative paths** are automatically prefixed with `/a/<domain>/api/`, so you only need to provide the resource and version.
+- **Absolute paths** (starting with `/`) are passed through unchanged, giving you full control over the URL.
+
+```js
+// Relative path — resolves to /a/my-project/api/case/v1
+http.get('case/v1');
+
+// Absolute path — used exactly as written
+http.get('/a/my-project/api/case/v1');
+```
--- a/.changeset/slick-wombats-travel.md
+++ b/.changeset/slick-wombats-travel.md
@@ -3,16 +3,3 @@
 ---
 
 Add `http` namespace with `http.get()`, `http.post()`, and `http.request()` for direct access to the CommCare REST API.
-
-All functions support two URL conventions:
-
-- **Relative paths** are automatically prefixed with `/a/<domain>/api/`, so you only need to provide the resource and version.
-- **Absolute paths** (starting with `/`) are passed through unchanged, giving you full control over the URL.
-
-```js
-// Relative path — resolves to /a/my-project/api/case/v1
-http.get('case/v1');
-
-// Absolute path — used exactly as written
-http.get('/a/my-project/api/case/v1');
-```
--- a/packages/commcare/src/Utils.js
+++ b/packages/commcare/src/Utils.js
@@ -5,7 +5,7 @@ import {
   logResponse,
 } from '@openfn/language-common/util';
 
-export const stripUrlPath = (url, domain) => {
+export const buildUrl = (url, domain) => {
   let finalUrl = '';
 
   const absoluteUrl = url.startsWith('/');
--- a/packages/commcare/src/Utils.js
+++ b/packages/commcare/src/Utils.js
@@ -5,6 +5,19 @@ import {
   logResponse,
 } from '@openfn/language-common/util';
 
+export const stripUrlPath = (url, domain) => {
+  let finalUrl = '';
+
+  const absoluteUrl = url.startsWith('/');
+  if (absoluteUrl) {
+    finalUrl = url;
+  } else {
+    finalUrl = `/a/${domain}/api/${url}`;
+  }
+
+  return finalUrl;
+};
+
 export const configureAuth = (auth, headers = {}) => {
   if ('apiKey' in auth) {
     Object.assign(headers, {
@@ -14,7 +27,7 @@ export const configureAuth = (auth, headers = {}) => {
     Object.assign(headers, makeBasicAuthHeader(auth.username, auth.password));
   } else {
     throw new Error(
-      'Invalid authorization credentials. Include an apiKey or password in state.configuration'
+      'Invalid authorization credentials. Include an apiKey or password in state.configuration',
     );
   }
 
--- a/packages/commcare/src/http.js
+++ b/packages/commcare/src/http.js
@@ -13,7 +13,7 @@ import * as util from './Utils.js';
 /**
  * Make a POST request to CommCare. Use this to send resources directly to Commcare REST API.
  * You can pass CommCare body data as a JSON object.
- * @example <caption>Create a case using a relative path (prefixed with /a/<domain>/api/)</caption>
+ * @example <caption>Post to the V2 case API with JSON data (prefixed with /a/<domain>/api/)</caption>
  * http.post('case/v2', {
  *   case_type: 'patient',
  *   case_name: 'Elizabeth Harmon',
@@ -34,7 +34,7 @@ export function post(path, data, params = {}) {
 
 /**
  * Make a GET request to CommCare. Use this to retrieve resources directly from the CommCare REST API.
- * @example <caption>Get cases using a relative path (prefixed with /a/<domain>/api/)</caption>
+ * @example <caption>Get cases using the v1 API (prefixed with /a/<domain>/api/)</caption>
  * http.get('case/v1');
  * @function
  * @public
@@ -49,7 +49,7 @@ export function get(path, params = {}) {
 
 /**
  * Make a general HTTP request against the CommCare server. Use this to make any request to CommCare REST API.
- * @example <caption>GET cases with a relative path (prefixed with /a/<domain>/api/)</caption>
+ * @example <caption>Get cases using the v1 API (prefixed with /a/<domain>/api/)</caption>
  * http.request('GET', 'case/v1');
  * @function
  * @public
--- a/packages/commcare/src/http.js
+++ b/packages/commcare/src/http.js
@@ -1,0 +1,139 @@
+import { expandReferences } from '@openfn/language-common/util';
+import * as util from './Utils.js';
+
+/**
+ * State object
+ * @typedef {Object} CommcareHttpState
+ * @property data - The response body (as JSON)
+ * @property response - The HTTP response from the CommCare server (excluding the body)
+ * @property references - An array of all previous data objects used in the Job
+ * @private
+ */
+
+/**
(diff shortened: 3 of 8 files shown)


packages/commcare/test/http.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+describe('stripUrlPath', () => {
+  it('prefixes a relative path with /a/<domain>/api/', () => {
+    expect(stripUrlPath('case/v1', 'my-domain')).to.equal(
+      '/a/my-domain/api/case/v1'
+    );
+  });
+
+  it('passes an absolute path through unchanged', () => {
+    expect(stripUrlPath('/a/my-domain/api/case/v1', 'my-domain')).to.equal(
```
mtuchi: @hunterachieng does this mean if i specify 
```
a/my-domain/api/case/v1
```
This won't be treated as absolute url ?

mtuchi:
CommCare's URL scheme changed, which matters a lot for this pr There are now two shapes:

- Legacy: `/a/[domain]/api/v0.5/[resource]/` — one global version number for all resources
- Current: `/a/[domain]/api/[resource]/v1/` — each resource versioned independently

[Dimagi's overview](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2279637003/CommCare+API+Overview#Documented-APIs-and-use-of-Standards) says the version placeholder should be replaced with one of v2, v1, v0.5, or v0.4, and that the docs only describe the latest version of each resource, with some prior versions kept only for backwards compatibility. In practice v1 is the successor to v0.5 for most resources, and only two resources have a v2: cases and locations

### Data APIs

Resource | Current path | Legacy v0.5 equivalent
-- | -- | --
Application structure | /a/[domain]/api/application/v1/[app_id] | /api/v0.5/application/
Form data / list forms | /a/[domain]/api/form/v1/[form_id]/ | /api/v0.5/form/
Cases (v1) | /a/[domain]/api/case/v1/ and /[case_id]/ | /api/v0.5/case/
Cases (v2) | /a/[domain]/api/case/v2/ | n/a
Groups | /a/[domain]/api/group/v1/ | /api/v0.5/group/
Mobile workers | /a/[domain]/api/user/v1/ | /api/v0.5/user/
Web users | /a/[domain]/api/web-user/v1/ | /api/v0.5/web-user/
Bulk user | /a/[domain]/api/bulk-user/v1/ | /api/v0.5/bulk-user/
List reports | /a/[domain]/api/simplereportconfiguration/v1/ | /api/v0.5/simplereportconfiguration/
Report data | /a/[domain]/api/configurablereportdata/v1/[report_id]/ | /api/v0.5/configurablereportdata/
Locations (v1) | /a/[domain]/api/location/v1/ | /api/v0.5/location/
Locations (v2) | /a/[domain]/api/location/v2/ | n/a
Location types | /a/[domain]/api/location_type/v1/ | /api/v0.5/location_type/
Fixtures | /a/[domain]/api/fixture/v1/ | /api/v0.5/fixture/
Lookup tables | /a/[domain]/api/lookup_table/v1/ | /api/v0.5/lookup_table/
Lookup table rows | /a/[domain]/api/lookup_table_item/v1/ | /api/v0.5/lookup_table_item/
DET exports | /a/[domain]/api/det_export_instance/v1/ | newer, no v0.5

mtuchi:
### User APIs

Purpose | Endpoint
-- | --
Create / edit / delete mobile worker | POST/PUT/DELETE /a/[domain]/api/user/v1/
User domain list | GET https://www.commcarehq.org/api/user_domains/v1/ (no domain in path)
User identity | GET https://www.commcarehq.org/api/identity/v1/ (no domain in path)
Web user invitation | POST /a/[domain]/api/invitation/v1/
Web user edit | PATCH /a/[domain]/api/web-user/v1/[id]/
Enable / disable user | POST /a/[domain]/api/web-user/v1/[id]/enable and .../disable
SSO credential check |  

### Not version-namespaced
These sit outside the `/api/[resource]/[version]/` convention entirely:

Purpose | Endpoint
-- | --
Form submission (OpenRosa) | POST /a/[domain]/receiver/api/ or /receiver/[app_id]/ or /receiver/submission/
OTA restore | GET /a/[domain]/phone/restore/
Bulk case upload (Excel) | POST /a/[domain]/importer/excel/bulk_upload_api/
Bulk lookup table upload | POST /a/[domain]/fixtures/fixapi/
Import application | POST /a/[domain]/apps/api/import_app/
App multimedia upload | POST /a/[domain]/apps/api/[app_id]/multimedia/
Multimedia status poll | GET /a/[domain]/apps/api/[app_id]/multimedia/status/[processing_id]/
Messaging events | GET /a/[domain]/api/messaging-event/v1/
FHIR | /a/[domain]/fhir/R4/Patient/[case-id] (superusers only)

mtuchi:
@josephjclark due to the url schema change in commcare, [See this comment](https://github.com/OpenFn/adaptors/pull/1753#issuecomment-5204969461), Should we instead have a util that will build the url to support both legacy and current url schema ?, Or the http.namespace should support the new schema and the adaptor functions support the legacy ?

packages/commcare/test/http.test.js
hunterachieng: Yes. @mtuchi  with the absolute url, you can send a request for whichever version you desire, but if it is relative you will only be able to send a request to endpoints with `/api` in the url

mtuchi: @hunterachieng but for the adaptor to recognised it's absolute url it must start with `/`?

josephjclark:
@mtuchi the whole point of this PR is to add a http namespace which is compatible with the current API format. We are not supporting the legacy format. 

I'm not sure what your list of URL is tring to tell me 🤔 The spec supports every URL in that list through the two conventions outlined by Hunter's PR description (including the legacy URLs actually)

We should not add a utility to build URLs. What would it do? `get(buildUrl( 'forms', 'v1'))` is a LOT harder to use than `get('forms/v1')`.

This is a minor update so the existing main namespace functions should not be affected by the change

mtuchi:
@josephjclark 
I am seeing a problem with `get("forms/v1")` vs `http.get("forms/v1")` clearly they are not the same, 
How do we make it simple for user to understand when to use? (docs things probably). I probably need fresh eyes on this 
I will sync with @hunterachieng tomorrow and share some feedback 

I think for `http.namespace` it should `builUrl => `[baseUrl]/a/[domain]/api/[resolvedPath]`
resolvedPath can be any `v0.5/[resource]` or `[resource]/v1/`

josephjclark:
@mtuchi but `get()` is pretty much broken. We're deprecating it. Ignore it - we're gonna remove it when we think it's safe

EDIT: get only works with the legacy 0.5 api. `get('forms/v1')` doesn't work. We've gotta modernise this adaptor

mtuchi:
Okay i had a chat with @hunterachieng all is clear now, Since we are deprecating `get()` in coming tickets I am happy with this pr. Because it will give users more flexibility in which endpoint to use,  It's bugging me why these are not the same.
```js
http.get("case/v1")
http.get("/case/v1") // this fails
```
The UX from other adaptor it doesn't matter if path start  with `/` or without.

mtuchi (approved):

.changeset/slick-wombats-travel.md
```diff
@@ -0,0 +1,18 @@
...
+
+- **Relative paths** are automatically prefixed with `/a/<domain>/api/`, so you only need to provide the resource and version.
+- **Absolute paths** (starting with `/`) are passed through unchanged, giving you full control over the URL.
+
+```js
+// Relative path — resolves to /a/my-project/api/case/v1
+http.get('case/v1');
+
+// Absolute path — used exactly as written
+http.get('/a/my-project/api/case/v1');
```
mtuchi: Morning @hunterachieng, since domain is part of the credential, can we change this example to 
```js
http.get('/api/case/v1')
```

packages/commcare/src/http.js
```diff
@@ -0,0 +1,95 @@
...
+ * You can pass Commcare body data as a JSON object.
+ * @example <caption>Create a case using a relative path (prefixed with /a/<domain>/api/)</caption>
+ * http.post('case/v2', {
+ *   case_type: 'patient',
+ *   case_name: 'Elizabeth Harmon',
+ *   owner_id: '20cc9dda-b90a-4af3-aa3d-fc67184e73ef',
+ *   properties: { dob: '1948-11-02' },
+ * });
+ * @example <caption>Create a case using an absolute path (used as-is, starting with /)</caption>
+ * http.post('/a/plan-global-hub-staging/api/case/v2', {
```
mtuchi: I think domain is part of the credential, this should be 
```
 http.post('/api/case/v2', {})
```

mtuchi (changes requested):
@hunterachieng for absolute path examples can we remove the domain part ?

packages/commcare/src/http.js
```diff
@@ -0,0 +1,79 @@
...
+ * State object
+ * @typedef {Object} CommcareHttpState
+ * @property data - The response body (as JSON)
+ * @property response - The HTTP response from the CommCare server (excluding the body)
+ * @property references - An array of all previous data objects used in the Job
+ * @private
+ */
+
+/**
+ * Make a POST request to CommCare. Use this to send resources directly to Commcare REST API.
```
josephjclark, line 14: Can we please capitalise Commcare consistently here? I'm not sure the correct form. Check their docs and make sure we use the same

packages/commcare/src/http.js
```diff
@@ -0,0 +1,79 @@
...
+ * You can pass Commcare body data as a JSON object.
+ * @example <caption>Create a case using a relative path (prefixed with /a/<domain>/api/)</caption>
```
josephjclark: Don't talk about relative paths in the example. And I wouldn't describe the prefix here either tbh.

And the example isn't really "create a case", the example is "post to the V2 case API with JSON data"

packages/commcare/src/Utils.js
```diff
@@ -5,6 +5,19 @@ import {
   logResponse,
 } from '@openfn/language-common/util';
 
+export const stripUrlPath = (url, domain) => {
```
josephjclark: Hmm, the name of this function isn't very good. It doesn't strip anything - it just maybe appends the domain.

Could just call it `buildUrl`?

mtuchi (approved):

PR #1756, opened 2026-08-10
Added getResource & list helpers

### LINKED ISSUE #1742: commcare: add fetch and list helpers
Consider introducing operations `fetch` and `list`, which are designed to fetch one or many resource from commcare

This is NOT a HTTP API and should not expose HTTP semantics. Users do not pass a URL, they just request a resource.

Signatures like

```
get(resourceType, options)
list(resourceType, options)
```

Options might include:
- one or more resource ids to fetch
- maybe some query parameters, as a JSON object
- the API version to use (would affect the parameters available I guess)
- Pagination semantics maybe
- A callback function for pagination? Or would that be a different argument?

PiusKariuki:
## Summary

Added 2 helper functions to `adaptor.js`: `list` & `getResource`

Fixes #1742 

#### `requestWithPagination` helper in util.js
- Cursor-based pagination over API V2.

#### `list` operation
- Wraps `requestWithPagination`

#### `getResource` operation
- Get a single or mutiple resources by ID

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


PiusKariuki:
@hunterachieng this now ready to review. See description for summary

packages/msgraph/ast.json
```diff
@@ -177,7 +177,7 @@
               }
             },
             "name": "callback",
-            "default": "s"
+            "default": "s=>s"
```
mtuchi: @PiusKariuki why is this file changed ?

packages/surveycto/package.json
```diff
@@ -22,6 +22,7 @@
   ],
   "dependencies": {
     "@openfn/language-common": "workspace:*",
+    "date-fns": "^4.1.0",
```
mtuchi: why was this added ? @PiusKariuki , I am not seeing changes in surveycto. If there are changes that require dateFns, i would suggest you use dateFns from language-common and update pr notes

packages/commcare/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@openfn/language-commcare",
   "label": "CommCare",
-  "version": "4.1.2",
+  "version": "4.2.0",
```
mtuchi: 🤔, Something is not right here @PiusKariuki did you rebase `main` branch ?

mtuchi (changes requested):
Hey @PiusKariuki can you rebase from main and make sure your branch is up to date and doesn't contain other packages changes eg: `msgprah, fhir-opensrp or surveycto`

PiusKariuki:
@mtuchi 
I have made the changes you have requested
- Rebased from main
- Removed files not related to this pr: `msgraph/ast.json`, `fhir-opensrp/ast.json`, `surveycto/package.json`

Note: Notice that the build is now failing in the GIT status check. This was the reason I included the files above in the PR since this step kept failing. Do you have any recommendations on how to fix this in a cleaner way? See error below
```
Run pnpm run test:git

> adaptors@1.0.0 test:git /home/runner/work/adaptors/adaptors
> pnpm exec scripts/status-check.sh

Checking if git working tree is clean...
❌ Git working tree has uncommitted changes

 M packages/msgraph/ast.json

Note: ast.json changes are often caused by the file not being checked into the build

Please commit or stash these changes before proceeding.
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

packages/commcare/src/Utils.js
```diff
@@ -72,3 +72,35 @@ export async function request(configuration, path, opts) {
 
   return commonRequest(method, path, options).then(logResponse);
 }
+
+export async function requestWithPagination(configuration, path, options = {}) {
+  const { domain, apiVersion = 'v2' } = configuration;
+  const { resultsKey } = options;
+  const targetUrl = `/a/${domain}/api/${path}/v2`;
```
mtuchi, line 79: @PiusKariuki i don't think we should hardcode version `/v2`, Are you sure all endpoints will be supported ?
what if i want to use `v1` ?

packages/commcare/src/Utils.js
```diff
@@ -72,3 +72,35 @@ export async function request(configuration, path, opts) {
...
+      params
+    });
+
+    const key = resultsKey ?? Object.keys(body).find(key => Array.isArray(body[key]));
+    if (key) results.push(...body[key]);
+
+    if (params.limit && results.length >= params.limit) break;
+    if (!body.next) break;
+
+    params.cursor = new URL(body.next).searchParams.get('cursor');
```
mtuchi, line 102: is cursor an actual commcare param ? or this is openfn option?

packages/commcare/src/Adaptor.js
```diff
@@ -14,6 +14,21 @@ import * as util from './Utils.js';
...
+ * @typedef {Object} ListOptions
+ * @property {string} [resultsKey] - The property where the results are stored e.g. `cases` for fetching case data
+ * @property {Object} [params] - HTTP query parameters passed through to CommCare
+ * @property {Object} [headers] - Additional HTTP request headers
+ */
+
+/**
+ * Options provided to the `getResource()` operation.
+ * @typedef {Object} GetResourceOptions
+ * @property {Object} [params] - HTTP query parameters passed through to CommCare.
```
mtuchi, line 28: CommCare query parameters is clear i think

packages/commcare/src/Adaptor.js
```diff
@@ -36,6 +51,124 @@ export function execute(...operations) {
...
+    try {
+      const data = await util.requestWithPagination(state.configuration, resolvedResourceType, resolvedOptions);
+      const nextState = util.prepareNextState(state, data);
+
+      return {
+        ...nextState,
+        data
+      };
+    } catch (e) {
+      if (e.statusCode === 404) {
```
mtuchi, line 100: what's the purpose of this ? why we only care about 404?

packages/commcare/src/Adaptor.js
```diff
@@ -14,6 +14,21 @@ import * as util from './Utils.js';
...
+ * List Operation Options
+ * @typedef {Object} ListOptions
+ * @property {string} [resultsKey] - The property where the results are stored e.g. `cases` for fetching case data
+ * @property {Object} [params] - HTTP query parameters passed through to CommCare
+ * @property {Object} [headers] - Additional HTTP request headers
+ */
+
+/**
+ * Options provided to the `getResource()` operation.
+ * @typedef {Object} GetResourceOptions
```
mtuchi, line 27: If it's just a single property do we really need typedef ?
Maybe we should just document
```
@property {Object} [params] - CommCare query params @link <link to the docs>
```

packages/commcare/src/Adaptor.js
```diff
@@ -14,6 +14,21 @@ import * as util from './Utils.js';
  * @private
  */
 
+/**
+ * List Operation Options
+ * @typedef {Object} ListOptions
```
mtuchi, line 19: Are there other ListOptions key other than `resultsKey` ?

packages/commcare/src/Adaptor.js
```diff
@@ -36,6 +51,124 @@ export function execute(...operations) {
...
+ * @returns {Operation}
+ */
+export function list(resourceType, options = {}) {
+  return async state => {
+    const [resolvedResourceType, resolvedOptions] = expandReferences(state, resourceType, options);
+    try {
+      const data = await util.requestWithPagination(state.configuration, resolvedResourceType, resolvedOptions);
+      const nextState = util.prepareNextState(state, data);
+
+      return {
```
mtuchi, line 95: I think we just need to `return nextState` no need for  `{ ...nextState, data}`, thats what prepareNextState do

mtuchi (changes requested):
Hey @PiusKariuki , left couple of comments here, please review them. Also see my comments on CommCare url scheme change from the previous pr 👇🏽  

- https://github.com/OpenFn/adaptors/pull/1753#issuecomment-5204969461
- https://github.com/OpenFn/adaptors/pull/1753#issuecomment-5205009374

packages/commcare/src/Utils.js
PiusKariuki: @mtuchi The challenge here is that CommCare `v1` and `v2` APIs have semantic differences when it comes to pagination.
- `V1` uses the common `limit` and `offset`
- `V2` uses a `cursor` to paginate.
- The shape of the response is also different depending on what version you are using. 
That is the reason I am using version 2 so I can be able to control what shape the response comes in and how I move on to the next batch.

The docs state on that 2 CommCare resources versioned as v1/v2: the `Case Data API` and the `Location API`. On other endpoints V2 is just a superset of V1 so anything you can do in v1 you could also do in v2

packages/commcare/src/Utils.js
PiusKariuki: The cursor is an actual param in v2. It is is how version 2 paginates. Once you have fetched a page CommCare returns a `next` link which you can then follow to get the next page. Heres is an example of a response

```
{
	"matching_records": 46164,
        "cases": [...], 
        "next": "https://eu.commcarehq.org/a/example/api/case/v2?cursor=aW5kZXhlZF9vbi5ndGU9MTc3NjkxNTI5OTMzMSZsYXN0X2Nhc2VfaWQ9MzA1MTdiOGQ2MmRiNGUzN2FiNjJjMDhhZDk0ZGYxMGY%3D"
        
}
```

packages/commcare/src/Adaptor.js
PiusKariuki: @mtuchi 
This is just me following convention I have seen in the rest of the operations in the same adaptor like `get` further down the file.

On reasearching this I noticed that CommCare will send you a HTML body when you hit an endpoint that doesn't exist so it is important we intercept this 404 and send the user an error they can understand. See below for an example response from the server when I hit a non-existent endpoint.

```
<!DOCTYPE html>


<!--[if lt IE 7]><html lang="en" class="lt-ie9 lt-ie8 lt-ie7"><![endif]-->
<!--[if IE 7]><html lang="en" class="lt-ie9 lt-ie8"><![endif]-->
<!--[if IE 8]><html lang="en" class="lt-ie9"><![endif]-->
<!--[if gt IE 8]><!--><html lang="en"><!--<![endif]-->
  <head>
    
    
    <title>
      404 Not Found
      - 
      CommCare HQ
    </title>

    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta http-equiv="content-language" content="en">

    
    <link rel="shortcut icon" href="/static/hqwebapp/images/favicon.png?version=a0e1c41" />
    <link type="application/opensearchdescription+xml"
          rel="search"
          href="/searchDescription.xml"
          title="CommCare HQ - eu" />

    <link href="https://fonts.googleapis.com/css?family=Nunito+Sans:100,300,400,500,700,800" rel="stylesheet">

    
      


<link rel="stylesheet" href="/static/CACHE/css/output.4e2692151041.css" type="text/css" media="all">

    

    
      <link rel="apple-touch-icon" sizes="57x57" href="/static/hqstyle/images/commcare-icon-57x57.png?version=8bfb2db"/>
      <link rel="apple-touch-icon" sizes="76x76" href="/static/hqstyle/images/commcare-icon-76x76.png?version=b4f6739"/>
      <link rel="apple-touch-icon" sizes="120x120" href="/static/hqstyle/images/commcare-icon-120x120.png?version=ffbac30"/>
      <link rel="apple-touch-icon" sizes="152x152" href="/static/hqstyle/images/commcare-icon-152x152.png?version=75e8399"/>
      <link rel="apple-touch-icon" sizes="167x167" href="/static/hqstyle/images/commcare-icon-167x167.png?version=925a328"/>
      <link rel="apple-touch-icon" sizes="180x180" href="/static/hqstyle/images/commcare-icon-180x180.png?version=d20ea3f"/>
    

    

    <link rel="stylesheet" href="/static/CACHE/css/output.a4ed04f99b78.css" type="text/css" media="all">

    <script>
      window.USE_BOOTSTRAP5 = true;
    </script>

    
    
  </head>
  <body>

    
      <input id="csrfTokenContainer" type="hidden" value="Fw5VmCQgDxG9HLcJyBvSmgRll9goBpczHPJhwvnHFSJ6Wl5Xva3umPXrs8LZyNv5"

File: packages/commcare/src/Adaptor.js (hunk 8 above)
Comment (PiusKariuki) [reply packages/commcare/src/Adaptor.js:27]: @mtuchi 
- It is not just one property. We have `params` and `headers`.

File: packages/commcare/src/Adaptor.js (hunk 9 above)
Comment (PiusKariuki) [reply packages/commcare/src/Adaptor.js:19]: @mtuchi 
- We have the the `params` and the `headers`.

File: packages/commcare/src/Adaptor.js (hunk 10 above)
Comment (PiusKariuki) [reply packages/commcare/src/Adaptor.js:95]: @mtuchi I tried it locally and it returned state without a `data` object

**Comment by PiusKariuki** [comment]
@mtuchi please take a quick look at this PR when you can.

**Comment by mtuchi** [comment]
@PiusKariuki can you check the box in review checklist

File: packages/commcare/src/Utils.js [hunk 11]
```diff
@@ -72,3 +72,35 @@ export async function request(configuration, path, opts) {
 
   return commonRequest(method, path, options).then(logResponse);
 }
+
+export async function requestWithPagination(configuration, path, options = {}) {
+  const { domain, apiVersion = 'v2' } = configuration;
```
mtuchi: @PiusKariuki `apiVersion` is expanded but not used,

packages/commcare/src/Utils.js
mtuchi: I don't understand what you mean by `V2 uses a cursor to paginate.` but my problem is `getResource()` will always use `v2` because it's hard coded regardless if i specify `apiVersion` in my configuration. Maybe the that's fine and we should document in the function `this function uses v2` only. 

If we want to support auto pagination then regardless of v1 or v2 `getResource()` should support auto pagination and not force the function to only use `v2`.

mtuchi: Maybe @josephjclark have different opinion  on this

mtuchi (changes requested):
Hey @PiusKariuki i left you couple, Please address them and ask joe for final review

packages/commcare/src/Utils.js
PiusKariuki: I think the question here @josephjclark is,  how opinionated should the `requestWithPagination` function be when it comes to the API version ?

 The 2 API versions handle pagination very differently and give the response in different shapes so I made the call to use v2.

PiusKariuki:
Hi @josephjclark could you please have a look at this when you get the time?

josephjclark:
I can look but I don't see an approve from @mtuchi  yet.

There are also conflicts because the version was bumped too earlier.

Please submit a changeset file with all changes, but don't bump versions until we're ready to merge (usually I'll ask when ready for that)

i recommend you revert the package.json and changelog files.

josephjclark:
I don't know where there are so many changes in the package lock either - you should probably revert those

mtuchi (changes requested):
@PiusKariuki 

  * **Extract HTTP gate utility logic -** to support pagination for V1 and V2 in the list function.
  * **Review 404 handling -** test removing the 404 protection logic to check if default headers work properly. use `parseAs: json`

PR #1762, opened 2026-08-14
 msgraph: wrap `uploadFile` upload-session metadata in an `item` property

### LINKED ISSUE #1761: `msgraph`: `uploadFile` ignores `onConflict` — always replaces existing files
### Describe the bug
When calling `uploadFile` with `onConflict` set to `"fail"` or `"rename"`, the setting has no effect. Microsoft Graph always falls back to its default behaviour and **replaces** the existing file, regardless of the value passed.

### Adaptor
`@openfn/language-msgraph@0.10.0`

### To Reproduce
Upload a file to a drive/folder where a file with the same name already exists:

```js
uploadFile(
  state => ({
    driveId: state.driveId,
    folderId: state.folderId,
    fileName: 'Report.zip',
    onConflict: 'fail', // also tried 'rename'
  }),
  state => state.buffer
);
```

**Expected:** the upload fails (or creates a renamed copy).
**Actual:** the existing file is silently replaced.

### Root cause
In `uploadFile`, the `createUploadSession` request body places the conflict behavior and name at the top level:

```js
body: JSON.stringify({
  '@microsoft.graph.conflictBehavior': onConflict,
  name: fileName,
}),
```

Per the [createUploadSession API docs](https://learn.microsoft.com/en-us/graph/api/driveitem-createuploadsession), these must be nested inside an `item` object (`driveItemUploadableProperties`). Because they're sent at the top level, Graph ignores them and uses its default (`replace`).

### Suggested fix
Wrap the properties in `item`:

```js
body: JSON.stringify({
  item: {
    '@microsoft.graph.conflictBehavior': onConflict,
    name: fileName,
  },
}),
```

### Environment
- Adaptor: `@openfn/language-msgraph@0.10.0`
- Affected function: `uploadFile`

mtuchi:
## Summary

Fix `onConflict` option in `uploadFile()` function 

Fixes #1761

## Details

The `createUploadSession` request body was sending the file metadata
flat:

```js
body: JSON.stringify({
  '@microsoft.graph.conflictBehavior': onConflict,
  name: fileName,
})
```

MSGraph expects that metadata nested under a single `item` key of type `driveItemUploadableProperties`. Sent flat, the conflict behaviour and filename were ignored. This PR wraps them:

```js
body: JSON.stringify({
  item: {
    '@microsoft.graph.conflictBehavior': onConflict,
    name: fileName,
  },
})
```

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/eighty-pugs-tickle.md
+++ b/.changeset/eighty-pugs-tickle.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-msgraph': patch
----
-
-Fix `uploadFile` to wrap upload session metadata in an `item` property, as required by the Microsoft Graph createUploadSession API
--- a/.changeset/eighty-pugs-tickle.md
+++ b/.changeset/eighty-pugs-tickle.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-msgraph': patch
+---
+
+Fix `uploadFile` to wrap upload session metadata in an `item` property, as required by the Microsoft Graph createUploadSession API
--- a/.changeset/public-planes-dig.md
+++ b/.changeset/public-planes-dig.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-commcare': patch
----
-
-Fix an issue in docs which caused mdx parsers to fail
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,6 @@ dist
 .idea
 
 /pnpm-publish-summary.json
+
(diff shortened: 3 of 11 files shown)


mtuchi:
Ooh man 🤦‍♂️, I forgot to run `pnpm run version` for the latest commcare adaptor version. It will be included in this release

PR #1763, opened 2026-08-15
feat(chatgpt,claude): allow users to specify `baseUrl`

LoneRifle:
## Summary

Allow users to specify an OpenAI-compatible (for `chatgpt`) or Claude-compatible (for `claude`) endpoint
to the corresponding adaptors, for the following use cases:

- Routing requests through a reverse proxy
- Sending requests to self-hosted infrastructure, eg: Ollama, vLLM
- Replacing OpenAI or Claude with alternatives, eg:
  - [Mistral](https://docs.mistral.ai/resources/migration-guides), [Groq](https://console.groq.com/docs/openai) (OpenAI-compatible)
  - [DeepSeek](https://api-docs.deepseek.com/guides/anthropic_api/) (Claude-compatible)

## Details

- Add `baseUrl` as an optional property in both `chatgpt` and `claude` config schemas
- When present, pass `baseUrl` along with `apiKey` to the OpenAI or Anthropic client constructor

## AI Usage

know!):

- [x] I have not used AI

## Review Checklist

### DIFF
--- a/packages/chatgpt/src/Adaptor.js
+++ b/packages/chatgpt/src/Adaptor.js
@@ -28,11 +28,10 @@ let client;
  * @returns {state}
  */
 export function createClient(state) {
-  const { apiKey } = state.configuration;
+  const { apiKey, baseUrl } = state.configuration;
 
-  client = new OpenAI({
-    apiKey,
-  });
+  const clientOptions = baseUrl ? { apiKey, baseUrl } : { apiKey };
+  client = new OpenAI(clientOptions);
 
   return state;
 }
--- a/packages/chatgpt/src/Adaptor.js
+++ b/packages/chatgpt/src/Adaptor.js
@@ -30,8 +30,10 @@ let client;
 export function createClient(state) {
   const { apiKey, baseUrl } = state.configuration;
 
-  const clientOptions = baseUrl ? { apiKey, baseUrl } : { apiKey };
-  client = new OpenAI(clientOptions);
+  client = new OpenAI({
+    apiKey,
+    baseUrl,
+  });
 
   return state;
 }
--- a/packages/claude/src/Adaptor.js
+++ b/packages/claude/src/Adaptor.js
@@ -24,8 +24,10 @@ let client;
 export function createClient(state) {
   const { apiKey, baseUrl } = state.configuration;
 
-  const clientOptions = baseUrl ? { apiKey, baseUrl } : { apiKey };
-  client = new Anthropic(clientOptions);
+  client = new Anthropic({
+    apiKey,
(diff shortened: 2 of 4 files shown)


packages/claude/src/Adaptor.js
```diff
@@ -22,11 +22,10 @@ let client;
  * @returns {state}
  */
 export function createClient(state) {
-  const { apiKey } = state.configuration;
+  const { apiKey, baseUrl } = state.configuration;
 
-  client = new Anthropic({
-    apiKey,
-  });
+  const clientOptions = baseUrl ? { apiKey, baseUrl } : { apiKey };
```
josephjclark: not a big deal but can you not just pass `baseUrl` as a falsy value? Then you can remove this complicated ternary. Alternatively we could consider defaulting the value (although I'd rather let the client set the default for us)

packages/chatgpt/src/Adaptor.js
```diff
@@ -28,11 +28,10 @@ let client;
...
  */
 export function createClient(state) {
-  const { apiKey } = state.configuration;
+  const { apiKey, baseUrl } = state.configuration;
 
-  client = new OpenAI({
-    apiKey,
-  });
+  const clientOptions = baseUrl ? { apiKey, baseUrl } : { apiKey };
+  client = new OpenAI(clientOptions);
```
josephjclark: Same here - I'm sure you can just forward `apiKey` through and let the client set a default value

josephjclark:
@LoneRifle Thanks for raising this, looks OK to me. Just a coupe of nitpicks on the syntax which I think are worth doing

packages/claude/src/Adaptor.js
LoneRifle: I indeed could, I was just being conservative about client ctor behaviour wrt falsy values. Tweaked this as well as the equiv line in the `chatgpt` adaptor.

josephjclark (approved):
Great, thank you!

LoneRifle:
Just to point out that the Node 22.20.0 test failure appears to be a timing issue, corroborated by passes for Nodes 24 and 26:
```sh
packages/memento test:   1 failing
packages/memento test:   1) handleRateLimit
packages/memento test:        should log appropriate messages:
packages/memento test:      AssertionError: expected undefined to equal 'Waiting 0.2s before next request'
packages/memento test:       at Context.<anonymous> (file:///home/runner/work/adaptors/adaptors/packages/memento/test/util.test.js:101:35)
packages/memento test: Failed
```

josephjclark:
Thanks @LoneRifle ! Staging for release now

PR #1767, opened 2026-08-19
Add `util.derieveUuid()` in `dhis2` adaptor 

### LINKED ISSUE #1771: Add deterministic DHIS2 UID generation in `dhis2` adaptor
## Use case

I am building a workflow (Kobo Collect → DHIS2 event program) that must be **idempotent**: re-posting the same source record has to *update* the existing DHIS2 event rather than create a duplicate.

The standard way to do that with `POST /api/tracker` is to supply the event UID ourselves and use `importStrategy=CREATE_AND_UPDATE`. The UID must be derived **deterministically** from a business key, so the same key always yields the same UID. This is not specific to our project: any workflow doing idempotent upserts into DHIS2 from an external source needs exactly this.

## The gap

There is no way to hash anything inside an OpenFn job. Jobs execute in a bare V8 context (`vm.createContext()`), which provides only the ECMAScript built-ins — no Node globals. Verified in the sandbox:

| Global | Available |
|---|---|
| `crypto`, `TextEncoder`, `Buffer`, `process` | **no** |
| `Math`, `JSON`, `BigInt`, `Uint8Array`, `DataView`, `String.prototype.normalize` | yes |

(The runtime already injects `Buffer`, `console` and the timers explicitly, which
is precisely because a fresh vm context has none of them.)

## What I had to write

**136 lines** inside the job body, since jobs cannot use `import`:

- the 64 SHA-256 round constants and a full SHA-256 implementation — no `crypto`
- a UTF-8 encoder — no `TextEncoder`
- long division over a byte array — see the compiler note below
- base-62/52 encoding into the valid DHIS2 UID space

This is a lot of security-sensitive code to carry in a job file, and it is easy to get subtly wrong. A DHIS2 UID is 1 letter followed by 10 alphanumerics, i.e. 52 × 62^10 ≈ 2^65.2 values, so the correct construction is a **mixed-radix mapping**, not a truncated hash — the obvious naive implementations either produce invalid UIDs or waste entropy and collide.

## Secondary issue: the compiler and the runtime disagree on the language

I would have used `BigInt` for the base conversion — the runtime provides it — but the **OpenFn compiler rejects BigInt literals**: `0n` fails with `Unexpected token ILLEGAL` at build time. So a job can be valid JavaScript, run correctly under Node, and still be undeployable.

## Proposal

Either would solve it for me; the first is the smaller change.

1. **Ship `deriveUid(seed)` in `@openfn/language-dhis2`** — a deterministic seed → valid-DHIS2-UID function. Small, pure, and the natural home for the DHIS2-specific UID format rules.

2. **Expose a hashing primitive in the job runtime** (e.g. a `crypto` global with  `createHash`, or WebCrypto's `crypto.subtle`). More general — adaptors and jobs could both build on it.

## Anticipating one answer

`GET /api/system/id` does not solve this: the UIDs it returns are **random**, so they cannot provide idempotency, and it would cost one round-trip per record. Determinism is the whole point — same business key → same UID → update instead of duplicate.

## Reference implementation

The DHIS2-specific part:

```js
const B62 = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
const ALPHA = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';

// Divide a big-endian byte array by a small divisor (BigInt avoided: see above).
function divmodSmall(bytes, divisor) {
  const quotient = new Array(bytes.length);
  let rem = 0;
  for (let i = 0; i < bytes.length; i++) {
    const cur = rem * 256 + bytes[i];
    quotient[i] = Math.floor(cur / divisor);
    rem = cur % divisor;
  }
  return { quotient, rem };
}

// A valid DHIS2 UID is 1 letter + 10 alphanumerics = 52 x 62^10 values.
// Read the hash's first 16 bytes as one integer N: the 10 tail characters are
// N mod 62^10 in base 62, the leading letter is floor(N / 62^10) mod 52.
// Reducing 128 bits into ~65 keeps the modulo bias negligible.
function deriveUid(key) {
  const h = sha256Bytes(key);            // our own SHA-256, 32 bytes
  let cur = Array.prototype.slice.call(h, 0, 16);
  const tail = new Array(10);
  for (let i = 9; i >= 0; i--) {
    const { quotient, rem } = divmodSmall(cur, 62);
    tail[i] = B62[rem];
    cur = quotient;
  }
  return ALPHA[divmodSmall(cur, 52).rem] + tail.join('');
}
```

Validated: 
- byte-identical to Node's `crypto` SHA-256 (UTF-8 and every padding
boundary), always matches `/^[A-Za-z][A-Za-z0-9]{10}$/`, and zero collisions over
500,000 distinct keys.

mtuchi:
## Summary

Add `util.deriveUid()`, which derives a stable DHIS2 UID from a seed string, so workflows can set a resource's id before creating it and be safely re-run without creating duplicates.

Fixes #1771

## Details
 
Idempotent creates against the tracker API need a UID that is known before the request is made and identical on every run. DHIS2 has no deterministic UID facility (`/api/system/id` and the `dhis2-uid` package both mint random values),
so workflows either look up the existing record first or hand-roll the derivation in job code. 
 
Notes for reviewers:
- The seed is hashed exactly as given, with no adaptor-supplied namespace.  Callers own seed uniqueness, and existing hand-rolled job code keeps producing identical ids after switching over.
- String seeds only; throws on anything else rather than stringifying, since  accepting objects would make output depend on key insertion order.
- The seed → UID mapping is frozen once released, as changing it would orphan existing records. The tests pin it with hard-coded vectors.

### Usage
 
```js
create('tracker', state => ({
  events: [
    {
      event: util.deriveUid(`event:${state.orgUnit}:${state.period}`),
      program: state.program,
      orgUnit: state.orgUnit,
      occurredAt: state.date,
    },
  ],
}));
```

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist


packages/dhis2/src/util.js
```diff
@@ -206,3 +207,73 @@ export async function request(configuration, requestData) {
 
   return commonRequest(method, path, opts).then(logResponse);
 }
+
+const UID_HEAD_CHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
+const UID_TAIL_CHARS =
+  '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
+
+/**
+ * Derives a stable DHIS2 UID from a seed string. The same seed always returns
```
josephjclark: Docs here are way too long, I'd just put the critical information in the public docs and leave more detailed stuff as comments in the code.

Don't let the AI ramble too much Mtuchi. Docs should be concise and relevant

packages/dhis2/src/util.js
```diff
@@ -206,3 +207,73 @@ export async function request(configuration, requestData) {
...
+ * release, because doing so would orphan every record already created with it.
+ *
+ * @public
+ * @function
+ * @namespace util
+ * @param {string} seed - The string to derive the UID from.
+ * @returns {string} An 11-character DHIS2 UID, matching /^[A-Za-z][A-Za-z0-9]{10}$/
+ * @example <caption>Derive an event UID from an org unit and period</caption>
+ * fn(state => {
+ *   const uid = util.deriveUid(`event:${state.orgUnit}:${state.period}`);
```
josephjclark, line 225: why are you prefixing these string with `event:`? They'll all have it so it's just bloating the id right? And I don't think the hash needs to be reversible or readable?

josephjclark (commented):
Looks good to me and to Claude. Can we just tidy the docs up? We don't need all the cryptographic detail in public docs

packages/dhis2/src/util.js
mtuchi: Anhaaa so sorry joe i missed this fixing it now

mtuchi: pushed changes

packages/dhis2/src/util.js
mtuchi: an id for event is the same as an id for tei, If you use `{state.orgUnit}:${state.period}` only you will get an id, but you have to be very careful not to use it ina different place

josephjclark: ah, and you're doing the same sort of encoding for teis and events?

mtuchi: No for current use case, but you can do it

josephjclark:
Merged into #1773

PR #1768, opened 2026-08-20
Commcare: implement pagination in http.get

### LINKED ISSUE #1739: commcare: add pagination support to http.get
The now-deprecated `get()` function will paginate commcare automatically, pulling down multiple pages of data.

We should add the same pagination support to `http.get()`, although with a few deviations from our usual patterns:

1. Pagination should be disabled by default. The `http` namespace is supposed to be dumb. So users can pass an option like `paginate: true` or `pages: 10` or `paginatiom: 'auto'` to opt-in to automatic pagination.
1. Commcare uses two styles of pagination. Most APIs use page/offset (which is what's implemented in `get()` right now). But some use cursor positions to handle pagination.  I would like the adaptor to automatically detect the style of pagination and jsut handle it. Of course users may pass query arugments like `http.get($.url, { page: 2 })`, which would give a clue.
1. The problem with pagination is that users can download a million records and blow memory, which doens't help. We should start providing smart pagination APIs which invoke a callback for each page of data. In this callback mode, each page of data is passed to the caller to process, and the whole dataset is NOT returned to state.data (yes, I know callbacks have been largely removed, bit this is a case more like `each()` in which callbacks are correct and good). If using callbacks, each page must be discarded by the adaptor after use and must not be trapped in a closure or saved to state.

I wonder if we can do the callback as something like:

```
http.request($.url).each(() => {})
```

That's neater than taking a function on an object

hunterachieng:
## Summary

Implement pagination in `http.get()`.

Fixes #1739 

## Details

Add opt-in pagination to `http.get`

`http.get` now supports two pagination modes, both disabled by default:

- **Accumulate** — pass `paginate: true` to fetch all pages and return them
  as a single array in `state.data`:
  ```js
  http.get('case/v1', { paginate: true });
  ```
- **Stream** — pass a callback as the third argument to process each page
  individually without accumulating the full result in `state.data`:
  ```js
  http.get('case/v1', { paginate: true }, state => {
    // state.data is the current page's records only
    return state;
  });
  ```

Both modes support the v1 (offset/limit) and v2 (cursor) CommCare APIs.
In stream mode `state.data` is `{}` after the operation completes.

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/quiet-eggs-yawn.md
+++ b/.changeset/quiet-eggs-yawn.md
@@ -1,0 +1,24 @@
+---
+'@openfn/language-commcare': minor
+---
+
+Add opt-in pagination to `http.get`
+
+`http.get` now supports two pagination modes, both disabled by default:
+
+- **Accumulate** — pass `paginate: true` to fetch all pages and return them
+  as a single array in `state.data`:
+  ```js
+  http.get('case/v1', { paginate: true });
+  ```
+- **Stream** — pass a callback as the third argument to process each page
+  individually without accumulating the full result in `state.data`:
+  ```js
+  http.get('case/v1', { paginate: true }, state => {
+    // state.data is the current page's records only
+    return state;
+  });
+  ```
+
+Both modes support the v1 (offset/limit) and v2 (cursor) CommCare APIs.
+In stream mode `state.data` is `{}` after the operation completes.
--- a/.changeset/quiet-eggs-yawn.md
+++ b/.changeset/quiet-eggs-yawn.md
@@ -1,5 +1,0 @@
----
-'@openfn/language-commcare': minor
----
-
-Add opt-in pagination to `http.get`. Pass `params.paginate: true` to accumulate all pages into `state.data`, or also provide a callback to stream each page without accumulating.
--- a/.changeset/quiet-eggs-yawn.md
+++ b/.changeset/quiet-eggs-yawn.md
@@ -2,23 +2,4 @@
 '@openfn/language-commcare': minor
 ---
 
-Add opt-in pagination to `http.get`
-
-`http.get` now supports two pagination modes, both disabled by default:
-
-- **Accumulate** — pass `paginate: true` to fetch all pages and return them
-  as a single array in `state.data`:
-  ```js
-  http.get('case/v1', { paginate: true });
-  ```
-- **Stream** — pass a callback as the third argument to process each page
-  individually without accumulating the full result in `state.data`:
-  ```js
-  http.get('case/v1', { paginate: true }, state => {
-    // state.data is the current page's records only
-    return state;
-  });
-  ```
-
-Both modes support the v1 (offset/limit) and v2 (cursor) CommCare APIs.
-In stream mode `state.data` is `{}` after the operation completes.
+Add opt-in pagination to `http.get`. Pass `params.paginate: true` to accumulate all pages into `state.data`, or also provide a callback to stream each page without accumulating.
--- a/packages/commcare/src/http.js
+++ b/packages/commcare/src/http.js
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
  * Make a GET request to CommCare. Use this to retrieve resources directly from the CommCare REST API.
  * @example <caption>Get cases using the v1 API (prefixed with /a/[domain]/api/)</caption>
  * http.get('case/v1');
+ * @example <caption>Get all cases paginated into state.data </caption>
+ * http.get('case/v2', { paginate: true });
+ * @example <caption>Get all cases paginated and return each response to a callback</caption>
+ * http.get('case/v2', { paginate: true }, (state) => {
+ *   console.log(state.data); // one page at a time
+ *   return state;
+ * });
  * @function
  * @public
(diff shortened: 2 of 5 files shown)


.changeset/quiet-eggs-yawn.md
```diff
@@ -0,0 +1,24 @@
+---
+'@openfn/language-commcare': minor
+---
+
+Add opt-in pagination to `http.get`
```
josephjclark: the changelog is way too long - it should be very terse, sjust acouple of sentences at most.

What you've add here is docs. But I don;t think this level of documentation has been added anywhere else? So for users to learn how pagination works, they ave to consult the changelog.

I think you should move this to overview docs or the `http.get` docs. You've got examples to cover it but you need a general high level explanation

packages/commcare/src/http.js
```diff
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
...
+ * @example <caption>Get all cases paginated and return each response to a callback</caption>
+ * http.get('case/v2', { paginate: true }, (state) => {
+ *   console.log(state.data); // one page at a time
+ *   return state;
+ * });
  * @function
  * @public
  * @param {string} path - Path to resource. Relative paths are prefixed with `/a/[domain]/api/`; paths starting with `/` are used as-is.
- * @param {Object} [params] - Optional request params
+ * @param {Object} [params] - Optional request params. Set `paginate: true` to enable automatic pagination.
```
josephjclark: Don't put the paginate docs here. Better to do `params.paginate` or something as we do in other docs.

While we're here, we should clearly not that the params (apart from the openfn ones) are appended to the url as query strings. That needs clearly documenting.

packages/commcare/src/http.js
```diff
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
...
+  return async state => {
+    const { domain } = state.configuration;
+    const [resolvedPath, resolvedParams] = expandReferences(
+      state,
+      path,
+      params,
+    );
+
+    const { paginate, ...requestParams } = resolvedParams;
+    const shouldPaginate = paginate === true;
```
josephjclark: this variable is only used once. It would be cleaner to remove it

packages/commcare/src/http.js
```diff
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
...
+    // Return response data to callback if provided
+    const callbackFn = typeof callback === 'function';
```
josephjclark: maybe this should be called `callbackmode` or something. Then again, it also is only used once. Wouldnt it be clearer just to put `typeof callback === 'function'` in the condition below?

packages/commcare/src/http.js
```diff
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
...
+      do {
+        const response = await util.request(state.configuration, url, {
+          method: 'GET',
+          params: currentParams,
+          contentType: 'application/json',
+        });
+
+        const body = response.body;
+        // Only return response array data
+        const items = Object.values(body).find(Array.isArray) ?? [];
```
josephjclark, line 86: I'm not quite sure I understand what this is doing - a better comment might help

is it just assuming that all array values in the response object are response data? That makes me a little nervous. Then again, relying on a structure like `{ objects. meta }` makes me a bit bervous too. The different APIs have different structures.

This might be OK for now. Maybe we need to take a key or function from the user to help parse the response

packages/commcare/src/http.js
```diff
@@ -36,15 +36,92 @@ export function post(path, data, params = {}) {
...
+
+          const pageState = util.prepareNextState(state, response);
+          nextState = await callback({ ...pageState, data: items });
+        } else {
+          nextState = util.prepareNextState(state, response);
+          if (!results) results = [];
+          results.push(...items);
+        }
+
+        const v1Next = body?.meta?.next;
```
josephjclark: Rather than making this a v1/v2 style lookup, can we structure it around a cursor/page lookup? So if there's a `body.next` we treat this as a cursor API, otherwise we assume it's a paging API

packages/commcare/test/http.test.js
```diff
@@ -55,18 +76,279 @@ describe('http.get', () => {
...
 
-    const finalState = await http.get(
-      '/a/test-staging/api/case/v1/'
-    )(baseState);
+    const finalState = await http.get('/a/test-staging/api/case/v1/')(baseState);
 
     expect(finalState.data).to.eql([{ case_id: 'abc-002' }]);
   });
+
+  describe('no pagination (default)', () => {
```
josephjclark: don't like seeing `default` in this test group - it's irrelevant.

josephjclark (commented):
Pretty good! Tests are nice. Just a few things I'd like to look at before we can merge.

packages/commcare/src/http.js
hunterachieng: I used that to ensure we just get the array values which carries the output instead of doing `{response.body.objects || response.body.cases}

packages/commcare/src/http.js
hunterachieng: Its being used in two other areas. I am renaming it to `callbackMode`

josephjclark (approved):

PR #1770, opened 2026-08-21
Commcare: deprecate `get`, `post`, and `request` functions

### LINKED ISSUE #1741: commcare: deprecate get, post, request
The latest commcare adaptor, 4.x, is hard-coded not just to the 0.5, but uses the old URL structure.

The adaptor is designed around commcare's old API, which uses the prefix `a/domain/v0.5/api/<resource>`. But the latest commcare APIs use the form `a/domain/api/<resource>/v0.5/`

I don't think we can change `get`, `post()` and `request` without introducing a breaking change. So we should deprecate them now, find a way to notify users, and remove them in a year or so.

These functions should all point to the 0.5 docs, not to the latest docs (I think they do already)

Docs for these functions MUST make it clear that they only work against the 0.5 API, and to use other path structures  users shoud use the `request` function (or, if it's implemented, the http namespace)

hunterachieng:
## Summary

Deprecate `get`, `post`, and `request`. These functions only work with CommCare's legacy v0.5 API. Use `http.get`, `http.post`, and `http.request` instead.

Fixes #1741 

## AI Usage

know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/.changeset/thirty-seas-juggle.md
+++ b/.changeset/thirty-seas-juggle.md
@@ -3,3 +3,22 @@
 ---
 
 Deprecate `get`, `post`, and `request`. These functions only work with CommCare's legacy v0.5 API. Use `http.get`, `http.post`, and `http.request` instead.
+
+Note that the URL structure changes: the version segment moves from before the resource (`v0.5/case`) to after it (`case/v2`). Update the path itself, not just the function name.
+
+```js
+// before
+get('/case/12345');
+// after
+http.get('case/v2/12345');
+
+// before
+post('/user', { username: 'test', password: 'somepassword' });
+// after
+http.post('user/v2', { username: 'test', password: 'somepassword' });
+
+// before
+request('GET', '/a/asri/api/v0.5/case');
+// after
+http.request('GET', 'case/v2');
+```
--- a/.changeset/thirty-seas-juggle.md
+++ b/.changeset/thirty-seas-juggle.md
@@ -8,12 +8,12 @@ Note that the URL structure changes: the version segment moves from before the r
 
 ```js
 // before
-get('/case/12345');
+get('v0.5/case/12345');
 // after
 http.get('case/v2/12345');
 
 // before
-post('/user', { username: 'test', password: 'somepassword' });
+post('v0.5/user', { username: 'test', password: 'somepassword' });
 // after
 http.post('user/v2', { username: 'test', password: 'somepassword' });
 
--- a/.changeset/thirty-seas-juggle.md
+++ b/.changeset/thirty-seas-juggle.md
@@ -1,0 +1,5 @@
+---
+'@openfn/language-commcare': patch
+---
+
+Deprecate `get`, `post`, and `request`. These functions only work with CommCare's legacy v0.5 API. Use `http.get`, `http.post`, and `http.request` instead.
--- a/.changeset/thirty-seas-juggle.md
+++ b/.changeset/thirty-seas-juggle.md
@@ -4,21 +4,21 @@
 
 Deprecate `get`, `post`, and `request`. These functions only work with CommCare's legacy v0.5 API. Use `http.get`, `http.post`, and `http.request` instead.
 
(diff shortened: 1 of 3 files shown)


packages/commcare/src/Adaptor.js
```diff
@@ -38,29 +38,36 @@ export function execute(...operations) {
...
- * @param {string} path - Path to resource
- * @param {Object} [params] - Input parameters for the request. These vary by endpoint,  see {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957366/Data+APIs CommCare docs}.
+ * @param {string} path - Path to a v0.5 resource (e.g. `case`, `form`)
+ * @param {Object} [params] - Input parameters for the request. These vary by endpoint, see {@link https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143957366/Data+APIs CommCare v0.5 docs}.
  * @param {function} [callback] - Optional callback function. Invoked once per page of data retrieved.
  * @state {CommcareHttpState}
  * @returns {Operation}
  */
 export function get(path, params = {}, callback = s => s) {
+  console.warn(
```
josephjclark: hmm I think this might be a little too annoying on each request. @mtuchi  what do you think?

Maybe we should limit this to 1 warning per job?

.changeset/thirty-seas-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/language-commcare': patch
+---
+
+Deprecate `get`, `post`, and `request`. These functions only work with CommCare's legacy v0.5 API. Use `http.get`, `http.post`, and `http.request` instead.
```
josephjclark, line 5: We should include an example of how to convert, because job code will change right? With http users have to use a different url

.changeset/thirty-seas-juggle.md
```diff
@@ -0,0 +1,24 @@
...
+
+Note that the URL structure changes: the version segment moves from before the resource (`v0.5/case`) to after it (`case/v2`). Update the path itself, not just the function name.
+
+```js
+// before
+get('/case/12345');
+// after
+http.get('case/v2/12345');
```
josephjclark: Should be `v0.5/case/12345` right? because `get` uses the 0.5 API?

packages/commcare/src/Adaptor.js
hunterachieng: @josephjclark  For this section, do I implement only one warning?

.changeset/thirty-seas-juggle.md
josephjclark: No, that's not what I mean hunter

The old commcare `get` uses `get('/case/12345')`  which is hard coded to the 0.5 API

To get the same behaviour on the new http API, you have to do `http.get('v0.5/case/12345`)`.

We're trying to show how users can migrate to the same behaviour. So staying on the 0.5 API.

All "after" examples must use 0.5 API urls (ideally using the same structure as the adaptor uses now)

hunterachieng: Ah. Yes now I understand. I had misinterpreted what you meant

hunterachieng: @josephjclark  I have fixed it

josephjclark (approved):

packages/commcare/src/Adaptor.js
josephjclark: Oh dang I forgot this bit.  Yeah sorry - I think we should only do a console.log the first time any of these functions are used. Just use a local `let` to control it.

hunterachieng: @josephjclark  This is fixed

hunterachieng:
@josephjclark  I implemented the feedback on the deprecation messages

mtuchi (approved):

josephjclark (approved):
please bump versions and merge someone. Remember to use the `pnpm run version` command

PR #1774, opened 2026-08-25
Primero: add  a`get` function with  auto  pagination

### LINKED ISSUE #1765: primero: add pagination to get
### Description

- Implement an auto pagination in `primero` to ensure we get all the records present. 
- Create a generic `get()` in Adaptors.js and use the shared pagination function to paginate records with:
```
get('registry_records', { count: 20_000 })

```
- Reuse the pagination function in `getCases()`, `getLookups()`, `getForms()`, and `getReferrals()`

hunterachieng:
## Summary

- Add `get('cases', { per: 2 })` for making paginated GET requests to any Primero endpoint. 
- Added a reusable pagination function

Fixes #1765 

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Review Checklist

### DIFF
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -1,6 +1,6 @@
 import { execute as commonExecute } from '@openfn/language-common';
 import { expandReferences } from '@openfn/language-common/util';
-import { assembleError, scrubResponse, tryJson } from './Utils.js';
+import { assembleError, scrubResponse, tryJson, getNextPageParams, request as utilRequest } from './Utils.js';
 import request from 'request';
 
 export const composeNextState = (state, data, meta) => {
@@ -898,6 +898,47 @@ export function getLocations(query, callback) {
         }
       });
     });
+  };
+}
+
+/**
+ * Make a GET request to any Primero endpoint. Automatically paginates through all pages.
+ * @public
+ * @function
+ * @example <caption>Fetch all registry records</caption>
+ * get('registry_records');
+ * @example <caption>Fetch with a fixed page size of 50</caption>
+ * get('registry_records', { count: 50 });
+ * @param {string} path - Path to the resource
+ * @param {object} query - Query parameters to append to the URL. Use `count` (or `per`) to control the page size.
+ * @returns {Operation}
+ */
+export function get(path, query = {}) {
+  return async state => {
+    const [resolvedPath, resolvedQuery] = expandReferences(state, path, query);
+    const { count, per, ...queryParams } = resolvedQuery;
+
+    let currentQuery = { ...queryParams };
+    if (count ?? per) currentQuery.per = count ?? per;
+
+    let results = [];
+
+    do {
+      const response = await utilRequest(state, 'GET', resolvedPath, {
+        query: currentQuery,
+      });
+
+      results.push(...(response.body.data ?? []));
+
+      const next = getNextPageParams(response.body.metadata, resolvedQuery);
+      if (next) {
+        currentQuery = { ...queryParams, page: next.page, per: next.per };
+      } else {
+        break;
+      }
+    } while (true);
+
+    return composeNextState(state, results);
   };
 }
 
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -907,7 +907,7 @@ export function getLocations(query, callback) {
  * @function
  * @example <caption>Fetch all registry records</caption>
  * get('registry_records');
- * @example <caption>Fetch with a fixed page size of 50</caption>
+ * @example <caption>fetch all records in pages of 50 at a time</caption>
  * get('registry_records', { per: 50 });
  * @example <caption>Fetch at most 5000 records</caption>
  * get('registry_records', { limit: 5000 });
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -926,9 +926,9 @@ export function get(path, query = {}) {
       });
 
       const { data, metadata } = response.body;
-      results = results.concat(data);
+      results.push(...data);
 
-      const next = getNextPageParams(metadata, resolvedQuery);
+      const next = getNextPageParams(metadata);
       if (!next) break;
 
       currentQuery = { ...resolvedQuery, page: next.page, per: next.per };
--- a/packages/primero/src/Adaptor.js
+++ b/packages/primero/src/Adaptor.js
@@ -902,25 +902,22 @@ export function getLocations(query, callback) {
 }
 
 /**
- * Make a GET request to any Primero endpoint. Automatically paginates through all pages.
+ * GET resources from Primero. Automatically paginates through all pages.
  * @public
  * @function
  * @example <caption>Fetch all registry records</caption>
  * get('registry_records');
  * @example <caption>Fetch with a fixed page size of 50</caption>
- * get('registry_records', { count: 50 });
+ * get('registry_records', { per: 50 });
  * @param {string} path - Path to the resource
- * @param {object} query - Query parameters to append to the URL. Use `count` (or `per`) to control the page size.
+ * @param {object} query - Query parameters to append to the URL. Use `per` to control the page size..
  * @returns {Operation}
  */
 export function get(path, query = {}) {
   return async state => {
     const [resolvedPath, resolvedQuery] = expandReferences(state, path, query);
-    const { count, per, ...queryParams } = resolvedQuery;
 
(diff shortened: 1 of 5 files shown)


packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,47 @@ export function getLocations(query, callback) {
   };
 }
 
+/**
+ * Make a GET request to any Primero endpoint. Automatically paginates through all pages.
```
josephjclark: But this isn't support to be a HTTP get request. It's not a HTTP helper - that's the whole point. You can't set headers or anything, and you don't get a HTTP response back.

This is just a general "get a resource". It's close to `getLocations` than `http.get`

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,47 @@ export function getLocations(query, callback) {
...
+ * @param {object} query - Query parameters to append to the URL. Use `count` (or `per`) to control the page size.
+ * @returns {Operation}
+ */
+export function get(path, query = {}) {
+  return async state => {
+    const [resolvedPath, resolvedQuery] = expandReferences(state, path, query);
+    const { count, per, ...queryParams } = resolvedQuery;
+
+    let currentQuery = { ...queryParams };
+    if (count ?? per) currentQuery.per = count ?? per;
```
josephjclark: Worried about this - count and per should not be the same thing?

I think `count` is an openfn invention? I can only see `per` in the primero docs

I might want `{ per:10, count: 50 }` to fetch 50 items 10 pages at a time

So this `count` is an openfn only value, which we do not include in the query params, and for consistency with other APIs should probably be called limit

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,47 @@ export function getLocations(query, callback) {
...
+
+    let results = [];
+
+    do {
+      const response = await utilRequest(state, 'GET', resolvedPath, {
+        query: currentQuery,
+      });
+
+      results.push(...(response.body.data ?? []));
```
josephjclark: This is a messy line. If `repsonse.body.data` is empty, doesn't that feel like an error state? Should we do  somethign about that?

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,47 @@ export function getLocations(query, callback) {
...
+ * @param {string} path - Path to the resource
+ * @param {object} query - Query parameters to append to the URL. Use `count` (or `per`) to control the page size.
+ * @returns {Operation}
+ */
+export function get(path, query = {}) {
+  return async state => {
+    const [resolvedPath, resolvedQuery] = expandReferences(state, path, query);
+    const { count, per, ...queryParams } = resolvedQuery;
+
+    let currentQuery = { ...queryParams };
```
josephjclark: `queryParams` has already been spread into a new object. No need to spread again.

packages/primero/src/Utils.js
```diff
@@ -14,6 +14,22 @@ export const prepareNextState = (state, response) => {
...
 
+const DEFAULT_PAGE_SIZE = 1000;
+
+
+export function getNextPageParams(metadata, queryParams = {}) {
+  const { total, page } = metadata ?? {};
+  if (!total || !page) return null;
+
+  // If the user sets per or count, treat it as a one-page limit
+  if (queryParams.count != null || queryParams.per != null) return null;
```
josephjclark: dubious `count` stuff here too

josephjclark (changes requested):
Nervous about the `count` stuff and the API

PR description mentions a re-usable pagination function. Do you mean reusable for other adaptors, or just this?

packages/primero/src/Adaptor.js
hunterachieng: Ah I think because the example was with `count` I went with both. But I can remove `count` and remain with `per`

josephjclark: Just one option please

packages/primero/src/Utils.js
```diff
@@ -14,6 +14,19 @@ export const prepareNextState = (state, response) => {
   };
 };
 
+const DEFAULT_PAGE_SIZE = 1000;
+
+
+export function getNextPageParams(metadata) {
```
josephjclark, line 19: This only uses the default page size to work out the pagination. It doesn't take `per` into account.

But what if I only want 10 items per page? The values will all be wrong

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,43 @@ export function getLocations(query, callback) {
...
+ * @function
+ * @example <caption>Fetch all registry records</caption>
+ * get('registry_records');
+ * @example <caption>Fetch with a fixed page size of 50</caption>
+ * get('registry_records', { per: 50 });
+ * @param {string} path - Path to the resource
+ * @param {object} query - Query parameters to append to the URL. Use `per` to control the page size..
+ * @returns {Operation}
+ */
+export function get(path, query = {}) {
```
josephjclark, line 920: Are you planning to add pagination to the other `getX` functions, or just this one?

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,43 @@ export function getLocations(query, callback) {
   };
 }
 
+/**
+ * GET resources from Primero. Automatically paginates through all pages.
```
josephjclark: Shouldn't be a `GET` in capitals because the HTTP verb is irrelevant. It just `gets a resource by type from primero`

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,43 @@ export function getLocations(query, callback) {
...
+ * @public
+ * @function
+ * @example <caption>Fetch all registry records</caption>
+ * get('registry_records');
+ * @example <caption>Fetch with a fixed page size of 50</caption>
+ * get('registry_records', { per: 50 });
+ * @param {string} path - Path to the resource
+ * @param {object} query - Query parameters to append to the URL. Use `per` to control the page size..
```
josephjclark: Is there also an option to set a limit? What if I only want 5000 items? I cant use  `per` for that.

We need a `limit` paramter as well, which is OpenFN only and doesn't get appeneded to the URL

packages/primero/src/Adaptor.js
```diff
@@ -901,6 +901,43 @@ export function getLocations(query, callback) {
...
 
+/**
+ * GET resources from Primero. Automatically paginates through all pages.
+ * @public
+ * @function
+ * @example <caption>Fetch all registry records</caption>
+ * get('registry_records');
+ * @example <caption>Fetch with a fixed page size of 50</caption>
+ * get('registry_records', { per: 50 });
+ * @param {string} path - Path to the resource
```
josephjclark: Is this really a path a resource, or is it just a resource type?

`path to a resource` is very much HTTP language.

Are there use cases in primero where a `get` will include a path?

Does this API support getting a single item by id? `get('registry_records/12345')` ? Will that break the pagination logic?

josephjclark:
@hunterachieng I've added two small commits but the pagination stuff still isn't looking right to me.

packages/primero/src/Adaptor.js
hunterachieng: Ah I had missed the importance of this section. I have updated everything

packages/primero/src/Adaptor.js
```diff
@@ -902,37 +902,31 @@ export function getLocations(query, callback) {
 }
 
 /**
- * GET resources from Primero. Automatically paginates through all pages.
+ * Get resources from Primero. Automatically paginates through all pages.
  * @public
  * @function
  * @example <caption>Fetch all registry records</caption>
  * get('registry_records');
  * @example <caption>Fetch with a fixed page size of 50</caption>
```
josephjclark: should we update this example to "fetch all records in pages of 50 at a time"?

packages/primero/test/index.js
```diff
@@ -88,3 +110,111 @@ describe.skip('The getCases() function', () => {
     });
   });
 });
+
+describe('getNextPageParams', () => {
```
josephjclark, line 114: If I ask for 6 items (`limit:60`) with a page size of 5 (`per: 5)`, and there are 20 items in total, what will happen?

Will I get back 6, 10 or 20 items?

Please add a test so show the behaviour.

(I suspect we'll get 10 items back, which is incorrect)

hunterachieng: I saw the fault here. I have fixed and it will now return 6 item with per 5 then 1

josephjclark (approved):

